PR #106, opened 2023-08-30
Add an interface class for PredictoorHelper contract and tests

### LINKED ISSUE #105: Add an interface class for PredictoorHelper contract and tests
Add an interface class for PredictoorHelper contract and tests

trizin:
Fixes #105

Changes proposed in this PR:

- Add PredictoorBatcher class and tests
- Add ERC721 model

### DIFF
--- a/pdr_backend/models/data_nft.py
+++ b/pdr_backend/models/data_nft.py
@@ -17,7 +17,7 @@ class DataNft:
         self.contract_address = config.w3.to_checksum_address(address)
         self.contract_instance = config.w3.eth.contract(
             address=config.w3.to_checksum_address(address),
-            abi=get_contract_abi("IERC721Template"),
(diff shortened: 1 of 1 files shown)


pdr_backend/models/erc721.py
```diff
@@ -0,0 +1,25 @@
+from typing import Union
```
trentmc, line 1: Q: should this be called erc721.py (class ERC721) or data_nft.py (class DataNft)?

Investigate: well what does it _do_ here?

Result: it's calling `addToCreateERC20List`. DataNft has this functionality, but canonical ERC721 does not.

Answer to Q: call it data_nft.py (class DataNft)

trentmc (changes requested):
Smallish change requested. Then I'll be ready to approve:)

trentmc (approved):

PR #113, opened 2023-08-31
Refactor and test trader

calina-c:


pdr_backend/models/feed.py
```diff
@@ -74,10 +74,10 @@ def dictToFeed(feed_dict: Dict[str, Any]):
...
-        trueval_submit_timeout=int(d["trueval_submit_timeout"]),
-        owner=d["owner"],
-        pair=d["pair"],
-        timeframe=d["timeframe"],
-        source=d["source"],
+        trueval_submit_timeout=int(d.get("trueval_submit_timeout", 0)),
+        owner=d.get("owner", ""),
+        pair=d.get("pair", "") or "",
+        timeframe=d["timeframe"] or "",
+        source=d["source"] or "",
```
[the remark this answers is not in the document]
calina-c: the runner is failing for me even before introducing my own tests, due to these missing... I'll reinvestigate why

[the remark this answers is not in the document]
trentmc: Please add a tests for this updated functionality. ie when certain envvars don't exist

pdr_backend/models/predictoor_contract.py
```diff
@@ -193,28 +193,24 @@ def get_current_epoch_ts(self) -> int:
...
+
+        print("Reading contract values...")
+        auth = self.get_auth_signature()
+        (nom_wei, denom_wei) = self.contract_instance.functions.getAggPredval(
+            timestamp, auth
+        ).call({"from": self.config.owner})
+        nom = float(self.config.w3.from_wei(nom_wei, "ether"))
+        denom = float(self.config.w3.from_wei(denom_wei, "ether"))
+        print(f" Got {nom} and {denom}")
+        if denom == 0:
```
calina-c, line 211: how about if denom == 0: nom = 0? the return statement would be the same

pdr_backend/models/feed.py
```diff
@@ -74,10 +74,10 @@ def dictToFeed(feed_dict: Dict[str, Any]):
         symbol=d["symbol"],
         seconds_per_epoch=int(d["seconds_per_epoch"]),
         seconds_per_subscription=int(d["seconds_per_subscription"]),
-        trueval_submit_timeout=int(d.get("trueval_submit_timeout", "100") or 100),
-        owner=d.get("owner", "") or "",
-        pair=d.get("pair", "") or "",
-        timeframe=d.get("timeframe") or "",
-        source=d.get("source") or "",
+        trueval_submit_timeout=int(d["trueval_submit_timeout"]),
```
calina-c, line 77: I'm fine with you reverting this, but it was an issue when running main.py (it's true, it doesn't cause problems with tests). We can reopen this if it occurs again, maybe it was something I published wrongly. If you want, we can create a ticket to track this, because without error handling it's possible that someone breaks the entire agent by publishing something unexpected/wrong.

pdr_backend/trader/trader_agent.py
```diff
@@ -8,6 +7,20 @@
 from pdr_backend.models.feed import Feed
 from pdr_backend.trader.trader_config import TraderConfig
 
+MAX_TRIES = 5
+
+
+def timer_decorator(func):
```
calina-c, line 13: I think the "decorator" name for a decorator is a bit redundant

trentmc (approved):

PR #134, opened 2023-09-04
Reduce random stake amount

### LINKED ISSUE #133: Reduce random stake amount
We can't afford that much!

trizin:
Fixes #133 

Changes proposed in this PR:

- Reduce random stake amount


trentmc:
Before I review this, the change from int -> float implies that we need to agree on something first:

does this represent stake in units of Eth, or in units of wei?

We need to make sure it's well documented all the way along.

trentmc (approved):

PR #159, opened 2023-09-08
#157 - [trader] Move max tries to trader agent config class

### LINKED ISSUE #157: [trader] MAX_TRIES should be in trader config, not a global
### Background / motivation

MAX_TRIES is set at the top of trader_agent.py as a global variable.

Yet we have a much better place for this: TraderConfig.

### TODO

Move MAX_TRIES into TraderConfig. It can be hardcoded, we don't need an envvar.

trizin:
Fixes #157

Changes proposed in this PR:

- Move max tries to trader agent config clas

### DIFF
--- a/pdr_backend/trader/test/test_trader_agent.py
+++ b/pdr_backend/trader/test/test_trader_agent.py
@@ -46,7 +46,6 @@ def test_run():
     trader_config.max_tries = 10
     agent = TraderAgent(trader_config, do_trade)
 
-
     with patch.object(agent, "take_step") as ts_mock:
         agent.run(True)
 


pdr_backend/trader/trader_config.py
```diff
@@ -10,4 +10,5 @@ class TraderConfig(BaseConfig):
     def __init__(self):
         super().__init__()
 
-        self.trader_min_buffer = int(getenv("TRADER_MIN_BUFFER", "60 "))
+        self.trader_min_buffer = int(getenv("TRADER_MIN_BUFFER", "60"))
+        self.max_tries = int(getenv("TRADER_MAX_TRIES", "10"))
```
trentmc, line 14: I don't understand.

Why are there two config variables being set by one envvar?

And what is "trader min buffer" anyway?

As mentioned : we don't need to have an envvar for every parameter. We should hardcode some vars in the config for now: ones that don't need much changing because they have a very good default.

trentmc (changes requested):

pdr_backend/trader/trader_config.py
trentmc: It's not always better to make an envvar and make a default. Every envvar adds complexity because you have to document it elsewhere.

Hardcoding is fine, a good thing actually, in specially fenced locations. These are: constants.py, config classes, and "ss" (solution strategy) classes. You'll see as classes coming in when Calina updates predictoor approach 2.

trentmc: Btw: you need to document in the code what trader min buffer means. Just a short one liner above where it's set.

trentmc (approved):

PR #259, opened 2023-10-18
[predictoor] Use float instead of int for stake amount

trizin:
Hotfix

Changes proposed in this PR:

- [predictoor] Use float instead of int for stake amount
- Remove `/ 100` in stake calculation in approach 2 so stake amount is equal to: `STAKE_AMOUNT * confidence`


trentmc (changes requested):
It's not clear to me if this is represented in units of wei or of eth (and therefore int vs float).

At this level, and everywhere related to.

A "1" could have been anything, as a placeholder.

Please add comments here - and elsewhere as needed - to clarify what's in units of wei vs eth.

trentmc (approved):

PR #269, opened 2023-10-19
Add instructions for adding OASIS Sapphire to Metamask to testnet-faucet.md

ChristianCasazza:
Fixes # .

Added link to the Oasis docs to add their network to your Metamask, and a link to setting up a metamask

- 
- 
-

### DIFF
--- a/READMEs/testnet-faucet.md
+++ b/READMEs/testnet-faucet.md
@@ -1,10 +1,16 @@
 # Getting (fake) Tokens on Sapphire testnet
 
+Adding Oasis Saphire EVM to Metamask
+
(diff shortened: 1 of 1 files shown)


trentmc:
We will merge this as part of [#283](https://github.com/oceanprotocol/pdr-backend/issues/283) "Consolidate tokens READMEs". 

_Not_ before, it's too confusing.

trentmc:
I thought more about this PR. I was planning to integrate its ideas as part of [#283](https://github.com/oceanprotocol/pdr-backend/issues/283) "Consolidate tokens READMEs". However I realize it does NOT make sense to consolidate to a single .md file, because there is complexity in each sub-step. One md would be way too big.

As for _this_ PR, it's in the wrong place. Too prominent of a location for what it does. TBH I don't think people need it: if you're skilled enough to run a bot, you're skilled enough to add Sapphire network. Therefore, closing.

PR #300, opened 2023-10-31
Fix #296 & pdr-web#284: get_predictoor_info.py refactor & lite module for serving pdr metrics

### LINKED ISSUE #296: get_predictoor_info.py needs to be more modular, and tested
### Background / motivation

[get_predictoor_info.py](https://github.com/oceanprotocol/pdr-backend/blob/main/scripts/get_predictoor_info.py) was recently built for to help EthMiami judging. 

Example output, from [this slack thread](https://oceanprotocol.slack.com/archives/C054ZHZ11RU/p1698332084628139?thread_ts=1698332065.964699&cid=C054ZHZ11RU):
```text
Overall Accuracy: 51.08%
Accuracy for Pair: ADA/USDT, Timeframe: 5m: 50.63%
Total stake: 282.70319568418233
Total payout: 174.74360545310648

Accuracy for Pair: XRP/USDT, Timeframe: 1h: 50.00%
Total stake: 26.458942555410236
Total payout: 11.77390896354474

Accuracy for Pair: ETH/USDT, Timeframe: 5m: 52.08%
Total stake: 277.54203431195
Total payout: 149.30581199262107
...
```

This is a great start. It / its components can be useful in many places.

However it's not possible to re-use it elsewhere because its components are all within the script itself. 

And, it's not tested. It could have errors.

### Components

Each bullet below is a component in the script. 

- `class Prediction` 
- `get_all_predictions()` subgraph query
- `write_csv()` to output predictions to a csv 
- `date_to_unix()` 
- `get_statistics()` 

### TODOs

Here are recommended refactors. (They may be a bit different in practice, as it gets done. That's fine.)

- [ ] `class Prediction` --> put in `models/`, as its own module `prediction.py`. With unit tests.
- [ ] `get_all_predictions()` subgraph query --> put in `util/subgraph.py` or new file `util/subgraph-predictions.py`. Add unit test.
- [ ] deprecate `date_to_unix()`, use functions in predictoor/approach3/timeutil.py instead. (BTW the location of timeutil.py is not general enough; we have another github issue to handle that.)
- [ ] `write_csv()` --> put into new module `csvs.py`, call it `predictions.csv`. Unit test it. (Have it follow the same pattern as the csvs.py files in df-py, such as [volume/csvs.py](https://github.com/oceanprotocol/df-py/blob/main/df_py/volume/csvs.py))
- [ ] `get_statistics()` currently (a) retrieves data, (b) computes statistics, and (c) prints it. Move the function into a new module `util/predictoor_stats.py`. Refactor the function into two functions: (a)(b) together, and (c) separate. Unit test each function.
- [ ] Have `get_predictoor_info.py` use the above modules in each of their locations. (And now, they're all tested, and reusable FTW)

kdetry:
Fixes #296 and https://github.com/oceanprotocol/pdr-web/issues/284
Changes proposed in this PR:

- the get_predictoor_info script has been modularized
- the pdr_backend/accuracy module is added with flask. It creates JSON output every 5 minutes and serves it with an endpoint
- some methods have been replaced with their alternatives according to Trent's comments
- tests are not completed ~~created some tests but these are not working as expected, I need some backup about testing with Python~~
- the Slot based calculation is added which is like in the FE app (pdr_backend/util/subgraph_slot.py)


kdetry:
> We should be using slot data to determine the accuracy for pairs. Predictoor data becomes available when predictoors claim payouts, otherwise it's not available and will output inaccurate results for pair accuracies.
> 
> The [calculation based on slot data in pdr-web](https://github.com/oceanprotocol/pdr-web/blob/af8bd8ed0c8670bef87144025beee0a9646ab718/src/utils/subgraphs/getAssetAccuracy.ts#L90) is correct.
> 
> I assume the `/statistics` endpoint would only be used for predictoor statistics and not for pairs.

thank you for the feedback, I will fix it asap and update the PR

pdr_backend/util/subgraph_predictions.py
```diff
@@ -0,0 +1,211 @@
...
+from pdr_backend.util.subgraph import query_subgraph, info_from_725
+from pdr_backend.models.prediction import Prediction
+
+
+class FilterMode(Enum):
+    CONTRACT = 1
+    PREDICTOOR = 2
+
+
+def get_subgraph_url(network: str) -> str:
```
idiom-bytes, line 14: Great work consolidating these!

pdr_backend/util/predictoor_stats.py
```diff
@@ -0,0 +1,128 @@
...
+                "payout": data["payout"],
+                "number_of_predictions": data["total"],
+                "details": list(data["details"]),
+            }
+        )
+
+    return results
+
+
+def get_cli_statistics(all_predictions):
```
idiom-bytes, line 105: Nice pattern w/ these:
```
get_endpoint_statistics
get_cli_statistics
```

I'm guessing we'll probably reuse/improve this pattern in the future.

pdr_backend/models/prediction.py
```diff
@@ -0,0 +1,23 @@
...
+    ) -> None:
+        self.pair = pair
+        self.timeframe = timeframe
+        self.prediction = prediction
+        self.stake = stake
+        self.trueval = trueval
+        self.timestamp = timestamp
+        self.source = source
+        self.payout = payout
+        self.user = user
```
idiom-bytes, line 23: Good to see this get moved over to `/models/`

pdr_backend/util/subgraph_predictions.py
```diff
@@ -0,0 +1,211 @@
...
+                payout,
+                predictoor_user,
+            )
+            predictions.append(prediction_obj)
+
+    return predictions
+
+
+def get_all_contracts(owner_address: str, network: str) -> List[str]:
+    if network != "mainnet" and network != "testnet":
```
idiom-bytes, line 146: Not a comment for this PR, and this won't work for testnet, but I think we can consolidate this functionality vs. `scripts/addresses.py` and `scripts/get_predictoor_info.py`

pdr_backend/util/subgraph_slot.py
```diff
@@ -0,0 +1,201 @@
...
+    return (
+        total_staked_yesterday,
+        total_staked_today,
+        total_correct_predictions,
+        total_slots_evaluated,
+    )
+
+
+# Function to calculate stats for all assets
+def calculate_statistics_for_all_assets(
```
idiom-bytes, line 174: Great to see this implemented in py/pdr-backend!

idiom-bytes (approved):
Approving this PR.

@kdetry the only comment I would have is around tests. Maybe we can create a separate PR with clearer test conditions, @trizin would perhaps suggest that these should be included in here.

Cheers!

pdr_backend/accuracy/utils/get_start_end_params.py
```diff
@@ -0,0 +1,19 @@
+from datetime import datetime, timedelta
+from typing import Tuple
+
+
+def get_start_end_params(contract_timeframe: str) -> Tuple[int, int]:
```
trentmc, line 5: This is a weird module name, and function name, for what it does. 

And, you shouldn't need a new module.

pdr_backend/util/csvs.py
```diff
@@ -0,0 +1,36 @@
+import os
+import csv
+
+
+def write_csv(all_predictions, csv_output_dir):
```
trentmc, line 5: Rename to `write_prediction_csv()`. We will surely have other csvs.

Also, as mentioned in the github issue, it points to a pattern in df-py csvs how to have function names for writing, reading, and naming the csvs; and accompanying unit tests. We should have that here.

pdr_backend/util/predictoor_stats.py
```diff
@@ -0,0 +1,128 @@
...
+
+PredictionDetail = Tuple[str, str, str]
+PredictoorStats = Dict[str, Union[str, float, int, List[PredictionDetail]]]
+PairTimeframeStats = Dict[str, Union[str, float, int]]
+StatisticsResult = Dict[
+    str, Union[float, List[PredictoorStats], List[PairTimeframeStats]]
+]
+
+
+def get_pure_stats(all_predictions: List[Prediction]):
```
trentmc, line 12: What does "pure" mean? This needs a better name.

pdr_backend/util/predictoor_stats.py
```diff
@@ -0,0 +1,128 @@
...
+    # This is similar to the get_statistics function
+    # but returns the results instead of printing them.
+    total_predictions = len(all_predictions)
+    stats, correct_predictions = get_pure_stats(all_predictions)
+
+    overall_accuracy = (
+        correct_predictions / total_predictions * 100 if total_predictions else 0
+    )
+
+    results = {
```
trentmc, line 69: Why have `results` dict, vs returning a tuple of `(overall_accuracy, pair_timeframe_stats, predictoor_stats)` ?

pdr_backend/util/predictoor_stats.py
```diff
@@ -0,0 +1,128 @@
...
+PredictoorStats = Dict[str, Union[str, float, int, List[PredictionDetail]]]
+PairTimeframeStats = Dict[str, Union[str, float, int]]
+StatisticsResult = Dict[
+    str, Union[float, List[PredictoorStats], List[PairTimeframeStats]]
+]
+
+
+def get_pure_stats(all_predictions: List[Prediction]):
+    stats: Dict[str, Dict] = {"pair_timeframe": {}, "predictoor": {}}
+    correct_predictions = 0
```
trentmc, line 14: Why is `correct_predictions` apart from `stats`? Isn't it a stat too?

And, why have this all as a return dict of `stats` versus returning a tuple of `(correct_predictions, pair_timeframe, predictoor_stats)`?

pdr_backend/util/predictoor_stats.py
```diff
@@ -0,0 +1,128 @@
...
+                "pair": pair,
+                "timeframe": timeframe,
+                "accuracy": accuracy,
+                "stake": data["stake"],
+                "payout": data["payout"],
+                "number_of_predictions": data["total"],
+            }
+        )
+
+    for predictoor, data in stats["predictoor"].items():
```
trentmc, line 89: `predictoor` needs a better label. `predictoor_addr` ?

`data` is way too general, definitely needs a better label.

pdr_backend/util/subgraph_predictions.py
trentmc, line 14: This does not belong in this module, because it shouldn't be specific to subgraph predictions. It either belongs in subgraph.py or networkutil.py. I recommend the latter.

pdr_backend/util/subgraph_predictions.py
trentmc: Perhaps create a new github issue for that @idiom-bytes ?

pdr_backend/util/subgraph_predictions.py
```diff
@@ -0,0 +1,211 @@
...
+        raise Exception("Error fetching contracts: No data returned")
+
+    # Parse the results and construct Contract objects
+    contract_data = result["data"]["tokens"]
+    contracts = [contract["id"] for contract in contract_data]
+
+    return contracts
+
+
+def get_contract_informations(
```
trentmc, line 178: Need a better label. "Get information" is vague.

pdr_backend/util/subgraph_slot.py
```diff
@@ -0,0 +1,201 @@
+from typing import List, Dict, Any, TypedDict
+
+from pdr_backend.util.subgraph import query_subgraph
+from pdr_backend.util.subgraph_predictions import get_subgraph_url
+
+
+class Slot(TypedDict):
```
trentmc, line 7: This overlaps heavily with models/slot.py's Slot class. Please resolve.

pdr_backend/accuracy/app.py
```diff
@@ -0,0 +1,94 @@
+import threading
```
trentmc, line 1: General comment for every new function added in this PR:

Each new function needs:
- a docstring
- one or more unit tests
- a `@enforce_types` decorator on top (so that dynamic type checking is actually invoked)

This is extremely important to maintainability and usability.

scripts/get_predictoor_info.py
```diff
@@ -1,253 +1,8 @@
...
-        print(f"Accuracy for Predictoor Address: {predictoor}: {accuracy:.2f}%")
-        print(f"Stake: {data['stake']}")
-        print(f"Payout: {data['payout']}")
-        print(f"Number of predictions: {data['total']}")
-        print("Details of Predictions:")
-        for detail in data["details"]:
-            print(f"Pair: {detail[0]}, Timeframe: {detail[1]}, Source: {detail[2]}")
-        print("\n")
-
+from pdr_backend.util.csvs import write_csv
```
trentmc, line 2: This file is in way better shape than before. What I was hoping for. Great stuff!

trentmc (changes requested):

pdr_backend/util/subgraph_predictions.py
idiom-bytes: Done - https://github.com/oceanprotocol/pdr-backend/issues/307

pdr_backend/util/predictoor_stats.py
idiom-bytes: Because this data should be served via an API, keeping it as a dict or a named tuple makes sense. This makes it easy to just output json format where you can understand the content because of the labels.

pdr_backend/accuracy/utils/get_start_end_params.py
kdetry: renamed as "calculate_timeframe_timestamps"

pdr_backend/util/csvs.py
kdetry: done

pdr_backend/util/predictoor_stats.py
kdetry: renamed as "aggregate_prediction_statistics"

pdr_backend/util/predictoor_stats.py
kdetry: done

pdr_backend/util/subgraph_predictions.py
kdetry: done

pdr_backend/util/subgraph_predictions.py
kdetry: renamed as "fetch_contract_id_and_spe"

pdr_backend/util/subgraph_slot.py
kdetry: used @dataclass here and created a class named as PredictSlot

pdr_backend/accuracy/app.py
kdetry: - Docs are added to functions
- @enforce_types are added and tested
- Unit tests are not ready yet

kdetry:
tests are added now, I guess ready to review

pdr_backend/accuracy/app.py
trentmc: Looks like docstrings are there, great.

Looks like some @enforce_types are still missing.

trentmc (approved):
Great stuff putting in all the docstrings and tests. Just a few @enforce_types left, and good to go.

I'm going to pre-approve. Once you put in the remaining enforce_types, go ahead and merge. Therefore you don't have me as a bottleneck

PR #304, opened 2023-11-04
add topup cron

alexcos20:
Changes proposed in this PR:

- cosmetic changes to topup.py
- add cron to run every 6 hours, for both mainnet and testnet.   Private key is already set in secrets and bitwarden
- for reference, topup address is 0xbadDf37EbE3763f7c1D1245b1B241c246b96D3dF

Why:
  - Main account should not be used in crons.  Sometimes, we have large amount of tokens in that address. So let's use another one, with small amounts
  - Right now, the script is running on my pc, 24/7.  Let's move it to github
  
Logs from testing:
 - mainnet:  https://github.com/oceanprotocol/pdr-backend/actions/runs/6753225972/job/18359473261?pr=304
 - testnet:  https://github.com/oceanprotocol/pdr-backend/actions/runs/6753225972/job/18359473189?pr=304

### DIFF
--- a/.github/workflows/cron_topup.yml
+++ b/.github/workflows/cron_topup.yml
@@ -1,7 +1,6 @@
 name: Topup accounts
 
 on:
-  pull_request:
   schedule:
     - cron: "0 */6 * * *"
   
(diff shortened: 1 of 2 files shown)


trentmc:
I defer to @trizin to review this

PR #318, opened 2023-11-09
[Pdr bot] Import model.py and update agent2 such that it can run.

### LINKED ISSUE #317: [Predictoor bot approach2] Pdr approach 2 requires model.py/OceanModel
### System Problem:
- Without model.py from `pdr-model-simple`, approach 2 does not run.

### Model Problems:
- The current available trained model inside `pdr-model-simple` does not work
- The current trained gradient boosted models do not work with agent2

### DoD:
- [ ] Implement model.py from pdr-model-simple into approach2, such that you can run a model after having trained it with [pdr-model-simple](https://github.com/oceanprotocol/pdr-model-simple)

idiom-bytes:
Fixes #317 

Changes proposed in this PR:
- Have model.py inside `predictoor/agent2/` such that it can run

Future changes:
- Improve pdr-backend dependencies such that [simple](https://github.com/oceanprotocol/pdr-model-simple), model.py, or other requirements can be installed via pip rather than copied

### DIFF
--- a/pdr_backend/predictoor/approach2/main2.py
+++ b/pdr_backend/predictoor/approach2/main2.py
@@ -21,7 +21,7 @@ from pdr_backend.util.web3_config import Web3Config
 model_dir: str = getenv_or_exit("MODELDIR")
 trained_models_dir = os.path.join(model_dir, "trained_models")
 sys.path.append(model_dir)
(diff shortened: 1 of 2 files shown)


idiom-bytes:
Should we get this merged such that model2 works, or perhaps disable/document model2 such that users are pointed to the `pdr-model-example` and they can take it from there?

I'd love to close this issue.

PR #321, opened 2023-11-09
Update BasePredictoorAgent to only predict once

### LINKED ISSUE #320: Give BasePredictoorAgent memory and make it not submit multiple times
### Problem
Right now predictoor_agent registers they predicted the epoch, but.... 
- It doesn't check if it's already submitted
- It doesn't cache it
- It submits multiple times

Rather, the user should configure `s_until_epoch_end` to set the maximum time window to guarantee/submit a prediction before the epoch ends, only once per epoch prediction/trade.

### DoD:
- [ ] Update base_predictoor_agent (and therefore all predictoor agents) such that it caches + checks if submitted already

idiom-bytes:
Fixes #320 

Changes proposed in this PR:
- Add caching to predictoor agent
- Check if already submitted prediction
- Only submit predictions once

### DIFF
--- a/pdr_backend/predictoor/base_predictoor_agent.py
+++ b/pdr_backend/predictoor/base_predictoor_agent.py
@@ -21,11 +21,7 @@ class BasePredictoorAgent(ABC):
     - When a value can be predicted, qcall predict.py::predict_function()
     """
 
-    def __init__(
-        self,
-        config: BasePredictoorConfig,
-        cache_dir=".cache"
-    ):
+    def __init__(self, config: BasePredictoorConfig, cache_dir=".cache"):
         self.config = config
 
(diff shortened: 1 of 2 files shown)


trentmc:
This "fix" is for an "issue" that isn't actually an issue.

I just [commented](https://github.com/oceanprotocol/pdr-backend/issues/320#issuecomment-1805254896) in the github issue:

> This "issue" is not an issue. It's by design.
> 
> The design is: keep submitting every few seconds until no time left. Each submission may have a bit more information than the previous one, therefore a bit better model.
> 
> Closing this issue. If you feel otherwise, please reopen and we can discuss.

Closing this PR too. Once we are in sync we can delete the branch.

PR #374, opened 2023-11-16
Fixes #359: [PDR Scripts] Predictoor Analytics Raw Transaction Data

### LINKED ISSUE #359: Generate Python Script for Predictoor Analytics Raw Transaction Data
**Is your feature request related to a problem? Please describe.**
Create a Python script to execute the following [query](https://v4.subgraph.sapphire-mainnet.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph/graphql?query=%7B%0A++predictPredictions+%28%0A++++where%3A+%7B%0A++++++slot_%3A+%7B%0A++++++++predictContract%3A+%220x18f54cc21b7a2fdd011bea06bba7801b280e3151%22%2C%0A++++++++slot_gt%3A+1696118400%2C+slot_lt%3A+1699401600%0A++++++%7D%0A++++%7D%0A++++skip%3A+0%2C+first%3A+1000%2C+%0A++%29+%7B%0A++++id%0A++++timestamp%0A++++user+%7B%0A++++++++id%0A++++%7D%0A++++stake%0A++++payout+%7B%0A++++++++payout%0A++++++++trueValue%0A++++++++predictedValue%0A++++%7D%0A++++slot+%7B%0A++++++slot%0A++++++predictContract+%7B%0A++++++++token+%7B%0A++++++++++name%0A++++++++%7D%0A++++++%7D%0A++++%7D%0A%7D%0A%7D)

The query will produce a dataset for a particular query with the schema predictionID, epoch, timestamp, stake, wallet address, payout, trueValue, and predicted value

**Describe the solution you'd like**
The script should produce CSV files for each feed. They can originally be manually consolidated together, and aggregation can be automated into the script later.

**Additional context**
The goal is to support this [epic ](https://github.com/oceanprotocol/pdr-analytics/issues/1). We want to use the subgraphs to create a raw source dataset of all predictions so that all analytics can be conducted downstream with tools better optimized for column-level queries.

kdetry:
Fixes #359 

Changes proposed in this PR:

- added a new script called as get_contract_predictions_info.py
- the csvs.py file is refactored

### DIFF
--- a/pdr_backend/models/prediction.py
+++ b/pdr_backend/models/prediction.py
@@ -11,6 +11,8 @@ class Prediction:
         source,
         payout,
         user,
+        submittimestamp,
+        predictionId
     ) -> None:
         self.pair = pair
         self.timeframe = timeframe
@@ -21,3 +23,5 @@ class Prediction:
         self.source = source
         self.payout = payout
         self.user = user
+        self.submittimestamp = submittimestamp
+        self.predictionId = predictionId
--- a/pdr_backend/util/csvs.py
+++ b/pdr_backend/util/csvs.py
@@ -1,28 +1,57 @@
 import os
 import csv
+from typing import List, Dict
+from io import TextIOWrapper
+
+from enforce_typing import enforce_types
+from pdr_backend.util.subgraph_predictions import Prediction
 
 
-def write_prediction_csv(all_predictions, csv_output_dir):
-    if not os.path.exists(csv_output_dir):
-        os.makedirs(csv_output_dir)
+@enforce_types
+def create_csv_header_columns(columns: List[str], file: TextIOWrapper):
+    writer = csv.writer(file)
+    writer.writerow(columns)
+    return writer
 
-    data = {}
-    for prediction in all_predictions:
+@enforce_types
+def generate_csv_file_path(csv_output_dir: str, key: str) -> str:
+    return os.path.join(
+        csv_output_dir,
+        key + ".csv",
+    )
+
+@enforce_types
+def generate_prediction_data_structure(
+    predictions: List[Prediction]) -> Dict[str, List[Prediction]]:
+
+    data: Dict[str, List[Prediction]] = {}
+    for prediction in predictions:
         key = (
             prediction.pair.replace("/", "-") + prediction.timeframe + prediction.source
         )
         if key not in data:
             data[key] = []
         data[key].append(prediction)
+    return data
+
+@enforce_types
+def check_and_create_dir(dir_path: str):
+    if not os.path.isdir(dir_path):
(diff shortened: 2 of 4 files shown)


pdr_backend/util/csvs.py
```diff
@@ -1,28 +1,62 @@
 import os
 import csv
+from typing import List, Dict
```
trentmc, line 3: In this file, please make all the names of functions follow the same patterns as in df-py. Eg https://github.com/oceanprotocol/df-py/blob/main/df_py/volume/csvs.py. 

This reduces cognitive load as we go from repo to repo.

pdr_backend/util/csvs.py
```diff
@@ -1,28 +1,62 @@
...
 
     for key, predictions in data.items():
         predictions.sort(key=lambda x: x.timestamp)
-        filename = os.path.join(csv_output_dir, key + ".csv")
+        filename = generate_csv_file_path(csv_output_dir, key)
         with open(filename, "w", newline="") as file:
-            writer = csv.writer(file)
-            writer.writerow(
-                ["Predicted Value", "True Value", "Timestamp", "Stake", "Payout"]
+            writer = create_csv_header_columns(
```
trentmc, line 56: It's unwanted complexity to have a separate method just for writing columns. Keep it in the same method.

pdr_backend/models/prediction.py
```diff
@@ -11,6 +11,8 @@ def __init__(
         source,
         payout,
         user,
+        submittimestamp,
```
trentmc, line 14: Any reason this is not `submit_timestamp` ? (Snake case)

pdr_backend/models/prediction.py
```diff
@@ -11,6 +11,8 @@ def __init__(
...
+        predictionId,
```
trentmc, line 15: Any reason this is not `prediction_id` ? (Snake case)

trentmc (changes requested):
Close! Some smallish changes requested.

pdr_backend/models/prediction.py
kdetry: solved

pdr_backend/models/prediction.py
kdetry: done

pdr_backend/util/csvs.py
kdetry: removed

pdr_backend/util/csvs.py
kdetry: changed function names

kdetry:
I changed the logic in the fetch function for reports and we assume non-claimed predictions as wrong.

PR #377, opened 2023-11-18
issue376: accuracy rate issue

### LINKED ISSUE #376: [PDR Accuracy]: Accuracy Rate Issue
Some accuracy rates look like lower than supposed 
DoD:
- take the output of all steps on the PDR accuracy and find what causes the accuracy rate issue

kdetry:
Fixes #376 

Changes proposed in this PR:

- fixed the accuracy/direction calculation issue

### DIFF
--- a/pdr_backend/util/subgraph_slot.py
+++ b/pdr_backend/util/subgraph_slot.py
@@ -179,19 +179,29 @@ def fetch_slots_for_all_assets(
 @enforce_types
 def calculate_prediction_prediction_result(
     round_sum_stakes_up: float, round_sum_stakes: float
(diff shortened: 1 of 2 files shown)


pdr_backend/util/subgraph_slot.py
```diff
@@ -179,19 +179,29 @@ def fetch_slots_for_all_assets(
 @enforce_types
 def calculate_prediction_prediction_result(
```
trentmc, line 180: Why does "prediction" appear twice in this function name?

trentmc (approved):

PR #391, opened 2023-11-22
Configure publisher and trueval for yaml

trizin:
# Trueval

- Add TruevalSS
- Update Trueval Baseconfig, agents and tests to use PPSS and TruevalSS.
- Removed Trueval main
- Added Trueval CLI command into pdr


trentmc (approved):
Just

PR #392, opened 2023-11-22
Dfbuyer yaml

trizin:
Changes proposed in this PR:

- Created DFBuyerSS + tests
- Updated DFBuyer agent and its tests to use PPSS
- Added DFBuyerSS into PPSS
- Removed DFBuyer main.py
- Updated CLI module do_dfbuyer to run DFBuyer agent.


pdr_backend/dfbuyer/test/conftest.py
```diff
@@ -1,7 +1,8 @@
 from unittest.mock import MagicMock, Mock, patch
 import pytest
+
```
trentmc, line 3: You'd deleted test_main.py, but we still need a way to test dfbuyer.

trentmc (changes requested):

trentmc (approved):

PR #431, opened 2023-12-01
#430 - Do not fetch pending slots older than 3 days

### LINKED ISSUE #430: Do not get pending slots older than 3 days
Do not get pending slots older than 3 days

trizin:
Fixes #430

Changes proposed in this PR:

- Do not fetch pending slots older than 3 days
- 
-

### DIFF
--- a/pdr_backend/util/subgraph.py
+++ b/pdr_backend/util/subgraph.py
@@ -284,8 +284,8 @@ def get_pending_slots(
     slots: List[Slot] = []
 
     now_ts = time.time()
-    # rounds older than 3 days are canceled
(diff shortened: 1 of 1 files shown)


alexcos20 (approved):

alexcos20 (changes requested):
actually, we need a way of updating old slots to 'Canceled'..
our predictors rely on subgraph to claim their rewards.   but if we just choose to ignore old slots, they will not be updated by the graph and users will never get their money back..

On a 2nd look, payout will trigger all unpaid slots, so somebody "will" change the status of the slot.
Let's go ahead and merge then

alexcos20 (approved):

PR #460, opened 2023-12-17
Fix #459: In CI, polars error: col timestamp_right already exists

### LINKED ISSUE #459: [YAML] In CI, polars error: 'column with name "timestamp_right" already exists'
### Summary 

My (Trent's) local tests all pass. But in CI, everything passes these two tests:

```text
=========================== short test summary info ============================
FAILED pdr_backend/aimodel/test/test_aimodel_data_factory.py::test_create_xy__2exchanges_2coins_2signals - polars.exceptions.DuplicateError: unable to hstack, column with name "timestamp_right" already exists
FAILED pdr_backend/lake/test/test_ohlcv_data_factory.py::test_get_mergedohlcv_df_happypath - polars.exceptions.DuplicateError: unable to hstack, column with name "timestamp_right" already exists
======= 2 failed, 300 passed, 3 skipped, 9 warnings in 102.95s (0:01:42) =======
```

[Full traceback here](https://github.com/oceanprotocol/pdr-backend/actions/runs/7239388782/job/19721189365?pr=458). It originated in the gql data factory work, when polars was introduced. Most CI runs since the gql merge fail due to this, though sometimes they _do_ pass (20%?).

[Here's a useful blog post](https://medium.com/@s.illesinghe/how-to-join-or-merge-data-in-polars-8f87cd708e1b) about polars "join()".

What 's almost certainly happening: it's successful on the first merge, but leave a "timestamp_right" column as an artifact. Then for the second merge it can't add "timestamp_right" column because it's already there, and complains.
- I should be able to reproduce this in a unit test

I don't know why it's not happening for me (Trent) locally and not in CI. 

I've also noticed that `_merge_rawohlcv_dfs()` got 2x more complex from the gql merge. Let's see if we can refactor to make it simpler again, as part of the fix. Below I show the full code of this function before & after gql.

### Details on problem area

It's in `ohlcv_data_factory::_merge_rawohlcv_dfs()` in this code:
```python
                # join rawohclv_df into merged_df
                if merged_df.shape[0] == 0:
                    merged_df = raw_df
                else:
                    merged_df = merged_df.join(
                        raw_df, on="timestamp", how="outer"
                    )
```

### Full code of problem area
```python
    def _merge_rawohlcv_dfs(self, rawohlcv_dfs: dict) -> pl.DataFrame:
        """
        @arguments
          rawohlcv_dfs -- see class docstring

        @return
          mergedohlcv_df -- see class docstring
        """
        raw_dfs = rawohlcv_dfs
        
        # init mergedohlcv_df such that it can do basic operations
        print("  Merge rawohlcv dataframes.")
        merged_df = initialize_df()  # grow this
        merged_cols = ["timestamp"]  # grow this
        for exch_str in raw_dfs.keys():
            for pair_str, raw_df in raw_dfs[exch_str].items():
                assert "/" in pair_str, f"pair_str={pair_str} needs '/'"
                assert "datetime" in raw_df.columns
                assert "timestamp" in raw_df.columns

                for raw_col in raw_df.columns:
                    if raw_col in ["timestamp", "datetime"]:
                        continue

                    signal_str = raw_col  # eg "close"
                    merged_col = f"{exch_str}:{pair_str}:{signal_str}"

                    raw_df = raw_df.with_columns(
                        [pl.col(raw_col).alias(merged_col)]
                    )
                    merged_cols.append(merged_col)

                # drop columns we won't merge
                # drop original OHLCV cols and datetime
                raw_df = raw_df.drop(OHLCV_COLS)
                if "datetime" in merged_df.columns:
                    raw_df = raw_df.drop("datetime")

                # only keep OHCLV cols
                merged_cols = [
                    col for col in merged_cols if col not in OHLCV_COLS
                ]

                # join rawohclv_df into merged_df
                if merged_df.shape[0] == 0:
                    merged_df = raw_df
                else:
                    merged_df = merged_df.join(
                        raw_df, on="timestamp", how="outer"
                    )

        # select columns in-order [timestamp, ..., datetime]
        merged_df = merged_df.select(merged_cols + ["datetime"])

        assert "datetime" in merged_df.columns
        assert "timestamp" in merged_df.columns

        mergedohlcv_df = merged_df
        return mergedohlcv_df
```

### Code before the gql changes

From [here](https://github.com/oceanprotocol/pdr-backend/blob/5c1a67796c680c610111cd238005f639d298d5f3/pdr_backend/data_eng/data_factory.py#L190):
```
python
    def _merge_csv_dfs(self, csv_dfs: dict) -> pd.DataFrame:
        """
        @arguments
          csv_dfs -- dict [exch_str][pair_str] : df
            where df has cols={signal_str}+"datetime", and index=timestamp
        @return
          hist_df -- df w/ cols={exch_str}:{pair_str}:{signal_str}+"datetime",
            and index=timestamp
        """
        print("  Merge csv DFs.")
        hist_df = pd.DataFrame()
        for exch_str in csv_dfs.keys():
            for pair_str, csv_df in csv_dfs[exch_str].items():
                assert "-" in pair_str, pair_str
                assert "datetime" in csv_df.columns
                assert csv_df.index.name == "timestamp"

                for csv_col in csv_df.columns:
                    if csv_col == "datetime":
                        if "datetime" in hist_df.columns:
                            continue
                        hist_col = csv_col
                    else:
                        signal_str = csv_col  # eg "close"
                        hist_col = f"{exch_str}:{pair_str}:{signal_str}"
                    hist_df[hist_col] = csv_df[csv_col]

        assert "datetime" in hist_df.columns
        assert hist_df.index.name == "timestamp"
        return hist_df
```

trentmc:


pdr_backend/aimodel/aimodel_data_factory.py
```diff
@@ -120,6 +119,9 @@ def create_xy(
         assert X.shape[1] == ss.n
         assert isinstance(x_df, pd.DataFrame)
 
+        assert "timestamp" not in x_df.columns
```
idiom-bytes, line 122: I'm assuming no timestamp because now it's the pd index

pdr_backend/lake/merge_df.py
```diff
@@ -0,0 +1,98 @@
...
+    return ordered_cols
+
+
+@enforce_types
+def _verify_df_cols(df: pl.DataFrame):
+    assert "timestamp" in df.columns
+    assert "datetime" not in df.columns
+    for col in df.columns:
+        assert "_right" not in col
+        assert "_left" not in col
```
idiom-bytes, line 97: the reason why we were getting _right,_left is because of outer joins. As long as we user inner joins, we should not see these.

pdr_backend/lake/merge_df.py
```diff
@@ -0,0 +1,98 @@
...
+    newraw_df = raw_df.with_columns(
+        pl.col(raw_col).alias(merged_col),
+    )
+    newraw_df = newraw_df.select(["timestamp", merged_col])
+
+    if merged_df is None:
+        merged_df = newraw_df
+    else:
+        merged_df = merged_df.join(newraw_df, on="timestamp", how="outer")
+        merged_df = merge_cols(merged_df, "timestamp", "timestamp_right")
```
idiom-bytes, line 59: I believe the reason why we are getting _right columns, is because of the outer join. Generally, I don't like using outer joins as they can introduce unexpected data.

As long as we use left joins on expected timestamps, this should not happen. This can be done in two ways:
1. With a dataset you trust (i.e. kaiko that may have guaranteed timestamps)
2. With a dataset you build (i.e. just populate 1 column w/ expect values)

example
```
# Create a range of timestamps with 5-minute intervals
timestamps = pd.date_range(start='2023-01-01', periods=100, freq='5T')
timestamps = timestamps.astype(int) // 10**9  # Convert to Unix timestamps in seconds

# Create a Polars DataFrame
df = pl.DataFrame({
    "timestamp": timestamps
})

merged_df.extend(df)
merged_df = merged_df.join(newraw_df, on="timestamp", how="left")
```

This avoids:
- situations where you get _right/_left columns
- the need to run merge_cols()

This also enforces strict timestamp records in your data structure, rather than potentially introducing weird records due to outer join.

pdr_backend/aimodel/aimodel_data_factory.py
trentmc: Yes

pdr_backend/lake/merge_df.py
trentmc: Fyi I was building off the outer join code that you had introduced. 

We can't just use "data we trust". There's always going to be missing values etc and we need to account for it. Eg most people using pdr-backend will not be buying Kaiko data.

The example improvement you gave us to a good idea. If you want to make it happen, I recommend to make a new github issue.

Issue #451, opened 2023-12-13
[YAML] Refine ppss.yaml for closer 1:1 mapping to bots


trentmc:
### Background / motivation

Parent: [#400](https://github.com/oceanprotocol/pdr-backend/issues/400) "Re-merge YAML & CLI: revive changes, fix bugs, test better, re-merge"

This helps via:
- Clarify what ohlcv feeds lake will work with
- Clarify what gql feeds lake will work with
- Clarify - for each bot on its own terms - what predictoor feeds it works with. Predictoor, trader, trueval, etc
- The 1:1 mapping makes it more clear how each bot uses each setup

### Pre-requisites and TODOs

- [x] [#446](https://github.com/oceanprotocol/pdr-backend/issues/446) [YAML] Rename/move files & dirs for proper separation among lake, AI models, analytics
  - [x] refactor feedstr, timeframestr, pairstr, and exchangestr from tuples -> classes/objects
- [x] [#462](https://github.com/oceanprotocol/pdr-backend/issues/462) Make `pdr lake` cli-driven command. Do this first!
- [x] Move {max_n_train, autoregressive_n} from data_ss into predictoor_ss.
- [x] Implement support for the rest of ppss.yaml (draft below)
- [x] Ensure: for trueval agent, that it's properly filtering out feeds according to trueval_ss. This will fix 
- [x] Ensure: that sim_ss is using predictoor_ss and trader_ss
- [x] Is each bot / command properly filtering out feeds according to its corresponding ss?
  - [x] base_predictoor_agent.py -> filters on predictoor_ss.predict_feed and predictoor_ss.ai_model_ss.input_feeds
  - [x] base_trader_agent.py -> filters on trader_ss.feed
  - [x] sim_agent.py -> no work, since it's using predictoor_ss and trader_ss
  - [x] publisher_agent.py -> filters on publisher_ss.feeds -> moved to https://github.com/oceanprotocol/pdr-backend/issues/490
  - [x] trueval_agent.py (?) -> filters on trueval_ss.feeds. Note: doing this will fix [#449](https://github.com/oceanprotocol/pdr-backend/issues/449)
  - [x] dfbuyer_agent.py (?) -> filters on dfbuyer_ss.feeds
  - [x] (For the above: should minimize DRY violations when the feeds are grabbed / printed / filtered / printed. Eg via a new helper module in subgraph/)  
  - [x] Any other `pdr` command for stats: what filtering is needed, if any?

[...]

PR #464, opened 2023-12-20
Make Feeds objects instead of tuples.

calina-c:
Work on #451.

### DIFF
--- a/pdr_backend/aimodel/aimodel_data_factory.py
+++ b/pdr_backend/aimodel/aimodel_data_factory.py
@@ -1,14 +1,14 @@
 import sys
 from typing import Tuple
 
-from enforce_typing import enforce_types
 import numpy as np
 import pandas as pd
 import polars as pl
+from enforce_typing import enforce_types
 
 from pdr_backend.ppss.data_pp import DataPP
 from pdr_backend.ppss.data_ss import DataSS
-from pdr_backend.util.mathutil import has_nan, fill_nans
+from pdr_backend.util.mathutil import fill_nans, has_nan
 
 
 @enforce_types
@@ -79,8 +79,7 @@ class AimodelDataFactory:
         x_df = pd.DataFrame()  # build this up
 
         target_hist_cols = [
-            f"{exch_str}:{pair_str}:{signal_str}"
-            for exch_str, signal_str, pair_str in ss.input_feed_tups
+            f"{feed.exchange}:{feed.pair}:{feed.signal}" for feed in ss.input_feeds
         ]
 
         for hist_col in target_hist_cols:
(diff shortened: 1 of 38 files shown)


calina-c:
@trentmc just one more question: should I support feeds without signals, already? I see the option will exist for publisher and trueval, but it is not used yet.

trentmc:
> @trentmc just one more question: should I support feeds without signals, already?

Yes.

pdr_backend/aimodel/aimodel_data_factory.py
```diff
@@ -1,14 +1,14 @@
 import sys
 from typing import Tuple
```
trentmc, line 2: ### Top-level comment

This PR is a great start! My feedback below is for things that still need doing for this bullet point of this issue, ie "refactor feedstr, timeframestr, pairstr, and exchangestr from tuples -> classes/objects".

You can do them as part of this PR, or you can do follow-up PR(s). 

Given that things are passing, I recommend to do it in follow-up PR(s). More tractable.

### 1. Changing **arg** modules / classes / tests:

You've created ArgFeed and got it working. Great! 
- You still need to change its module from feedstr.py -> arg_feed.py. 
- And test_feedstr.py -> test_arg_feed.py

We also want to do that for the other arg building blocks. (The github issue mentions this.)
- timeframestr.py -> arg_timeframe.py with ArgTimeframe. Change test file accordingly. There's a "Timeframe" class there too, you can either keep it there or have its own module, I'm fine either way
- pairstr.py -> arg_pair.py / ArgPair. Test file too.
- exchangestr.py -> arg_exchange.py / ArgExchange. Test file too.

### 2. Changing **subgraph** modules / classes / tests:
 - You've renamed util/feed.py::Feed to SubgraphFeed. Great!
  - You still need to change its module from feedstr.py -> subgraph_feed.py. 
  - And test_feedstr.py -> test_subgraph_feed.py

### 3. Let's rearrange the location of some files. 

We have three different groups of utility-ish (non-bot) class modules: subgraph classes, contract classes, and CLI & args classes. They're currently spread across util/ and models/
- so let's target 3 directories: "subgraph", "contract/", "cli/". How: we already have "subgraph/"; rename "models" -> "contract", and create "cli/"
- In "subgraph/", move recently-renamed subgraph*.py files from "models/". Move unit tests too ofc.
- In "cli/": move util/cli_arguments.py, util/cli_module.py, and all recently-renamed arg*.py files there. Move unit tests too.
- Then "models/" should be in good shape too. But double-check.

### 4. Q: Should I support feeds without signals, already?

Yes.

(From a Q from Calina in this PR's comments section

### Final comment

Like mentioned at top: I think this PR is good to go. Better to handle the feedback in follow-up PR(s). 

Therefore I am approving.

trentmc (approved):

PR #467, opened 2023-12-20
Move and rename utils

calina-c:
Follow-up from https://github.com/oceanprotocol/pdr-backend/pull/464.

Includes moving and renaming. Does not include refactoring for timestr, pairstr, exchangestr.


trentmc (changes requested):
.

trentmc (approved):

PR #469, opened 2023-12-21
[Lake] integrate pdr_subscriptions into GQL Data Factory

### LINKED ISSUE #468: [Lake] Integrate pdr_subscriptions into GQL Data Factory
### Motivation
Subscriptions are not being fetched from subgraph and added to lake. There is similar logic being used to determine consumptions for df_buyer, but the requirements are different.

Instead, simply use predictSubscriptions w/ a top-level timestamp filter to fetch all subscriptions between the expected [st_ut, fin_ut].

When implementing this, we'll have 2 different subgraph queries being handled by GQL Data Factory, using the same date-range. Due to subgraphs tables having records (or not), the Data Factory should elegantly handle empty gql quries, empty parquet files, and it's basic interface of get_dfs() without any issues. 

### DoD:
- [x] Add fn to fetch subscriptions from subgraph
- [x] Create subscription object for models
- [x] Create subscription schema for lake
- [x] Hook subscriptions to lake
- [x] Add mock utils for subscriptions
- [x] Add tests that verify that subscription objects, subgraph, and components are working as expected
- [x] Add tests that verify that GQL Data Factory that handle empty subgraph fetches + empty parquets files

idiom-bytes:
Fixes #468


pdr_backend/contract/subscription.py
```diff
@@ -0,0 +1,137 @@
+from typing import List
```
trentmc, line 1: `contract/` directory is only for classes that are wrappers of smart contracts.

subscription.py is not that.

 To solve: move subscription.py to `lake/`.

Second: where is `test_subscription.py`?

trentmc (changes requested):

pdr_backend/contract/subscription.py
idiom-bytes: wrt `subscrptions.py`
we used to have `models/`  => `contracts/`
so `predictions.py` should also be moved to `lake/`

we also have `subgraph/` files referencing these objects
such as `subgraph_predictions.py`

should we perhaps move these to lake too?
i.e. all lake-related `subgraph/models/mocks/testing/` etc... inside the same module.

if so, then can you perhaps recommend a structure for this? Here a very simple, but perhaps redundant structure... I think your "line-in-the-sand" feedback would be useful.
`lake/`
`lake/models/`
`lake/subgraph/`

trentmc: Please no hierarchy for now. (Remember, every sub-directory you add also means another sub-sub directory for tests. It's a huge pain to manage. Eg even with predictoor/approach1 and 3, and trader/approach1 and 2. Super annoying to jump around. So, for now, keep the rest flat.)

"lake" is higher-level than subgraph.

Given that subgraph uses `subscriptions.py` (I missed that), then I recommend: move `subscription.py` to `subgraph/` too.

idiom-bytes: Implemented test_subscription, moved `subscriptions.py` and `predictions.py`, expanded tests, and added logic such that we don't have extra calls being made to subgraph when they aren't needed.

trentmc (approved):

pdr_backend/lake/plutil.py
```diff
@@ -177,3 +177,16 @@ def text_to_df(s: str) -> pl.DataFrame:
     df = pl.scan_csv(filename, separator="|").collect()
     shutil.rmtree(tmpdir)
     return df
+
+
```
kdetry, line 181: we can add @enforce_types

PR #475, opened 2024-01-02
Improve DRY

calina-c:


pdr_backend/ppss/abstract/base_ss.py
```diff
@@ -0,0 +1,158 @@
...
+
+class MultiFeedSS(ABC):
+    FEEDS_KEY = ""
+
+    @enforce_types
+    def __init__(self, d: dict, assert_feed_attributes: Optional[List] = None):
+        assert self.__class__.FEEDS_KEY
+        self.d = d
+        # save self.exchs_dict
+        self.exchs_dict: dict = {}  # e.g. {"binance" : ccxt.binance()}
```
trentmc, line 19: There is no need for most ss classes to know anything about ccxt exchanges. The ccxt info only be seen by ohlcv data factory and maybe its ss.

pdr_backend/ppss/abstract/base_ss.py
```diff
@@ -0,0 +1,158 @@
...
+from typing import Dict, List, Optional, Set, Tuple, Union
+
+from enforce_typing import enforce_types
+
+from pdr_backend.cli.arg_feed import ArgFeed, ArgFeeds
+from pdr_backend.cli.arg_pair import ArgPair
+from pdr_backend.subgraph.subgraph_feed import SubgraphFeed
+
+
+class MultiFeedSS(ABC):
```
trentmc, line 11: I'd prefer if this class was labeled as, and treated more like a mixin. Where other ss classes mix in a slice of info about feeds. Therefore it does not need to be abstract.

Same thing fir the single feed class below.

pdr_backend/ppss/aimodel_ss.py
```diff
@@ -1,28 +1,21 @@
 import copy
-from typing import List, Set, Tuple
 
 import numpy as np
 from enforce_typing import enforce_types
 
-from pdr_backend.cli.arg_feed import ArgFeeds
```
trentmc, line 7: This should inherit both the multi feed ss mixin, ache the single feed ss mixin. (Single feed should not be at predictoor ss level.)

pdr_backend/ppss/abstract/base_ss.py
```diff
@@ -0,0 +1,158 @@
...
+    @property
+    def exchange_strs(self) -> List[str]:
+        return sorted(self.exchs_dict.keys())
+
+    @property
+    def n_input_feeds(self) -> int:
+        return len(self.input_feeds)
+
+    @property
+    def input_feeds(self) -> ArgFeeds:
```
trentmc, line 49: This should not be "input_feeds" for all ss's; that label is only appropriate for ai_model_ss. I would recommend to call it simply "feeds". 

Similar with "predict feed" for single feed ss. "Predict" is only appropriate for ai_model_ss. So just call it "feed".

pdr_backend/ppss/abstract/base_ss.py
```diff
@@ -0,0 +1,158 @@
+from abc import ABC
+from typing import Dict, List, Optional, Set, Tuple, Union
+
```
trentmc, line 3: Need to add a test module, ie test_base_ss.py.

trentmc (changes requested):

trentmc (approved):

PR #491, opened 2024-01-08
For #465 into main: Implement a delay to pause and retry in case of network errors

### LINKED ISSUE #465: [YAML] pdr bots shouldn't stop if RPC goes down briefly
### Background

Reported by @idiom-bytes [in slack](https://oceanprotocol.slack.com/archives/C05P3TBHUAC/p1702933904126029):

Hi, our web3 middleware that talks to the sapphire RPC is raising exceptions while trying to get blocks. The exception is raised after the # target_retries.

This is causing pdr-bots to throw an exception => causing bots to stop running => causing users to have to maintain/ops the bots. 

image

### DoD

I think in situations where we have a known exception (subgraph/rpc that might self-heal in 1-2 mins) that we could wrap an outer-loop wait such that the bot doesn't crash, forcing botters to monitor/ops the bots.

I.e. "just set it, and forget it"

trizin:
Fixes #465 for main branch. (Need to also do for yaml-cli2 branch.)

Changes proposed in this PR:

- Implement a delay to pause and retry in case of network errors in `get_block`

### DIFF
--- a/pdr_backend/util/web3_config.py
+++ b/pdr_backend/util/web3_config.py
@@ -1,3 +1,5 @@
+import time
+
 from typing import Optional
 
 from enforce_typing import enforce_types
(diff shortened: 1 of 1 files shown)


trentmc (approved):
Please also get this change into yaml-cli2 branch.

PR #496, opened 2024-01-09
[FROZEN-DO NOT MERGE] Fix #480: Add TrueVal to gql data factory

### LINKED ISSUE #480: [Lake] integrate pdr_truevals into GQL Data Factory
### Motivation
TrueVals are not being fetched from subgraph and added to lake. There is similar logic being used to determine predictions for predictoor, but we need to figure out if the trueVal is available or not.

Use predictTruevals w/ a top-level timestamp filter to fetch all trueVals between the expected [st_ut, fin_ut].

### DoD:
- [ ] Improve gql_data_factory so it can handle empty fetches, empty parquet files, and general mis-matches between fs.
- [x] Add fn to fetch truevals from subgraph
- [x] Create truevals object for models
- [x] Create truevals schema for lake
- [x] Hook truevals to lake
- [x] Add mock utils for truevals
- [x] Add tests that verify that truevals objects, subgraph, and components are working as expected

KatunaNorbert:
Fixes #480.


idiom-bytes:
@kdetry i'm helping to get this across the finish line
- [x] cherry-picked commits from another branch that handles errors w/ subgraph
- [x] Updated fetch logic to loop/fetch all records and to handle errors
- [x] Updated remaining fetch logic to follow latest implementations (such as changing timestamps to ms, and sorting by timestamp before returning the df)
- [x] Cleaned up logic to follow latest implementations and verified tests are working.

idiom-bytes:
Analytics + Lake + Other places had their functionality updated to use timestamp correctly.

idiom-bytes:
@kdetry @KatunaNorbert i have made a significant amount of updates in here.
**I want to freeze this PR, so we have part 1/3 done.**

@trizin, can you provide feedback on the cli_module part?
I'm guessing we should use nested_override_args to override `lake_ss.st_timestamp` + `lake_ss.fin_timestamp` w/ [args.ST + args.END]?
image

idiom-bytes (approved):
I believe that this PR is in a good place for part 1/3 to be done.

idiom-bytes:
Great work @KatunaNorbert! Closing PR.

PR #523, opened 2024-01-12
Issue512 add helper functions to manage accounts

### LINKED ISSUE #512: [YAML] Add helper functions to manage accounts and predictoors
### Motivation
As part of deploying and managing predictoors, it would improve the UX by having functions that make this easier.

The following cli functions are to be added:
- create_accounts => creates 1->N accounts. prints pk + pub_key
- view_account_balances => shows OCEAN + NATIVE balances for each account submitted
- fund_accounts => distributes n_amount across "addresses,..." in either OCEAN or NATIVE tokens

### DoD
- [ ] cli module, mixins, functions, and utils are created to support 
- [ ] add helper text and descriptions to the cli module
- [ ] tests are written
- [ ] create_accounts is working as expected
- [ ] view_account_balances is working as expected
- [ ] fund_accounts is working as expected

idiom-bytes:
Fixes #512


pdr_backend/cli/cli_arguments.py
```diff
@@ -200,6 +240,58 @@ def __init__(self, description: str, command_name: str):
...
+class _ArgParser_NUM_PPSS_NETWORK(
+    CustomArgParser,
+    NUM_Mixin,
+    PPSS_Mixin,
+    NETWORK_Mixin,
+):  # pylint: disable=too-many-ancestors
+    @enforce_types
+    def __init__(self, description: str, command_name: str):
+        super().__init__(description=description)
+        self.add_argument("command", choices=[command_name])
```
calina-c, line 253: please use add_arguments_bulk instead, it is **much** more concise. Applies to the other parsers as well

pdr_backend/cli/cli_arguments.py
```diff
@@ -112,6 +115,43 @@ def add_argument_LOOKBACK(self):
         )
 
 
+@enforce_types
+class NUM_Mixin:
+    def add_argument_NUM(self):
+        self.add_argument("NUM", type=int, help="1|2|..")
```
calina-c, line 121: I think you can leave the default help from "int" type. If you want this to be positive, you can also implement a custom type function or validate it when running the command.

Ref here: https://stackoverflow.com/a/14117511

pdr_backend/cli/cli_arguments.py
```diff
@@ -112,6 +115,43 @@ def add_argument_LOOKBACK(self):
...
+
+
+@enforce_types
+class ACCOUNTS_Mixin:
+    def add_argument_ACCOUNTS(self):
+        self.add_argument(
+            "ACCOUNTS",
+            type=str,
```
calina-c, line 129: Similar to the stackoverflow I pasted for int types, please create a custom "csv_accounts" type that validates the account string. You can also find inspiration in dftool, where the functions are try/except blocks. The point is: by the time these arguments reach the command itself, they should be correct. It helps with UX a lot to get the error from the help/argparse part instead of looking for the failure inside the command

pdr_backend/cli/cli_arguments.py
```diff
@@ -112,6 +115,43 @@ def add_argument_LOOKBACK(self):
...
+        )
+
+
+@enforce_types
+class TOKEN_AMOUNT_Mixin:
+    def add_argument_TOKEN_AMOUNT(self):
+        self.add_argument(
+            "TOKEN_AMOUNT",
+            type=float,
+            help="Amount of token to send to each address",
```
calina-c, line 140: please specify in the help string that this value is not in wei

pdr_backend/cli/cli_arguments.py
```diff
@@ -112,6 +115,43 @@ def add_argument_LOOKBACK(self):
...
+
+
+@enforce_types
+class NATIVE_TOKEN_Mixin:
+    def add_argument_NATIVE_TOKEN(self):
+        self.add_argument(
+            "--NATIVE_TOKEN",
+            action="store_true",
+            default=False,
+            help="Is native token or OCEAN?",
```
calina-c, line 151: please make sure the help and type of these indicate it is a boolean. It is unclear to me if both native and ocean give "True". I suggest something like one of these a help string:
- "True if native token or OCEAN, False otherwise. Defaults to false"
- "True if native token, as opposed to OCEAN. Defaults to false (OCEAN)"

pdr_backend/cli/cli_arguments.py
calina-c: also, please add tests for each of these in test_cli_arguments.py

pdr_backend/util/web3_accounts.py
```diff
@@ -0,0 +1,98 @@
+import sys
```
calina-c, line 1: if you add a new file, please add its corresponding `test_` file and fill it out, as well.

pdr_backend/util/web3_accounts.py
```diff
@@ -0,0 +1,98 @@
...
+
+@enforce_types
+def view_accounts(addresses: List[str], web3_pp: Web3PP):
+    """
+    @description
+        View account balances for multiple addresses
+    """
+    # loop through n_wallets
+    for address in addresses:
+        checksum_address = web3_pp.web3_config.w3.to_checksum_address(address.lower())
```
calina-c, line 53: I would also add this check in the parse arguments part. By this point, the addresses should be definitely checksummable, or preferably already converted (checksummed).

calina-c: this would also remove the need for _get_account_balances as a separate function. Another advantage is that you only need to get the OCEAN_addr once (before the loop), while now you retrieve it for *each* address

pdr_backend/util/web3_accounts.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+
+        print(f"Account {checksum_address}")
+        print(f"Native token balance: {from_wei(native_token_balance)}")
+        print(f"OCEAN balance: {from_wei(OCEAN_balance)}")
+
+
+@enforce_types
+def fund_accounts(
+    amount: float, to_addressess: List[str], web3_pp: Web3PP, is_native_token: bool
```
calina-c, line 65: I think there's a typo here (an extra "s" in "to_addressess")

pdr_backend/util/web3_accounts.py
```diff
@@ -0,0 +1,98 @@
...
+):
+    """
+    @description
+        Fund multiple accounts using native or OCEAN tokens.
+        Sends a total of (Amount * n_accounts).
+    """
+    if web3_pp.network not in ["sapphire-testnet", "sapphire-mainnet"]:
+        print("Unknown network")
```
calina-c, line 73: please specify the network, i.e. print(f"Unknown network {network}")

pdr_backend/util/web3_accounts.py
```diff
@@ -0,0 +1,98 @@
...
+        sys.exit(1)
+
+    token = None
+    token_name = None
+    if is_native_token:
```
calina-c, line 78: I suggest adding a name property as a helper in NativeToken and OceanToken classes. I think it can be useful in the future.

Then:
- the if only decides "a single thing"
- later in print f"Sending {token.name} token to {address} for the amount of {amount} wei"

pdr_backend/util/web3_accounts.py
```diff
@@ -0,0 +1,98 @@
...
+        token = NativeToken(web3_pp)
+        token_name = "ROSE"
+    else:
+        OCEAN_addr = get_address(web3_pp, "Ocean")
+        token = Token(web3_pp, OCEAN_addr)
+        token_name = "OCEAN"
+
+    private_key = getenv("PRIVATE_KEY")
```
calina-c, line 86: I think this can be moved to the very top. No point in doing the other statements if there is no private key.

pdr_backend/util/web3_accounts.py
```diff
@@ -0,0 +1,98 @@
...
+    assert private_key is not None, "Need PRIVATE_KEY env var"
+
+    account = Account.from_key(private_key)  # pylint: disable=no-value-for-parameter
+
+    for address in to_addressess:
+        print(f"Sending {token_name} token to {address} for the amount of {amount} wei")
```
calina-c, line 92: Another thing that could work, depending on your preferences. Add a verbose: bool = False parameter to transfer. Then you can do the prints there, if verbose is True. This is an alternative to what I suggested before with `token.name`

calina-c (changes requested):

pdr_backend/cli/cli_arguments.py
idiom-bytes: ```also, please add tests for each of these in test_cli_arguments.py```

I see that `defined_parsers` and `test_arg_parser` would already cover these.

pdr_backend/cli/cli_arguments.py
idiom-bytes: good work, done

pdr_backend/cli/cli_arguments.py
idiom-bytes: Great feedback. I've updated the help comments. However, `store_true` & `default` is how you use booleans in python argparse. So, I kept it implemented like this, with the new comments and usage in place.

Here is the problem w/ using `type=bool` as described by SO
```
What I find misleading is that you can set type=bool, you get no error message, and yet, for both "False" and "True" string arguments, you get True in your supposedly boolean variable (due to how type casting works in python). So either type=bool should be clearly unsupported (emit some warning, error, etc.), or it should work in a way that is useful and intuitively expected.
```

https://stackoverflow.com/questions/15008758/parsing-boolean-values-with-argparse

pdr_backend/util/web3_accounts.py
idiom-bytes: in progress

please note that this PR is in draft, and not ready for review

pdr_backend/util/web3_accounts.py
idiom-bytes:
```
I would also add this check in the parse arguments part. By this point, the addresses should be definitely checksummable, or preferably already converted (checksummed).
```

The current to_checksum_address requires PPSS object which isn't instantiated in the argparse/cli_arguments module at the validation phase.

We can do this by importing `eth_utils.to_checksum_address` inside cli_arguments.py and just validate it directly.

I agree that it's easier to checksum at the entry points and let logic cascade down, but I'm not sure this has been addressed at every entrypoint. I'll update this.

idiom-bytes:
```
this would also remove the need for _get_account_balances as a separate function. Another advantage is that you only need to get the OCEAN_addr once (before the loop), while now you retrieve it for each address
```

The checksum() and get_balances() calls are unrelated. I'll modify this a bit so there are less redundant calls.

pdr_backend/util/web3_accounts.py
idiom-bytes: Updated

pdr_backend/util/web3_accounts.py
idiom-bytes: Done, but perhaps, similar to checksum_address above being done at the argparse level, similarly cli_module functions that rely on PRIVATE_KEY should perhaps do this assert upstream such that everything below can just use PRIVATE_KEY.

pdr_backend/util/web3_accounts.py
idiom-bytes: Added the previous implementation, resolving this conversation.

pdr_backend/cli/cli_arguments.py
```diff
@@ -112,6 +133,51 @@ def add_argument_LOOKBACK(self):
         )
 
 
+@enforce_types
+def check_positive(value):
+    ivalue = int(value)
```
calina-c, line 138: can you also wrap this line in a try/catch? If I input something that can't be converted to int, the type error will not be raised :)

pdr_backend/cli/cli_module.py
```diff
@@ -163,3 +164,22 @@ def do_publisher(args):
...
+
+@enforce_types
+def do_create_accounts(args):
+    create_accounts(args.NUM)
+
+
+@enforce_types
+def do_view_accounts(args):
+    ppss = PPSS(yaml_filename=args.PPSS_FILE, network=args.NETWORK)
+    accounts = args.ACCOUNTS.split(",")
```
calina-c, line 177: hmm... I think by this point they are already converted to a list, right?

pdr_backend/util/web3_accounts.py
```diff
@@ -0,0 +1,81 @@
...
+    OCEAN_addr = get_address(web3_pp, "Ocean")
+    OCEAN_token = Token(web3_pp, OCEAN_addr, "OCEAN")
+
+    # loop through all addresses and print balances
+    for address in addresses:
+        native_token_balance = native_token.balanceOf(address)
+        OCEAN_balance = OCEAN_token.balanceOf(address)
+
+        print(f"\nAccount {address}")
+        print(f"Native token balance: {from_wei(native_token_balance)}")
```
calina-c, line 45: I think for both of these you can print "wei" after the amount

pdr_backend/contract/token.py
```diff
@@ -6,8 +7,8 @@
 
 @enforce_types
 class Token(BaseContract):
-    def __init__(self, web3_pp, address: str):
-        super().__init__(web3_pp, address, "ERC20Template3")
+    def __init__(self, web3_pp, address: str, name: Optional[str] = "ERC20"):
```
calina-c, line 10: I would just keep the property, I don't think we need to change this part.

pdr_backend/cli/cli_arguments.py
```diff
@@ -77,12 +82,28 @@ def add_argument_NETWORK(self):
...
+@enforce_types
+def check_addresses(value):
+    """
+    @description
+        validates that all addressses is a comma-separated list of strings
+        each string, will be a valid Ethereum address
+    """
+    addresses = value.split(",")
+    checksummed_addresses = []
+    for address in addresses:
```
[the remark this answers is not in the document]
idiom-bytes: ty, i should have added the try/catch around the checksum

pdr_backend/util/web3_accounts.py
idiom-bytes: Added checksum and cleaned up calls inside `web3_accounts.py`

pdr_backend/cli/cli_arguments.py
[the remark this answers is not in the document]
idiom-bytes: Thank you, done.

Also pointed other cli_module functions that use addresses to do all of this validation/testing as well. Updated tests wherever appropriate as well such that all addresses that are passed in via address_str are taken through the same process.

pdr_backend/cli/cli_arguments.py
idiom-bytes: wrapped

pdr_backend/cli/cli_module.py
idiom-bytes: yes, i've updated all of this and a lot more

check other cli_arguments implementations and modules that were using 0x addresses, and are all passed in via an `address_str` structure

pdr_backend/util/web3_accounts.py
idiom-bytes: Please note that this is not printing in wei (1e0), but in ether (1e18).

**On that note, I would propose we identify a base unit of measurement to report everything in a standard way.** For reports, I'd really like to see everything reported in ether so we can see a meaningful decimal. All other commands right now are reporting the numbers in ether (1e18)

What you're suggesting
image

How it looks right now
image

Personally, I find the wei numbers useless because there are no meaningful decimals, and no one can count that many units. On the other side I don't recommend we say: "OCEAN Balance: 103.45 eth".

Again, I'd prefer that we simply state somewhere in the cli/apps that all numbers are reported in 1e18, and follow suit.

Cheers!

Issue #507, opened 2024-01-10
[YAML] Several async-related warnings in pdr_backend/trader


trentmc:
### Background / motivation

(In yaml-cli2 branch.)

From [this log](https://github.com/oceanprotocol/pdr-backend/actions/runs/7472573608/job/20335037499), which was a test in a yaml-cli2 subbranch that made a trivial change.

```text
=============================== warnings summary ===============================
pdr_backend/trader/approach1/test/test_trader_agent1.py::test_trader_agent1_do_trade
  pdr_backend/trader/approach1/test/test_trader_agent1.py:26: PytestWarning: The test <Function test_trader_agent1_do_trade> is marked with '@pytest.mark.asyncio' but it is not an async function. Please remove the asyncio mark. If the test is not marked explicitly, check for global marks applied via 'pytestmark'.
    @enforce_types

pdr_backend/trader/approach1/test/test_trader_agent1.py::test_trader_agent1_do_trade
pdr_backend/trader/approach2/test/test_trader_agent2.py::test_trader_agent2_do_trade
pdr_backend/trader/test/test_base_trader_agent.py::test_trader_agent_take_step
  /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/_pytest/python.py:183: PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.
  You need to install a suitable plugin for your async framework, for example:
    - anyio
    - pytest-asyncio
    - pytest-tornasync
    - pytest-trio
    - pytest-twisted
    warnings.warn(PytestUnhandledCoroutineWarning(msg.format(nodeid)))

pdr_backend/trader/approach2/test/test_trader_agent2.py::test_trader_agent2_constructor
  /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/yaml/tokens.py:4: RuntimeWarning: coroutine 'test_trader_agent1_do_trade' was never awaited
    self.start_mark = start_mark
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.

pdr_backend/trader/approach2/test/test_trader_agent2.py::test_trader_agent2_do_trade
[...]

PR #531, opened 2024-01-13
#507: fix asyncio issues

calina-c:
Work on #507.
Asyncio has some known issues, per their changelog. Namely issues with fixture handling etc., which I believe causes the warnings and test skips in our runs. They recommend using the previous version until they are fixed. It is also why my setup didn't spew up any warnings, my asyncio version was 21.1.

https://pytest-asyncio.readthedocs.io/en/latest/reference/changelog.html

I am wondering if I should mark #507 as closed or not. I do suggest keeping this around if newer versions of asyncio will work as expected, and we can remove the strict version setting in setup.py. Thoughts, @trentmc?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ install_requirements = [
     "pyarrow",
     "pylint",
     "pytest",
-    "pytest-asyncio",
+    "pytest-asyncio==0.21.1",
     "pytest-env",
     "pyyaml",
     "requests",


trentmc (approved):

trentmc:
> I am wondering if I should mark https://github.com/oceanprotocol/pdr-backend/issues/507 as closed or not

It's fine to close. No need to have the mental overhead to manage. 

I just merged this PR into the yaml-cli2 branch (given weekend, to keep things moving). I will also close the issue.

PR #535, opened 2024-01-15
[Lake] Issue-481: Adding Payouts to the Data Factory

### LINKED ISSUE #481: [Lake] integrate pdr_payouts into GQL Data Factory
### Motivation
Payouts are not being fetched from subgraph and added to lake. There is similar logic being used to determine predictions for predictoor, but we need to figure out if the payouts are available or not.

Use predictPayouts w/ a top-level timestamp filter to fetch all payouts between the expected [st_ut, fin_ut].

Please review the event id in subgraph/contract to identify how it's build. I think in the case of payouts it's {contract-slot-user}.

### DoD:
- [x] Add fn to fetch payouts from subgraph
- [x] Create payouts object for models
- [x] Create payouts schema for lake
- [x] Hook payouts to lake
- [x] Add mock utils for payouts
- [x] Add tests that verify that payouts objects, subgraph, and components are working as expected

kdetry:
Fixes #481 

Changes proposed in this PR:

- Added Payouts schema
- Created fetching process for payouts
-  Added fn to fetch payouts from subgraph
- Added mocks and tests


pdr_backend/subgraph/subgraph_payout.py
```diff
@@ -0,0 +1,143 @@
...
+        where_query_arr.append(
+            """
+                {
+                    timestamp_gte: %s,
+                    timestamp_lte: %s,
+                    prediction_contains: "%s"
+                }
+            """
+            % (start_ts, end_ts, asset_id)
+        )
```
idiom-bytes, line 39: I'm not sure why this has to be looped...

I.e. we should be able to use a single where clause
image

That generates the following query for multiple assets
```
query {
            predictPayouts (
            first: 10
            skip: 1
            where: {
        {
            timestamp_gte: 1622547000,
            timestamp_lte: 1622548800,
            prediction_contains: ['0x18f54cc21b7a2fdd011bea06bba7801b280e3151', '0x33334cc21b7a2fdd011bea06bba7801b280e3151']
        }
    }
        ) {
                id
                timestamp
                payout
                prediction {
                    user {
                        id
                    }
                    slot {
                        id
                        predictContract{
                            id
                            token{
                                name
                            }
                        }
                    }
                }
            }
        }
```

as defined in this test example

image

pdr_backend/subgraph/subgraph_payout.py
```diff
@@ -0,0 +1,143 @@
...
+        Payout(
+            **{
+                "payout": float(payout["payout"]),
+                "user": payout["prediction"]["user"]["id"],
+                "timestamp": payout["timestamp"],
+                "ID": payout["id"],
+                "token": payout["prediction"]["slot"]["predictContract"]["token"][
+                    "name"
+                ],
+                "slot": int(payout["id"].split("-")[1]),
```
idiom-bytes, line 138: Please make sure that we're retrieving all required data to generate the insights we want.

Example: Payout is where [predictedValue + prediction.slot.revenue + prediction.stake + prediction.slot.roundSumStakesUp + prediction.slot.roundSumStakes] are provided. So, we need to fetch this additional data such that we can update our local records.

We can then use the data from the payout event to update our lake/ `pdr_predictions` and `pdr_slots` tables.

```
query{  
  predictPayouts(...){
    prediction{
      slot {
        id
        predictContract {
          id
        }
        slot
        status
        revenue
        roundSumStakesUp
        roundSumStakes
      }
      user {
        id
      }
      stake
    }
    payout
    predictedValue
    trueValue
    timestamp
  }
}```

File: pdr_backend/subgraph/subgraph_payout.py (hunk 1 above)
Comment (kdetry) [reply pdr_backend/subgraph/subgraph_payout.py:39]: I took another look at your comment, @idiom-bytes. The "aaset_ids" argument is a list of strings that represent the IDs of the contracts we want to retrieve. Unfortunately, the "predictPayouts" query's "where" clause does not have a contract filter. We could use a nested query, like the one below, but our subgraph system does not allow it:

```
where: {
    prediction_: {
        slot_: {
            predictContract_in: ["0xfeed1", "0xfeed2"]
        }
    }
}
```

However, the "prediction_contains" argument searches for text within prediction IDs. It only allows strings. The structure of a prediction ID is as follows:

```
{contract address}-{slot}-{user}
```

Therefore, if the contract address is present in the prediction ID, we can retrieve it.

File: pdr_backend/subgraph/subgraph_payout.py (hunk 2 above)
Comment (kdetry) [reply pdr_backend/subgraph/subgraph_payout.py:138]: https://github.com/oceanprotocol/pdr-backend/issues/447
I have followed the BRONZE_FOR_PAYOUTS table on the epic so I didn't add these values.

I am going to add them asap.

File: pdr_backend/lake/gql_data_factory.py [hunk 3]
```diff
@@ -43,6 +47,10 @@ def __init__(self, ppss: PPSS):
         )
         contract_list = [f.lower() for f in contract_list]
 
+        # For debugging
+        # t_contract_list = [f.lower() for f in contract_list]
+        # contract_list = [t_contract_list[0], t_contract_list[1]]
```
KatunaNorbert, line 52: Do we need to keep this commented lines?

pdr_backend/subgraph/subgraph_payout.py
KatunaNorbert: Looked into this and I agree with @kdetry, we can't just pass the list of addresses with the way subgraph it's structured right now

kdetry:
I close this PR, it is replaced with the following ne:
https://github.com/oceanprotocol/pdr-backend/pull/559

PR #539, opened 2024-01-15
#538 - check network main subgraph query fails

### LINKED ISSUE #538: `check_network_main` subgraph query fails
**Describe the bug**
The subgraph query in the `check_network_main` fails with the following error from subgraph:
```
pdr check_network: Begin
Arguments:
PPSS_FILE=ppss.yaml
NETWORK=sapphire-mainnet
LOOKBACK_HOURS=24
{'errors': [{'locations': [{'column': 0, 'line': 0}], 'message': 'Invalid value provided for argument `where`: Object({"slot_gt": Int(Number(226855931)), "slot_lt": Int(Number(226942331))})'}]}
```

**To Reproduce**
On the yaml-cli2 branch, simply execute:
```
./pdr check_network ppss.yaml sapphire-mainnet
```

trizin:
Fixes #538

Changes proposed in this PR:

- Use UTC unix time in seconds for the query instead of ms

### DIFF
--- a/pdr_backend/analytics/test/test_check_network.py
+++ b/pdr_backend/analytics/test/test_check_network.py
@@ -31,7 +31,7 @@ MOCK_CUR_UT = 1702826080
     side_effect=Mock(return_value=MOCK_CUR_UT),
 )
 def test_check_dfbuyer(  # pylint: disable=unused-argument
-    mock_current_ut,
+    mock_current_ut_ms,
     mock_get_expected_consume_,
     mock_get_consume_so_far_per_contract_,
     capsys,
--- a/pdr_backend/lake/gql_data_factory.py
+++ b/pdr_backend/lake/gql_data_factory.py
@@ -16,7 +16,7 @@ from pdr_backend.lake.table_pdr_subscriptions import (
 from pdr_backend.ppss.ppss import PPSS
 from pdr_backend.subgraph.subgraph_predictions import get_all_contract_ids_by_owner
 from pdr_backend.util.networkutil import get_sapphire_postfix
-from pdr_backend.util.timeutil import current_ut, pretty_timestr
+from pdr_backend.util.timeutil import current_ut_ms, pretty_timestr
 
 
 @enforce_types
@@ -113,7 +113,7 @@ class GQLDataFactory:
 
             st_ut = self._calc_start_ut(filename)
             print(f"      Aim to fetch data from start time: {pretty_timestr(st_ut)}")
-            if st_ut > min(current_ut(), fin_ut):
+            if st_ut > min(current_ut_ms(), fin_ut):
                 print("      Given start time, no data to gather. Exit.")
                 continue
 
--- a/pdr_backend/lake/ohlcv_data_factory.py
+++ b/pdr_backend/lake/ohlcv_data_factory.py
@@ -22,7 +22,7 @@ from pdr_backend.lake.plutil import (
     save_rawohlcv_file,
 )
 from pdr_backend.ppss.lake_ss import LakeSS
-from pdr_backend.util.timeutil import current_ut, pretty_timestr
+from pdr_backend.util.timeutil import current_ut_ms, pretty_timestr
 
 
 @enforce_types
@@ -114,7 +114,7 @@ class OhlcvDataFactory:
         assert feed.timeframe
         st_ut = self._calc_start_ut_maybe_delete(feed.timeframe, filename)
         print(f"      Aim to fetch data from start time: {pretty_timestr(st_ut)}")
-        if st_ut > min(current_ut(), fin_ut):
+        if st_ut > min(current_ut_ms(), fin_ut):
             print("      Given start time, no data to gather. Exit.")
             return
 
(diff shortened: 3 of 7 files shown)


trentmc (changes requested):
We need to be uniform in our treatment of time. 

current_ut() was introduced to fix previous bugs:  to make sure there is consistency across time zones, and to ensure consistent use of ms vs s. Same for all "ut" expressions.

Therefore please update these changes to keep using current_ut(). You l will probably need to go deeper to find a more subtle bug. 

Please also make sure that a test catches the bug.

trentmc:
I updated my earlier comment to specify exactly what needs doing, in order to finish this PR and close it.

trentmc (approved):

trentmc:
Note: from my comment above, we had a slack discussion [here](https://oceanprotocol.slack.com/archives/C05P3TBHUAC/p1705337582348999). I'm copying here, for posterity.

**Berkay:** (written before he saw my comment above)

While working on the check_network issue I noticed that a function named `current_ut` is being used to get the unix time for the subgraph query. However, it returns the time in milliseconds, whereas the chain and subgraph uses "seconds".

I've updated both the check_network and check_dfbuyer functions to use unix time in seconds. The current_ut function is also used in a few places within analytics, so we should probably verify if those usages are accurate. By the way, getting Unix time through time.time() is prettier than using the datetime module. btw (2) the standard unit for unix time is seconds so we might wanna make the millisecond part more clear in the code

**Roberto:** (written before he saw my comment above)

I had to navigate the ms-sec subgraph aspects as I believe this function was derived from requirements in ohlcv. I created some helper functions to go between secs and ms, such that the implementations were cleaner.

In the end, I ended up using ms everywhere in the data and seconds in the subgraph, as I assumed that ms was the time format we wanted to maintain locally.

**Trent:** 

We need to be uniform in our treatment of time.

- A naked `time.time()` on its own doesn't explicitly acknowledge whether using timezone, or seconds vs ms. We need to make both more clear. This caused bugs.
- `current_ut()` was introduced to fix those bugs: to make sure there is consistency across time zones, and to ensure consistent use of ms vs s. Its docstring says so: "Return the current date/time as a unix time (int in # ms)"
- Everywhere in the code, we want expressions with "ut" to be ms. Eg `lake/ohlcv_data_factory.py` class constructor docstring has it [[Ref](https://github.com/oceanprotocol/pdr-backend/blob/0b411c2978bdf09293b38a7ce8b4574b5f8eb152/pdr_backend/lake/ohlcv_data_factory.py#L55)].
- 

I gave ^ as a [comment](https://github.com/oceanprotocol/pdr-backend/pull/539#pullrequestreview-1822083582) in the PR that you (Berkay) referred to. Along with further comments on specific changes for that PR.

**Trent too:** 

> getting Unix time through time.time() is prettier than using the datetime module
FYI time.time() doesn't properly support timezones, iirc. Or at least not cleanly. That's why I'd used datetime.

But rather than using either directly, we should use current_ut(),  which then calls whatever code it needs to accomplish its task (time.time or datetime).

"Correct" / "not buggy" / "not at risk of being buggy" is  more important than "pretty".

**Berkay:**

Just read your comment on Github. To clarify, we don't need to support other time zones for the chain & subgraph, time.time returns the utc time, using the local time might cause bugs. In this case, we can just divide the returned value by 1000 to convert to seconds, does that sound ok?

**Trent:**

I really prefer that we use "current_ut()" everywhere, because then it's not subject to quirks of different time libraries, and we have thorough testing for it. We used to have many bugs around this. The tests keep those bugs at bay, as long as we use "current_ut()". I don't care how "current_ut()" is implemented, as long as it respects UTC and returns ms (not s), and passes the tests of course! If you want to use time.time() instead of datetime in its implementation, go ahead.

**Trent too:**

BTW I saw that in your recent [commit](https://github.com/oceanprotocol/pdr-backend/commit/16c3d49021c5aa2217e7b5e4eef3aed01701ed15) you had set `cur_ut = current_ut() / 1000`. This is better, but not quite there, because there's still a a variable with postfix "ut" (implying ms) that's got numbers in s. That's against the naming convention of "ut". Instead, I recommend to rename `cur_ut` (to e.g. `cur_ut_s`), or make cur_ut store in ms and then do JIT conversion.  The first option is probably less error prone.

Another idea: whenever we use `ut` we always explicitly say `ut_ms` or `ut_s`. And we could have `current_ut_ms()` and `current_ut_s()`. If we did this, we should tweak other code that has "ut".

Even better: create two new types that inherit from `int`: `UtMs` and `UtS` and then use those everywhere for type-checking.

My whole goal is to avoid having ms where s were intended, and vice-versa. It's a huge source of bugs.

My recommendation for you (Berkay) moving forward pragmatically:

- In current commit: name the variables `cur_ut_s` etc. Get the PR approved & merged
- Create a new issue for: "create two new types that inherit from `int`: `UtMs` and `UtS` and then use those everywhere for type-checking".
- Do the big merge: merge `yaml-cli2` branch into main , and surrounding activities
- .. (other TODOs) ..

**Roberto:**

> Another idea: whenever we use ut ..
> Even better: create new types ..
> My whole goal is to avoid ... bugs

Love these comments. Specifying everywhere whether its `ut_s`, `ut_ms`, or using custom types, would be very helpful.

PR #550, opened 2024-01-17
Fix #512: Add helper functions to manage accounts

### LINKED ISSUE #512: [YAML] Add helper functions to manage accounts and predictoors

idiom-bytes:
Fixes #512 

This PR is a result of closing [PR #523](https://github.com/oceanprotocol/pdr-backend/pull/523), fetching latest from main, and cherry-picking to resolve the PR.


pdr_backend/cli/cli_arguments.py
```diff
@@ -112,6 +137,54 @@ def add_argument_LOOKBACK(self):
         )
 
 
+@enforce_types
+def check_positive(value):
+    try:
+        ivalue = int(value)
+        if ivalue <= 0:
+            raise Exception("Zero or below.")
```
calina-c, line 145: I think it makes sense to separate the conditions for a better readability.

I.e. try the int conversion, throw a type error if it fails specifying that it's not a valid int.
then do the comparison and throw a type error if it is negative

=> two separate, independent and easy to read try/catch blocks

calina-c (changes requested):
Looks great! I suggested just one more tweak

calina-c (approved):

PR #559, opened 2024-01-19
[FROZEN-DO NOT MERGE] Fix #481: Add Payouts to the Data Factory 

### LINKED ISSUE #481: [Lake] integrate pdr_payouts into GQL Data Factory

kdetry:
Fixes #481 

Changes proposed in this PR:

- Added Payouts schema
- Created fetching process for payouts
- Added fn to fetch payouts from subgraph
- Added mocks and tests

### DIFF
--- a/pdr_backend/lake/gql_data_factory.py
+++ b/pdr_backend/lake/gql_data_factory.py
@@ -46,11 +46,11 @@ class GQLDataFactory:
             owner_address=self.ppss.web3_pp.owner_addrs,
             network=network,
         )
-        contract_list = [f.lower() for f in contract_list]
+        # contract_list = [f.lower() for f in contract_list]
 
         # For debugging
-        # t_contract_list = [f.lower() for f in contract_list]
-        # contract_list = [t_contract_list[0], t_contract_list[1]]
+        t_contract_list = [f.lower() for f in contract_list]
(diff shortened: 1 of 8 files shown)


idiom-bytes:
I believe this is in a good state now.
Please do not update any further.

Let's focus on the ETL/merge step.

idiom-bytes (approved):
I have reviewed and iterated on a lot of this PR and it's nearly good to go.

There are various things that it needs from the Truevals PR to properly complete.
Freeze this PR, and cherry-pick into the final PR.

idiom-bytes:
Great work @kdetry! Closing PR.

Issue #482, opened 2024-01-03
[Lake] ETL has created all bronze tables


idiom-bytes:
# Motivation
The role of data_factory.py is to perform a single-step procedure to get the data written to disk. However, this data may need to be manipulated further, such as joined with other data, and then aggregated.

To solve for this, etl.py will be responsible for taking all steps required to generate/compute data towards analytics/intelligence.
- bronze_step()
- silver_step()
- gold_step()

To complete this task, we'll be using all the bronze_tables generated by gql_data_factory:
- pdr_predictions
- pdr_truevals
- pdr_payouts

For now GQLDataFactory will implement a post_fetch step, thtwill join data from truevals and payouts to `pdr_predictions`. At the end of this post_fetch step, other tables and summaries can be kicked off.

### DoD:
- [x] data factories should be responsible for generating all bronze records
- [x] GQL data factory can then execute the incremental post_pdr_predictions table #495
- [ ] Continue using GQLDataFactory or build ETL.py that is responsible for building core analytics tables (bronze=>silver=>gold)

idiom-bytes:
Alternatively, expand gql_data_factory, such that it can handle:
1. Post-fetch processing of bronze tables
2. Processing of silver tables (enrich, aggregate for time-series analysis)
3. Processing of gold tables (metrics, summaries, lookup tables)

```
    def __init__(self, ppss: PPSS):
        self.ppss = ppss

        # filter by feed contract address
        network = get_sapphire_postfix(ppss.web3_pp.network)
        contract_list = get_all_contract_ids_by_owner(
            owner_address=self.ppss.web3_pp.owner_addrs,
            network=network,
        )
        
        # TODO - make contract_list a constant, filtering happens after full sync
        # 1. configure all tables that will be recorded onto lake
        self.bronze_config = {
            "bronze_pdr_predictions": {
                "fetch_fn": get_pdr_predictions_df,
                "schema": predictions_schema,
                "config": {
                    "contract_list": contract_list,
                },
            },
            "bronze_pdr_truevals": {
                "fetch_fn": get_pdr_truevals_df,
                "schema": truevals_schema,
                "config": {
                    "contract_list": contract_list,
                },
            },
            "bronze_pdr_payouts": {
                "fetch_fn": get_pdr_payouts_df,
                "schema": payouts_schema,
                "config": {
                    "contract_list": contract_list,
                },
            },
            "bronze_pdr_subscriptions": {
[...]

PR #560, opened 2024-01-19
[Lake] Issue 495: TDD incremental pdr_predictions table

kdetry:
It was @idiom-bytes PR, I just created a new branch and I did a cherry-pick from his work.

I left the original message below:

### Motivation
As described in #482 create the basic set of functionality required to create/update the <post_pdr_predictions> table incrementally.

Consider the order of events, and how rows will be created/updated. 
1. Predictions
2. Truevals
3. Payouts

### Incremental update of post_pdr_predictions table
This implementation was provided as a way to not change any of the raw data. Instead, it creates a new table post_pdr_predictions.
1. from pdr_predictions_table - get new predictions from latest update (may already have up-to-now truevals, payouts) and append to post_pdr_predictions
1. from pdr_truevals_table - get new truevals for slots older than latest update (skip new truevals for up-to-now predictions).
1. from pdr_payouts_table - get new payouts for slots older than latest update (skip new payouts for up-to-now predictions).
1. left join existing post_pdr_predictions rows using (2)(3) as filters to reduce join size
1. function returns final, incremental, post_pdr_predictions df

### Incremental update of existing pdr_predictions table
This implementation modifies the subgraph raw data that is saved locally, such that it's up-to-date with what subgraph would provide. This updates the pdr_predictions table, which can then be consumed directly.
1. from pdr_truevals_table - get new truevals for slots older than latest update (skip new truevals for up-to-now predictions).
1. from pdr_payouts_table - get new payouts for slots older than latest update (skip new payouts for up-to-now predictions).
1. left join existing pdr_predictions rows using (2)(3) as filters to reduce join size
1. function returns final, updated, pdr_predictions df

Polars/Parquet currently doesn't support partitions, but it does support pre-filtering w/ timestamp. Focus on the least amount of work required to complete the dataframe/steps.

### DoD:
- [ ] Tests verify that <post_pdr_predictions> table is being built and updated correctly
- [ ] Tests have access to Predictions, Truevals, Payouts event/data
- [ ] Tests cover different data flows/scenarios for what's available during each run
- [ ] Tests covers logic that minimizes total number of records processed.
1. Test that new trueval events update existing post_prediction records
1. Test that new payout events update existing post_prediction records.
1. Test that only new prediction events are appended, and unnecessary updates are being minimized.
1. Test that unnecessary trueval and payout updates are being minimized.
1. Test that post_prediction records, and downstream records are updated incrementally.

### DIFF
--- a/pdr_backend/lake/test/test_gql_data_factory.py
+++ b/pdr_backend/lake/test/test_gql_data_factory.py
@@ -8,6 +8,7 @@ from pdr_backend.lake.table_pdr_predictions import predictions_schema
 from pdr_backend.lake.test.resources import _gql_data_factory, _filter_gql_config
 from pdr_backend.subgraph.subgraph_predictions import FilterMode
 from pdr_backend.util.timeutil import timestr_to_ut
+from pdr_backend.subgraph.prediction import mock_prediction
 
 # ====================================================================
 # test parquet updating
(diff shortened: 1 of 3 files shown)


pdr_backend/lake/test/test_gql_data_factory.py
```diff
@@ -352,3 +351,312 @@ def test_load_missing_parquet(
...
+    payouts = mock_payouts()
+
+    gql_dfs = {
+        "pdr_predictions": pl.DataFrame(predictions),
+        "pdr_truevals": pl.DataFrame(truevals),
+        "pdr_payouts": pl.DataFrame(payouts),
+    }
+
+    print("\n\ngql_dfs", gql_dfs)
+
```
idiom-bytes, line 653: before the post_processing, show/assert the original values

pdr_backend/lake/test/test_gql_data_factory.py
```diff
@@ -352,3 +351,312 @@ def test_load_missing_parquet(
...
+    # process post truevals
+    gql_data_factory._post_fetch_process_truevals(gql_dfs)
+    predictions_df = pl.read_parquet("pdr_predictions.parquet")
+    print(">>> truevals saved_parquet predictions_df", predictions_df)
+
+    # process post payouts
+    gql_data_factory._post_fetch_process_payouts(gql_dfs)
+    predictions_df = pl.read_parquet("pdr_predictions.parquet")
+    print(">>> payouts saved_parquet predictions_df", predictions_df)
```
idiom-bytes, line 662: verify that after the post_processing, show/assert the new, updated records, as you'd expect them

idiom-bytes:
I cleaned/finished both pdr_truevals and pdr_payouts and put them on a PR for main.
https://github.com/oceanprotocol/pdr-backend/pull/571

I then merged all of that into ETL PR #560 and pushed a major update.
https://github.com/oceanprotocol/pdr-backend/pull/560

Besides having the basics in place, we now also have a nice, clean, test setup ready to extend and build a bunch of coverage.
https://github.com/oceanprotocol/pdr-backend/blob/6833c03f9556811c0fbeac9b2d6e9714eb70511e/pdr_backend/lake/test/test_etl.py

**We now have 2 classes building out the lake:**
**GQLDataFactory** - Responsible for fetching from subgraph + dumping to disk
**ETL** - Responsible for running GQLDataFactory, then kicking off the rest of the pipeline

**We now also have 1 more table:**
**bronze_pdr_predictions**
- Built from [pdr_predictions, pdr_truevals, pdr_payouts]
- Has extra fields to enrich the table
- **WIP:** Build incrementally
- **WIP:** Incremental tests

**We now have clean conftest/fixtures/mockups to work with dataframes.**  
Please be mindful of this if you're building tests against the /lake or tests against subgraph/ and fetching. They are very different. This takes us closer to the epic diagram. Except that now, after some adjustments:  
**raw tables created by Data Factory will be called** = [pdr_predictions, pdr_truevals, pdr_payouts]
**bronze_tables created by ETL will be called** = [bronze_pdr_predictions, bronze_pdr_slots]

idiom-bytes:
query{
  predictPayouts{
    payout
    prediction{
      slot{
        predictContract(where: {id_in:["0x18f54cc21b7a2fdd011bea06bba7801b280e3151"]}){
          token {
            id
          }
          
        }
        
      }
    }
  }
}

pdr_backend/lake/bronze_table_pdr_predictions.py
```diff
@@ -0,0 +1,183 @@
...
+
+def _process_truevals(
+    dfs: Dict[str, pl.DataFrame], ppss: PPSS
+) -> Dict[str, pl.DataFrame]:
+    """
+    Perform post-fetch processing on the data
+    """
+    # get truevals within the update
+    truevals_df, _ = pick_df_and_ids_on_period(
+        target=dfs["pdr_truevals"],
```
KatunaNorbert, line 83: I suggest we make all the dfs table names as constants, and maybe we can define them in their file. For example "pdr_truevals" inside `table_pdr_truevals.py`

idiom-bytes: `Table. table_name`

idiom-bytes:
Hi @kdetry or @KatunaNorbert, could you perhaps do a review of the code base.

pdr_backend/lake/table_bronze_pdr_predictions.py
```diff
@@ -0,0 +1,187 @@
...
+        2. Transform predictions to bronze
+        3. Concat to existing table
+    """
+
+    # only add new predictions
+    predictions_df = dfs["pdr_predictions"].filter(
+        (pl.col("timestamp") >= ppss.lake_ss.st_timestamp)
+        & (pl.col("timestamp") <= ppss.lake_ss.fin_timestamp)
+        & (pl.col("ID").is_in(collision_ids).not_())
+    )
```
KatunaNorbert, line 50: Implementation looks good, an improvement we could have in the future is to only update the existing columns and add new rows whenever new data is fetched from subgraph instead of recreating the entire bronz table each time new data is fetched

idiom-bytes: Great, I agree! 
I added some logic (1) to reduce how many records need to be modified but this can be improved further.

I have also (2) shared latest on how to avoid recreating the bronze table every time.

KatunaNorbert (approved):

idiom-bytes:
@KatunaNorbert for visibility on how updates to fetching broke other functionality.
image

Basically, it broke because of the issues inside [#614](https://github.com/oceanprotocol/pdr-backend/issues/614)
So, let's address 614 in it's own PR. Then, merge it here.

I'm not sure how your tests passed... this looks like `test_table.py::test_get_pdr_df_multiple_fetches` should be breaking, as it is in this PR.

idiom-bytes:
Some notes from preparing the PR:
1. Please start using logger rather than print from now on.
2. I believe that logger might be muted during CICD, so tests that assert against `stdout` may not pass. More importantly, we should not be writing tests/validating based on print statements (I might have written these... :shrug:). There are various ways to validate/test edge cases, let's continue to improve how we complete these.

PR #566, opened 2024-01-22
#565 - Override PPSS parameters via nested arguments

### LINKED ISSUE #565: When running an agent, be able to override ppss.yaml parameters via arguments
When running an agent, be able to override ppss.yaml parameters via arguments.
Towards #369 

This feature is required to run multiple agents for different feeds without having to create many yaml files.

trizin:
Fixes #565

Changes proposed in this PR:

- Override PPSS parameters via nested arguments.
- Introduced NestedArgParser class, which parses nested arguments.
- CustomParser's base class is now NestedArgParser class.
- Added dict utils and recursive_update function.
- Updated ppss class to call recursive_update on nested argument dict to update original yaml attributes.
- Unit tests for new functions and NestedArgParser class.

**Any key in the yaml file can be overridden by passing the --key1.key2...keyn.value=x in the pdr cli arguments, this works for all pdr commands.**

Usage Example:
```
pdr predictoor 1 ppss.yaml sapphire-testnet --predictoor_ss.predict_feed="binance BTC/USDT c 5m" --predictoor_ss.aimodel_ss.input_feeds='["binance BTC/USDT c
 5m"]' --lake_ss.feeds='["binance BTC/USDT c 5m"]'
```

### DIFF
--- a/pdr_backend/cli/cli_module.py
+++ b/pdr_backend/cli/cli_module.py
@@ -52,7 +52,11 @@ def _do_main():
 
 @enforce_types
 def do_xpmt(args, nested_args=None):
-    ppss = PPSS(yaml_filename=args.PPSS_FILE, network="development", nested_override_args=nested_args)
+    ppss = PPSS(
+        yaml_filename=args.PPSS_FILE,
+        network="development",
+        nested_override_args=nested_args,
+    )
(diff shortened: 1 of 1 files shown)


trentmc (approved):
Great solution

Issue #387, opened 2023-11-21
[EPIC] {Mock, real} trades on {historical, live} data, by updating sim_engine & trader


trentmc:
### Target outcome

xpmt_engine (nee sim_engine)(xpmt = "experiment") should support these flows:
- Mock trade on historical price feeds. Therefore always local predictions
- Mock trade on live price feeds, local predictions
- Real trade on live price feeds, local predictions

Pdr bots + trader bots + a network should support these flows: (ie they don't need to be in sim_engine / xpmt_engine)
- Mock trade on live price feeds, network/contract predictions
- Real trade on live price feeds, network/contract predictions

### TODOs

Pre-requisite:
- [x] Trader kinda smells now. Refactor it to smell less.

Features needed to support the xpmt_engine flows:
- [x] Rename sim_engine -> xpmt_engine, sim_ss -> xpmt_ss, etc.
- [ ] **In trader bot, support for mock trading.** So that trader bot now supports two flows: {mock trade on live price feeds, real trade on live price feeds}. 
  - in YAML, `trader_ss` gets a new attribute `tradetype`. Value can be `livemock` or `livereal`,for {mock trade on live price feeds, real trade on live price feeds} respectively. Default is `livereal`. The labels `tradetype`, `livemock`, and `livereal` are chosen for consistency with xpmt_engine; its details are below.
  - `pdr trader` CLI has an additional optional arg: `--tradetype=livemock|livereal`. If it's specified in CLI, it overrides the YAML setting.
  - how to implement: where we currently call to a ccxt exchange's "buy" or "sell": override with a fake "buy" or "sell". It still uses ccxt exchange live data as-is. Note: make this easy to use, as xpmt_engine will use it to.
  - logging remains like before, except it notes whether "livemock" or "livereal".
  - need to update READMEs/trader.md too. Include that `mock` is like a "dry run" or "backtesting".
- [ ] **In xpmt_engine, add support for livemock & livereal trading.** So that xpmt_engine supports three flows: {mock trade on historical price feeds, mock trade on live price feeds, real trade on live price feeds}. 
  - in YAML, `xpmt_ss` gets a new attribute `tradetype`. Value can be `histmock`, `livemock` or `livereal`, for {mock trade on historical price feeds, mock trade on live price feeds, real trade on live price feeds} respectively. Default is `histmock`, because it runs super-fast. 
[...]

PR #568, opened 2024-01-23
Trader livemocking.

calina-c:
Work on #387.


pdr_backend/ppss/base_ss.py
```diff
@@ -174,3 +183,36 @@ def get_feed_from_candidates(
                 return feed
 
         return None
+
```
trentmc, line 186: MockExchange stuff should be in its own module, eg mock_exchange.py, not base_ss.py.

trentmc (approved):
One change please, then good to go.

Issue #369, opened 2023-11-15
[Predictoor/Infra/Bots] Easy to deploy and manage 20x agents. From pred to prod.


idiom-bytes:
### Problem
Right now, deploying and managing agents/infra is very manual and error-prone.
We can build tools around this, but it's a solved problem.
Further, we need to reduce the latency between sim->pred->off localhost

### Tech Spike:
Focused around shortest path towards getting all 20 agents running e2e.
Kiss implementation uses pm2 (1)5m and (2)1h feeds, on a single vm
(1) obtain gcp credentials ($500 free/90d)
(2) gitclone
(3) configure tfvars
(4) install terraform
(5) run `terraform init && terraform apply`

infra/agents are now easily managed via the cli.
Use `terraform destroy` to bring it all down.

### Towards a solution
- Build a terraform script that enables infra up/down on-command.
- tfvars file lets the user configure their pks in one place and they are passed through to the vm bootstrap process, down to pm2

### Next iteration
- consider moving terraform scripts to `pdr-infra`
- containerize pdr-backend and deploy on gcp containers
- `pdr-infra` can CI/CD on new releases on `/main/` branch
- make address.json w/ oasis contracts available somewhere official

### DoD
- [ ] user can easily deploy and manage 20x predictoor agents e2e
- [ ] improved dx from sim -> to pred -> to off localhost
- [ ] merge yaml update

idiom-bytes:
### Agent updates:
- Many updates focused around making N agents work on a single machine/instance filling nan, and addressing issues with getting N predictoors running locally e2e (i.e. configurable csvs).
- Implementation would be cleaner by using the similar setup as azure vps where local `pdr-backend` is put in a container and deployed on the vm/managed container service.

idiom-bytes:
P.S. The branch for the work done originally can be found here.
https://github.com/oceanprotocol/pdr-backend/tree/idiom-bytes-terraform-tech-spikes

Will delete when this is up and running

PR #570, opened 2024-01-24
#369 -  Make it easy to deploy manage multiple predictoor bots

trizin:
Towards #369

Changes proposed in this PR:

- Template generation.
- K8S Cloud Deployment
- Docker build and remote container registry management.


idiom-bytes:
Some really great code in here @trizin

trentmc (approved):
Approving. Note that I did not personally test yet, but it will be >= days before I do. Let's get this thing in. If problems emerge when I do personally test, we'll open new issues.

PR #580, opened 2024-01-29
Fix #582: Make ppss.yaml copy directions more explicit

### LINKED ISSUE #582: [UX] Make ppss.yaml copy directions more explicit
### Background / motivation

In the predictoor.md and trader.md READMEs, there is:

Screenshot 2024-01-30 at 08 25 56

The "copy" directions are accurate, though slightly confusing, because other console commands are explicit code-blocks.

### TODO

- [x] Update predictoor.md with an explicit code block
- [x] Update trader.md with an identical code block. (So that their "simulator" sections remain identical).

graceful-coder:
Fix #582.

Changes proposed in this PR:
- Added a small code block to the Sim flow readme for `cp ppss.yaml to my_ppss.yaml`

### DIFF
--- a/READMEs/predictoor.md
+++ b/READMEs/predictoor.md
@@ -47,6 +47,9 @@ codesign --force --deep --sign - venv/sapphirepy_bin/sapphirewrapper-arm64.dylib
 Simulation allows us to quickly build intuition, and assess the performance of the data / predicting / trading strategy (backtest).
 
 Copy [`ppss.yaml`](../ppss.yaml) into your own file `my_ppss.yaml` and change parameters as you see fit.
+```console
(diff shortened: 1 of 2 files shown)


trentmc (changes requested):
FYI, every PR should be accompanied by a github issue. So I created one: #582. Then I updated this PR to explicitly point to that issue.

As for the changes itself:
- this is good
- but it's only for predictoor.md so far
- you need to make an identical change in trader.md

(The newly-created github issue lists this as two specific TODOS.)

idiom-bytes (approved):
This is a straightforward update, this looks good. I'm approving.

trentmc (approved):

Issue #574, opened 2024-01-25
[Lake][Polars] Update tables, gql, and other polar logic to use lazy frames and perform single evaluations.


idiom-bytes:
### Motivation
We're using pl.DataFrames which means the data may be loaded into mem and computed multiple times rather than in one larger operation that polars engine can optimize.

Rather than continuing to use DataFrames we probably want to use LazyFrames more often. These may help w/ speed/memory as we can maintain a pointer to the parquet-metadata, and improve compute time by building up larger, singular dataframe evaluations that are solved once.

A simple example of how different polar commands may lead to multiple computations rather than a single one.
![image](https://github.com/oceanprotocol/pdr-backend/assets/69865342/d52376ae-a43c-438e-bba2-295c98f52c61)

### DoD:
- [ ] Review ETL and tables to use LazyFrames more judiciously and group them into single evaluations.

idiom-bytes:
Additional material for how to get structural/functional improvements for lazyframe execution.
https://towardsdatascience.com/data-pipelines-with-polars-step-by-step-guide-f5474accacc4

idiom-bytes:
We are going to shelve this PR.
- The code base has diverged quite a bit from lazy_evaluation with parquet
- Some of these changes might make a comeback after DuckDB is integrated

We're also going to close this ticket for now as we update the ETL to use DuckDB and SQL.

PR #584, opened 2024-01-30
Fix#574: DataFrame to LazyFrame - part 1

kdetry:
This pull request is currently a draft and addresses issue #574.

"he description will be completed soon.


idiom-bytes:
We are going to shelve this PR.
- The code base has diverged quite a bit from lazy_evaluation with parquet
- Some of these changes might make a comeback after DuckDB is integrated

PR #593, opened 2024-01-31
Created Table object and move logic from gql factory

### LINKED ISSUE #581: [Lake] Build table object and interface to access/update lake data
### Motivation
Rather than accessing/managing tables through files, let's abstract the tables into an object that let's us: load/save/update parquet, manage the schema, etc..

### Outline:
- [x] You shouldn't have to go to GQL DataFactory or ETL to access tables
- [ ] GQL + ETL should be authoritative writing mechanisms, that are updating the lake
- [x] Everything else, should be read-only from tables
- [x] Tables should be wrapped in an Object that helps to built, save, load, tables, etc..
- [x] Tables can help maintain a pointer to dataframe, reduce number of reads, handle caching, etc..

### DoD:
- [ ] Tables offer a clean interface to read/update data
- [ ] Code is updated to use tables through their interfaces

KatunaNorbert:
Fixes #581.


pdr_backend/lake/gql_data_factory.py
```diff
@@ -50,37 +55,32 @@ def __init__(self, ppss: PPSS):
...
-                    "contract_list": contract_list,
-                },
-            },
-            "pdr_truevals": {
-                "fetch_fn": get_pdr_truevals_df,
-                "schema": truevals_schema,
-                "config": {
-                    "contract_list": contract_list,
-                },
+            "tables": {
```
idiom-bytes, line 58: Please create a table registry outside of GQLDataFactory so we can define all tables in there.
Any other class should be able to access the registry and retrieve a table.

pdr_backend/lake/gql_data_factory.py
```diff
@@ -50,37 +55,32 @@ def __init__(self, ppss: PPSS):
...
+                "pdr_predictions": Table(
+                    predictions_table_name,
+                    predictions_schema,
+                    get_pdr_predictions_df,
+                    ppss,
+                ),
+                "pdr_subscriptions": Table(
+                    subscriptions_table_name,
+                    subscriptions_schema,
+                    get_pdr_subscriptions_df,
```
idiom-bytes, line 68: Please remove update() function from Table().

Make that a responsibility of GQLDataFactory, OHLCVDataFactory, or ETL.
Rather, have Table be more of an abstract class.

Think of a Table in a DB... It's dumb and abstract. It's only responsibility is to provide an interface for table functions. Schema, CRUD, querying-type functionality.

pdr_backend/lake/gql_data_factory.py
KatunaNorbert: I don't know where to move them within this PR. What I'm thinking is moving them to the ETL object which is responsible to create and update all the tables, but that is created inside another PR.

KatunaNorbert: Let's handle this inside the ETL

kdetry (approved):
LGTM

PR #594, opened 2024-02-01
issue572: Implement PartitionedDataStore for Efficient Data Management

### LINKED ISSUE #572: [Lake][I/O] Techspike - Middleware for parquet appending/mutating/other
### Motivation
Parquet files work w/ partitioning, but you need to have a way about it. As we work towards building an incremental pipeline, we need an approach to append records rather than what we have today. Today, we simply append data and yield a new parquet file. As data grows, this i/o will become more expensive.

### Towards a solution
Rather than creating new records, parquet/polars supports partitions. This works very similarly as a regular .parquet file, except you have `file_partitions.parquet` folder, w/ many files, that you can read natively by polars by pointing at the folder name and treating it like a regular file. Just like any other db, your partioning can impact performance.

To build the partitions effectively, perhaps we should look at some thin middleware like fastparquet or duckdb to handle the partitioning/bucketing.

### Potential solutions
As we look to support enough tables + size on-disk, we'll need a solution sooner rather than later to avoid long disk I/O times.

First, **Apache Arrow** which is the standard for how polars is built, supports a broad approach to partitioning as described below. By using [pyarrow.parquet.write_to_dataset](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_to_dataset.html)

image

https://arrow.apache.org/cookbook/py/io.html#writing-partitioned-datasets
https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_to_dataset.html
https://stackoverflow.com/questions/76121937/how-to-append-new-data-to-an-existing-parquet-file

Second, **Fastparquet** append/write w/ partitions: https://fastparquet.readthedocs.io/en/latest/quickstart.html#writing
Polars read parquet w/ partitions: https://docs.pola.rs/py-polars/html/reference/api/polars.read_parquet.html

Another potential solution would be to use **Duckdb**. However, as far as I understand it, we might need to first save the data to duckdb (db in-memory, persistent storage on analytics.db file). Such that we append to DuckDB, and then write new parquet files. DuckDB looks very promising, but leaves gaps as to whether it will allow us to append/grow incrementally.

Duckdb partitioned writes also seem to implement append in the same as pyarrow.parquet (using overwrite rules). **Please techspike to test/verify.**

https://duckdb.org/docs/data/partitioning/hive_partitioning
https://duckdb.org/docs/data/partitioning/partitioned_writes#overwriting

### Near-parquet
Polars supports other data-formats and processes (like Delta Lake) that may be able to help us across all 2 stages of the pipeline.
1. Delta Lake - Upstream/Fetching. Where we need to join + mutate records. I.e. Database like, to yield bronze-tables.
2. Arrow/Duck - Downstream/Computing. Where we just need to process records incrementally and can append-to-end/rebuild, to yield silver (aggregate) and gold (summary) tables.

### DoD
- [ ] Rather than creating new parquet files as a way to append, implement partitioning, in such a way that is saves files in efficient chunks.
- [ ] `gql_data_factory_save_parquet()` should be updated to benefit from append + partitioning
- [ ] `bronze_tables` should only save records changed, not entire partitions.
- [ ] `ohlcv_factory.save_rawohlcv_file` and any other parquet-component, should be updated to benefit from append + partitioning

kdetry:
Fixes #572 

Changes:
- Implemented DuckDB
- Introduced PartitionedDataStore for managing time-partitioned Parquet data.
- Implemented methods: append_data, finalize_data, query_data.
- Added pytest cases in test_partitioned_data_store.py.

### DIFF
--- a/pdr_backend/lake/partitioned_data_store.py
+++ b/pdr_backend/lake/partitioned_data_store.py
@@ -1,0 +1,114 @@
+import hashlib
+import os
+import polars as pl
+import duckdb
+from enforce_typing import enforce_types
+
+
+class PartitionedDataStore:
+    @enforce_types
+    def __init__(self, base_directory=str):
+        self.base_directory = base_directory
+        self.duckdb_conn = duckdb.connect(
+            database=":memory:"
+        )  # Keep a persistent connection
+
+    @enforce_types
+    def _generate_view_name(self, base_path=str) -> str:
+        """
+        Generate a unique view name for a given base path.
+        @arguments:
+            base_path - The base path to generate a view name for.
+        @returns:
+            str - A unique view name.
+        """
+
+        hash_object = hashlib.md5(base_path.encode())
+        return f"dataset_{hash_object.hexdigest()}"
+
+    @enforce_types
+    def append_data(self, df: pl.DataFrame, dataset_identifier: str):
+        """
+        Append data to an in-memory dataset.
+        @arguments:
+            df - The Polars DataFrame to append.
+            dataset_identifier - A unique identifier for the dataset.
+        """
+
+        view_name = self._generate_view_name(self.base_directory + dataset_identifier)
+        # Register the DataFrame as an Arrow table temporarily
+        # arrow_table = df.to_arrow()
+        self.duckdb_conn.register(view_name, df)
+
+    @enforce_types
+    def finalize_data(self, dataset_identifier: str):
+        """
+        Finalizes the data by writing the in-memory dataset to partitioned Parquet files.
+        @arguments:
+            dataset_identifier - A unique identifier for the dataset.
+        """
+
+        view_name = self._generate_view_name(self.base_directory + dataset_identifier)
+        query = f"SELECT * FROM {view_name}"
+        result_df = self.duckdb_conn.execute(query).df()
+
+        # Convert to Polars DataFrame for partitioned write
+        df = pl.DataFrame(result_df)
+
+        lazy_df = df.lazy()
+        # Determine partitions based on a timestamp column, assuming 'timestamp' column exists
+        lazy_df = lazy_df.with_columns(
+            [pl.from_epoch("timestamp", time_unit="s").alias("date")]
+        )
+        df = lazy_df.collect()
+
+        partitions = (
+            lazy_df.select(
+                pl.col("date").dt.year().alias("year"),
+                pl.col("date").dt.month().alias("month"),
+                pl.col("date").dt.day().alias("day"),
+            )
+            .unique()
+            .collect()
+        )
+
+        for partition in partitions.rows():
+            year, month, day = partition
+            partition_path = os.path.join(
+                self.base_directory,
+                dataset_identifier,
+                f"year={year}",
+                f"month={month}",
+            )
+            os.makedirs(partition_path, exist_ok=True)
+            partition_df = df.filter(
+                (pl.col("date").dt.year() == year)
+                & (pl.col("date").dt.month() == month)
+            )
+            file_path = os.path.join(partition_path, f"{day}.parquet")
+            partition_df.write_parquet(file_path)
+
+    @enforce_types
+    def query_data(self, dataset_identifier: str, query: str) -> pl.DataFrame:
+        """
+        Execute a SQL query across partitioned Parquet files using DuckDB.
+        @arguments:
+            dataset_identifier - A unique identifier for the dataset.
+            query - The SQL query to execute. The query must contain a {view_name} placeholder.
+
+        @returns:
+            pl.DataFrame - The result of the query.
+        """
+
+        dataset_path = os.path.join(self.base_directory, dataset_identifier)
+        view_name = self._generate_view_name(dataset_path)
+
+        # DuckDB automatically recognizes the partition scheme when reading from a directory
+        self.duckdb_conn.execute(
+            f"CREATE OR REPLACE VIEW {view_name} AS SELECT * FROM read_parquet('{dataset_path}/*/*/*.parquet', hive_partitioning=1)" # pylint: disable=line-too-long
+        )
+
+        # Execute the provided SQL query
+        result_df = self.duckdb_conn.execute(query.format(view_name=view_name)).df()
+
+        return pl.DataFrame(result_df)
--- a/pdr_backend/lake/test/test_partitioned_data_store.py
+++ b/pdr_backend/lake/test/test_partitioned_data_store.py
@@ -1,0 +1,65 @@
+import os
+import polars as pl
+from pdr_backend.lake.partitioned_data_store import (
+    PartitionedDataStore,
+)  # Adjust the import based on your project structure
+
+
+# Initialize the PartitionedDataStore instance for testing
+def _get_test_manager(tmpdir):
+    return PartitionedDataStore(str(tmpdir))
+
+
+def test_append_data(tmpdir):
+    test_manager = _get_test_manager(tmpdir)
+    example_df = pl.DataFrame(
+        {"timestamp": ["2022-01-01", "2022-02-01", "2022-03-01"], "value": [10, 20, 30]}
+    )
+    dataset_identifier = "test_append"
+    test_manager.append_data(example_df, dataset_identifier)
+
+    # Check if the view is registered
+    view_name = test_manager._generate_view_name(str(tmpdir) + dataset_identifier)
+    tables = test_manager.duckdb_conn.execute(
+        "SELECT table_name FROM information_schema.tables WHERE table_schema = 'main'"
+    ).fetchall()
+    assert view_name in [table[0] for table in tables]
+    # Since append_data doesn't immediately reflect in the filesystem, this test is limited
+
+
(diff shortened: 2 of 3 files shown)


pdr_backend/lake/partitioned_data_store.py
```diff
@@ -0,0 +1,114 @@
...
+        df = pl.DataFrame(result_df)
+
+        lazy_df = df.lazy()
+        # Determine partitions based on a timestamp column, assuming 'timestamp' column exists
+        lazy_df = lazy_df.with_columns(
+            [pl.from_epoch("timestamp", time_unit="s").alias("date")]
+        )
+        df = lazy_df.collect()
+
+        partitions = (
```
idiom-bytes, line 65: Something to keep in mind (**do not address now**) is that not every table will be partitioned the same way. Some may be by yy-mm-dd, others could be by 0xaddress or timestamp_slot.

pdr_backend/lake/partitioned_data_store.py
```diff
@@ -0,0 +1,114 @@
...
+        Append data to an in-memory dataset.
+        @arguments:
+            df - The Polars DataFrame to append.
+            dataset_identifier - A unique identifier for the dataset.
+        """
+
+        view_name = self._generate_view_name(self.base_directory + dataset_identifier)
+        # Register the DataFrame as an Arrow table temporarily
+        # arrow_table = df.to_arrow()
+        self.duckdb_conn.register(view_name, df)
```
idiom-bytes, line 41: Although this is called append, it doesn't technically append.
As far as I can tell, this re-creates a new view every time and registers the table.

But, if the table already exists, then shouldn't this insert rows into the duckdb table?

pdr_backend/lake/partitioned_data_store.py
```diff
@@ -0,0 +1,114 @@
...
+                f"year={year}",
+                f"month={month}",
+            )
+            os.makedirs(partition_path, exist_ok=True)
+            partition_df = df.filter(
+                (pl.col("date").dt.year() == year)
+                & (pl.col("date").dt.month() == month)
+            )
+            file_path = os.path.join(partition_path, f"{day}.parquet")
+            partition_df.write_parquet(file_path)
```
idiom-bytes, line 89: I think if all the data is already inside of duck_db through append(), then we can use duck db's partitioned-writes functionality to handle all of this, such that it doesn't have to be done twice: (1) read from duck_db (2) partition to disk

https://duckdb.org/docs/archive/0.9.2/data/partitioning/partitioned_writes

pdr_backend/lake/partitioned_data_store.py
```diff
@@ -0,0 +1,114 @@
...
+        @returns:
+            pl.DataFrame - The result of the query.
+        """
+
+        dataset_path = os.path.join(self.base_directory, dataset_identifier)
+        view_name = self._generate_view_name(dataset_path)
+
+        # DuckDB automatically recognizes the partition scheme when reading from a directory
+        self.duckdb_conn.execute(
+            f"CREATE OR REPLACE VIEW {view_name} AS SELECT * FROM read_parquet('{dataset_path}/*/*/*.parquet', hive_partitioning=1)"  # pylint: disable=line-too-long
```
idiom-bytes, line 108: Hi, we do not want to be loading all the data into memory. This will eventually become too big.

Please do the following:
(1) update query_data() to run queries against the self.duckdb persistent tables.
(2) create a different query function to read from polars (look for Norbert's Table update) such that we (a) generate the `parquet-hive-partition.parquet` path and then (b) return a pl.LazyFrames(path) such that the data is not loaded onto memory and a polars lazy interface is returned
(3) update current query function to read from duck db peristent tables

In the end:
1. saving to parquet will be done in 2 steps: (a) via the db by registering/appending, then (b) by calling finalize()
2. most of the ETL work will be done via sql, on the db, such that only the required records are reprocessed
3. you can then query clean, final tables, or just the latest records w/ sql, for immediate results
4. as a final step of the ETL, we dump the latest db/records to parquet

pdr_backend/lake/test/test_partitioned_data_store.py
```diff
@@ -0,0 +1,65 @@
...
+    dataset_identifier = "test_append"
+    test_manager.append_data(example_df, dataset_identifier)
+
+    # Check if the view is registered
+    view_name = test_manager._generate_view_name(str(tmpdir) + dataset_identifier)
+    tables = test_manager.duckdb_conn.execute(
+        "SELECT table_name FROM information_schema.tables WHERE table_schema = 'main'"
+    ).fetchall()
+    assert view_name in [table[0] for table in tables]
+    # Since append_data doesn't immediately reflect in the filesystem, this test is limited
```
idiom-bytes, line 27: Please show that we can continue to append data, and that the records in the db are increasing.
We should be able to call "select * from dataset_identifier" and see the `len(records)` growing.
As-is, right now, I don't believe this is working.

idiom-bytes (commented):
Great work so far @kdetry

pdr_backend/lake/partitioned_data_store.py
kdetry: Unfortunately we can not do that.

we can extract the day, month and year values as following, so it shortens our current method

`SELECT *, 
    EXTRACT(year FROM timestamp) AS year,
    EXTRACT(month FROM timestamp) AS month,
    EXTRACT(day FROM timestamp) AS day`
    
    
    However, the COPY command is not supporting this kind of manipulation on the data

pdr_backend/lake/test/test_partitioned_data_store.py
```diff
@@ -0,0 +1,65 @@
...
+    example_df = pl.DataFrame(
+        {"timestamp": [1640995200, 1643673600, 1640995200], "value": [10, 20, 30]}
+    )
+    test_manager.append_data(example_df, dataset_identifier)
+    test_manager.finalize_data(dataset_identifier)
+
+    query_result = test_manager.query_data(
+        dataset_identifier, "SELECT * FROM {view_name} WHERE value > 15"
+    )
+    assert len(query_result) == 2, "Query did not return the expected number of rows."
```
idiom-bytes, line 65: Working through reviews....

Fantastic, I was hoping it would be easy for us to do this.
Looks great!

pdr_backend/lake/partitioned_data_store.py
idiom-bytes, line 108: Looks great!

pdr_backend/lake/partitioned_data_store.py
```diff
@@ -0,0 +1,114 @@
...
+            dataset_identifier - A unique identifier for the dataset.
+        """
+
+        view_name = self._generate_view_name(self.base_directory + dataset_identifier)
+        # Register the DataFrame as an Arrow table temporarily
+        # arrow_table = df.to_arrow()
+        self.duckdb_conn.register(view_name, df)
+
+    @enforce_types
+    def finalize_data(self, dataset_identifier: str):
```
idiom-bytes, line 44: Please rename this to export_to_parquet()

pdr_backend/lake/test/test_partitioned_data_store.py
```diff
@@ -0,0 +1,65 @@
...
+    test_manager = _get_test_manager(tmpdir)
+    dataset_identifier = "test_query"
+    # 1643673600 is 2022-02-01 00:00:00 in Unix time
+    # 1640995200 is 2022-01-01 00:00:00 in Unix time
+    # 1640995200 is 2022-03-01 00:00:00 in Unix time
+    example_df = pl.DataFrame(
+        {"timestamp": [1640995200, 1643673600, 1640995200], "value": [10, 20, 30]}
+    )
+    test_manager.append_data(example_df, dataset_identifier)
+    test_manager.finalize_data(dataset_identifier)
```
idiom-bytes, line 60: please review feedback and update function names to:
- create_view
- export_to_parquet

idiom-bytes (changes requested):
@kdetry I reviewed this and provided you with some feedback.

This is a good tech spike towards Duck, we now need to adjust this towards the goal that we have.

Currently, the PartitionedDataStore does
1. append_data - register a dataframe as a view
2. finalize_data - exports from view to parquet
3. query_data - queries from a partitioned parquet hive store view

But, this isn't quite what we want.

Please review the comments, and perhaps even consider fixing them.

But, perhaps we should close this ticket, and the issue such that we can create a new one with a fresh/clean PR. You can then perhaps cherry-pick some code from here, such that the naming and functionality is closer towards what we know now and work towards our initial DuckDB implementation.

pdr_backend/lake/partitioned_data_store.py
kdetry: It is changed now

idiom-bytes:
@kdetry can you please:
- Address each conversation by providing a brief answer and resolving them
- Update the checklist as you work towards completing this

Cheers!

kdetry:
I will add a short description about the work. Tests are not completed yet

kdetry:
I am going to split it into two separate PRs

PR #603, opened 2024-02-06
[Freeze][Do Not Merge]: Create bronze slots table  - Prep for DuckDB

### LINKED ISSUE #595: [Lake] Hook up slots + bronze_slots table
### Motivation
Many dashboards may require a clean, up-to-date slots table.

This is the 2nd base table in predictoor to provide analytics.
Below are the e2e steps to get bronze_pdr_slots data implemented with the current system.

### DoD
- [x] subgraph_slot.py already implements a get_slot() to fetch latest slots.
- [x] subgraph object needs to be created
- [x] subgraph mock objects and fixtures
- [x] lake/table_pdr_slot.py needs to be created
- [x] schema for dfs needs to be created
- [x] hooks for gql_data_factory.py
- [x] hooks for bronze_pdr_slots.py
- [x] logic for bronze_pdr_slots.py to join data from [slots, predictions, truevals, and payouts]
- [x] tests and coverage

KatunaNorbert:
Fixes #595.


pdr_backend/lake/table_bronze_pdr_slots.py
```diff
@@ -0,0 +1,203 @@
...
+) -> Dict[str, pl.DataFrame]:
+    """
+    @description
+        Perform post-fetch processing on the data
+        1. Find predictions( within the update
+
+    """
+    # get predictions within the update
+    predictions_df, _ = pick_df_and_ids_on_period(
+        target=dfs["pdr_predictions"],
```
idiom-bytes, line 135: We need to join with the bronze_pdr_predictions table.

**The raw pdr_predictions table should be generally avoided to be queried directly.**

pdr_backend/lake/table_bronze_pdr_slots.py
```diff
@@ -0,0 +1,203 @@
...
+        collision_ids = gql_dfs[bronze_pdr_slots_table_name].filter(
+            (pl.col("timestamp") >= ppss.lake_ss.st_timestamp)
+            & (pl.col("timestamp") <= ppss.lake_ss.fin_timestamp)
+        )["ID"]
+
+    # do post sync processing
+    gql_dfs = _process_slots(collision_ids, gql_dfs, ppss)
+    gql_dfs = _process_predictions(gql_dfs, ppss)
+    gql_dfs = _process_truevals(gql_dfs, ppss)
+
```
idiom-bytes, line 199: [_process_payouts]
We need to process payouts, such that the slot is updated... Why? Because when the payout event is triggered, the `bronze_predictions` table (1) should already have been updated.

Also, since predictions were already submitted for this slot, our `predictions blob` should already have a bunch of users registered in there...

We should then find all of the users...
```
- prediction_id
- user_id
- timestamp
```

And join/update the blob such that their stake/payouts are also included...

We'll then get all users who were paid out... and join w/ the latest records from **bronze_predictions** that should be updated to include [predval, stake, payout, etc...]. We should then update the blob for the user to have the rest of the data we want, such that all the following data is saved per-user inside the slot.
```
- prediction_id
- user_id
- timestamp
- predval
- stake
- payout
```

idiom-bytes (changes requested):

pdr_backend/lake/table_bronze_pdr_slots.py
```diff
@@ -0,0 +1,203 @@
...
+        )
+        .drop(["pair", "source", "timeframe"])
+        .rename(
+            {
+                "pair_right": "pair",
+                "source_right": "source",
+                "timeframe_right": "timeframe",
+            }
+        )
+        .select(initial_schema_keys)
```
idiom-bytes, line 168: **[Columns Missing]**
We want to track all predictions that happened to the slot.

Please create a column named `predictions` with a blob-type parameter:
```
pseudo schema: {
...,
predictions: list[str] #or a comma delineated str
}
```

The key thing we want to append to this, are all the prediction_ids from all the users that have joined this epoch. If the user is already in here, it should not be added again.
```
- prediction_id
```

Now, when we got to enrich/explode this data, we can join w/ `bronze_prediction` tables.

idiom-bytes (changes requested):

pdr_backend/lake/table_bronze_pdr_slots.py
```diff
@@ -0,0 +1,203 @@
...
+    initial_schema_keys = slots_df.schema.keys()
+
+    # do work to join from pdr_payout onto bronze_pdr_predictions
+    slots_df = (
+        slots_df.join(predictions_df, on=["contract"], how="left")
+        .with_columns(
+            [
+                pl.col("pair").fill_null(pl.col("pair")),
+                pl.col("source").fill_null(pl.col("source")),
+                pl.col("timeframe").fill_null(pl.col("timeframe")),
```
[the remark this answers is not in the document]
KatunaNorbert: I'm a little bit confused, do we have a **pdr_contract** table? Should I create one?
If we don't store this values in **table_bronze_pdr_slots** we should probably not store inside **table_bonze_pdr_predictions** as well

pdr_backend/lake/table_bronze_pdr_slots.py
KatunaNorbert: Isn't this going to make the two bronze tables too dependent from each other? This way we have to always make sure that the bronze predictions table is always up to date before being able to get up to date from the slots data

pdr_backend/lake/table_bronze_pdr_slots.py
KatunaNorbert: I don't think this is necessary because we already have the contract and slot fields inside the bronze_slots and we can get the predictions based on that. Instead of prediction id just get by same contract and slot, then no need fot the extra logic to keep the list of prediction ids

idiom-bytes:
@KatunaNorbert I have updated and solved a lot of the conflicts... but, the code is now old relative to the latest Table/Saving/Backoff implementations... please:
- [ ] update subgraph_slot.py and slot fetching to work w/ saving + backoff
- [ ] finish integrating slots table with latest updates
- [ ] get tests to work
- [ ] slots should be clean, tested, and working with the remainder of the code base

pdr_backend/lake/table_bronze_pdr_slots.py
idiom-bytes: Ok, we're not going to fix this now and will fix this in the future when implementing the `pdr_contracts` table

pdr_backend/lake/table_bronze_pdr_slots.py
idiom-bytes: 1. Like we discussed, the `bronze_predictions` table should always be updated first.
2. `bronze_slots` should use `bronze_predictions` table in case there is any transformation/cleaning up that `bronze_predictions` has done.

Resolving this conversation.

pdr_backend/lake/table_bronze_pdr_slots.py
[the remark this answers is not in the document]
idiom-bytes: This is resolved. We'll create a separate `pdr_contract` table in the future.

pdr_backend/lake/table_bronze_pdr_slots.py
```diff
@@ -0,0 +1,161 @@
...
+
+    # create contract column from the ID column
+    slots_df = slots_df.with_columns(
+        contract=pl.col("ID").map_elements(lambda s: s.split("-")[0])
+    )
+
+    if len(slots_df) == 0:
+        return dfs
+
+    if dfs[bronze_pdr_slots_table_name].schema == bronze_pdr_slots_schema:
```
idiom-bytes, line 67: I don't know what this conditional statement is doing... as far as I can tell, it's a bug/extra...

pdr_backend/lake/test/test_bronze_tables.py
```diff
@@ -0,0 +1,84 @@
...
+    bronze_pdr_slots_schema,
+)
+from pdr_backend.lake.table_bronze_pdr_predictions import (
+    get_bronze_pdr_predictions_df,
+    bronze_pdr_predictions_schema,
+)
+
+
+@enforce_types
+def test_bronze_tables_coraltion(
```
idiom-bytes, line 16: please fix function name

pdr_backend/lake/test/test_table_bronze_pdr_slots.py
```diff
@@ -0,0 +1,61 @@
...
+    fin_timestr = "2023-11-07_0:00"
+
+    ppss, _ = _gql_data_factory(
+        tmpdir,
+        "binanceus ETH/USDT h 5m",
+        st_timestr,
+        fin_timestr,
+    )
+
+    gql_dfs = {
```
idiom-bytes, line 32: please rename gql_dfs to gql_tables

pdr_backend/subgraph/slot.py
```diff
@@ -0,0 +1,98 @@
...
+
+@enforce_types
+class Slot:
+    # pylint: disable=too-many-instance-attributes
+    def __init__(
+        self,
+        ID: str,
+        timestamp: int,
+        slot: int,
+        trueval: bool,
```
kdetry, line 14: It is decided to use the "truevalue" phrase instead of "trueval". Can you change it ?

kdetry (approved):

pdr_backend/subgraph/slot.py
KatunaNorbert: let's handle this inside this issue #650

pdr_backend/lake/table_bronze_pdr_slots.py
```diff
@@ -0,0 +1,152 @@
...
+    ]
+
+    # Add missing columns to df2 using column-wise assignment
+    for col in missing_columns:
+        slots_df = slots_df.with_columns(
+            pl.Series(name=f"{col}", values=[None] * len(slots_df))
+        )
+
+    # Reorder the DataFrame columns based on the schema
+    slots_df = slots_df.select(bronze_pdr_slots_schema)
```
idiom-bytes, line 72: I'm still confused by this code.
```
    # Identify missing columns
    missing_columns = [
        col
        for col in tables[bronze_pdr_slots_table_name].df.columns
        if col not in slots_df.columns
    ]

    # Add missing columns to df2 using column-wise assignment
    for col in missing_columns:
        slots_df = slots_df.with_columns(
            pl.Series(name=f"{col}", values=[None] * len(slots_df))
        )

    # Reorder the DataFrame columns based on the schema
    slots_df = slots_df.select(bronze_pdr_slots_schema)
```

What missing columns are you talking about?
It looks like you're joining a bunch of columns to a table and inserting "None" in a really inefficient way.

how to get an empty df with all required columns
```
slots_df = pl.DataFrame(schema=bronze_pdr_slots_schema)
```

idiom-bytes:
I think I noticed some implementation issues... the slots table isn't complete.
Please give me time to address it such that I can solve the join.

pdr_backend/lake/table_bronze_pdr_slots.py
KatunaNorbert: I don't see what is the issues here..
Why explicitly setting the columns helps? You have the bronze table schema defined above if you want to see the fields, this only adds them to slots table so that we can merge the slots values into our data frame.
We don't need to do any join in this step because the bronze_table is empty, what column should we join on?

idiom-bytes: ```Why explicitly setting the columns helps?```

Because whoever is reading the query/code can't just look at what you wrote and understand how it works.

In your original code, I can't tell what data is expected to be available when the slot is first created.. just that "there are a bunch of columns" that don't have the data.

With ETL/queries/dataframes, it's better to be explicit rather than implicit.

In your test, you're assuming that roundSumStakes and roundSumStakesUp are there when the slot is created... **but this is not how the data/events/system works.**

image

pdr_backend/subgraph/slot.py
```diff
@@ -0,0 +1,98 @@
...
+        1.5,
+        2.5,
+    ),
+    (
+        "0xe66421fd29fc2d27d0724f161f01b8cbdcd69690-1696838700",
+        1696838700000,
+        1696838700,
+        True,
+        10.5,
+        20.5,
```
idiom-bytes, line 96: **You've added test data assuming that roundSumStakes, and roundSumStakesUp come with this data when slots are first created... but they do not.**

In other words, you setup the est in a way that doesn't actually map to how the system works underneath.

pdr_backend/subgraph/slot.py
```diff
@@ -0,0 +1,98 @@
...
+        1696838400,
+        True,
+        1.5,
+        2.5,
+    ),
+    (
+        "0xe66421fd29fc2d27d0724f161f01b8cbdcd69690-1696838700",
+        1696838700000,
+        1696838700,
+        True,
```
idiom-bytes, line 94: There is also no trueval on the original slot. 

When slots are first created:
- there are no predictions
- truevalue is false
- there are no roundSumStakesUp, roundSumStakes

All of these values are joined later, as predictions are submitted by predictoors, truevalue is submitted by judge, and payouts are collected by users.

image

pdr_backend/lake/test/test_table_bronze_pdr_slots.py
```diff
@@ -0,0 +1,244 @@
...
+    gql_tables["pdr_payouts"].df = gql_tables["pdr_payouts"].df.vstack(new_row)
+    gql_tables["pdr_payouts"].df = gql_tables["pdr_payouts"].df.vstack(new_row2)
+    gql_tables["bronze_pdr_predictions"] = get_bronze_pdr_predictions_table(
+        gql_tables, ppss
+    )
+
+    gql_tables["bronze_pdr_slots"] = get_bronze_pdr_slots_table(gql_tables, ppss)
+
+    # We should have the roundSumStakes value updated withe the stakes from latest payout event
+    assert gql_tables["bronze_pdr_slots"].df["roundSumStakes"][6] == 200
```
idiom-bytes, line 243: Much better!!!

idiom-bytes:
This is much better... thanks for taking a more thorough swing at it.

I'm digging and finding some interesting details.
Will continue to iterate and improve test/coverage/validation such that things are working at expected.

I believe the following code as an example
image

Isn't correctly yielding the table as we would expect...(1) original_rows + (2) new_rows
image

Almost there... I'll see what I can do.

idiom-bytes:
A lot of the slots work has already been merged and done in the duckdb-pr.
I'm closing this off and adding a verification that this is working inside the epic.

PR #609, opened 2024-02-13
Save subgraph data to file while fetching

### LINKED ISSUE #596: [Lake][Subgraph] Add save_threshold = 10000 to fetch loop such that every 10k records, whatever has been fetched gets saved to disk
### Motivation
I'm in the middle of a giant fetch and I can't cancel it.

KatunaNorbert:
Fixes #596.


pdr_backend/lake/gql_data_factory.py
```diff
@@ -137,18 +137,49 @@ def _update(self, fin_ut: int):
...
-
-            # postcondition
-            if len(df) > 0:
-                assert df.schema == record["schema"]
-
-                # save to parquet
-                self._save_parquet(filename, df)
+            do_fetch: Callable[[str, int, int, int, int, Dict], pl.DataFrame] = record[
+                "fetch_fn"
+            ]
```
idiom-bytes, line 142: Fantastic, now all subgraph fetches handle: start, end, etc... but have no loop.

do_fetch/subgraph logic used to handle the pagination + subgraph loop, but now that's been moved to the core _update() loop of GQLDataFactory... but this is dirtying things up.

Please move the pagination/fetching loop + save-backoff logic into their own functions and parameterize them.

The _update() function for gql_data_factory() should be primarily to loop/kickoff all table updates.... 

Please do so by breaking up this function into 2 parts:
```
1. Handling do_fetch/pagination loop
2. Save-Backoff logic
```

While doing this, please adjust the param names as well so they are more objective. When you put these in their respective function, you may choose to adjust the names.
```
save_to_file_offset + fetched_since_last_save => save_backoff_limit, save_backoff_count 
records_per_page, offset => pagination_limit, pagination_offset
```

Nearly there!

Please get @kdetry or @calina-c to review and merge after.
I think the PR looks good, and can be merged in parallel to the remaining PRs.

idiom-bytes (dismissed):

idiom-bytes:
In the future, the way to approach this might be to write a `@yield` decorator that keeps writing/appending/yielding async to disk/db (function 1), while fetch/pagination logic (function 2) keeps yielding every record it gets from subgraph rather than appending to a list/array and waiting to save.

```
Decorator/Function 1 - Save-Backoff logic (@yield)
Function 2 - Handling do_fetch/pagination loop
```

pdr_backend/lake/gql_data_factory.py
KatunaNorbert: If I do the loop inside each table file then I also have to move the save function there that's why I choose to keep it inside GQL for now

KatunaNorbert: Moved logic out of GQL factory, but it's a little bit redundant now, we might want to better reuse the code once we have the Table class

idiom-bytes: Ok, let's get the Table PR merged and then look at this.

pdr_backend/subgraph/test/test_subgraph_predictions.py
```diff
@@ -117,19 +116,20 @@ def test_fetch_filtered_predictions(mock_query_subgraph):
...
-    assert len(predictions) == 2000
+    assert len(predictions) == 1000
     assert isinstance(predictions[0], Prediction)
     assert predictions[0].user == "0xd2a24cb4ff2584bad80ff5f109034a891c3d88dd"
     assert predictions[0].pair == "ADA/USDT"
     assert predictions[0].address[0] == "0x18f54cc21b7a2fdd011bea06bba7801b280e3151"
     assert predictions[0].trueval is False
     assert predictions[0].prediction is True
-    assert mock_query_subgraph.call_count == 3
+    assert mock_query_subgraph.call_count == 1
```
kdetry, line 132: It needs to be designed according to a multiple subgraph call scenario. 
There are mock data for three calls a few lines above. 
If we only test it for one subgraph query call, it won't be the proper test.

pdr_backend/subgraph/subgraph_subscriptions.py
```diff
@@ -58,79 +58,71 @@ def fetch_filtered_subscriptions(
...
+        info725 = subscription_sg_dict["predictContract"]["token"]["nft"]["nftData"]
+        info = info725_to_info(info725)
+        pair = info["pair"]
+        timeframe = info["timeframe"]
+        source = info["source"]
+        timestamp = subscription_sg_dict["timestamp"]
+        tx_id = subscription_sg_dict["txId"]
+        last_price_value = (
+            float(subscription_sg_dict["predictContract"]["token"]["lastPriceValue"])
+            * 1.201
```
kdetry, line 111: I see it is not added by you but I don't know what means this value

kdetry (changes requested):

pdr_backend/subgraph/test/test_subgraph_predictions.py
KatunaNorbert: Inside the subgraph now it's just 1 call so this in right, but in the table it should check that subgraph it's fetched multiple time, I'm just not sure how to mock that. I could just let the function query the subgraph directly and adjust the time inside a mocked ppss

pdr_backend/subgraph/subgraph_subscriptions.py
KatunaNorbert: I'm not sure either..

Issue #599, opened 2024-02-05
[Pdr bot, sim] Implement two-sided prediction into sim_engine, pdr bot 1, pdr bot 3


trizin:
### TODOs

Parent epic: [#680](https://github.com/oceanprotocol/pdr-backend/issues/680)  Two-sided prediction bots

Berkay:
- [x] First cut of this issue (classifier, bot 3): https://github.com/oceanprotocol/pdr-backend/pull/613

Trent:
- [x] Make Aimodel do classification (not regression), where classifier outputs probabilities of class 1 vs 2 (up vs down).
- [x] In sim_engine predictoor: model two-sided prediction
- [x] In sim_engine trader: use up/down confidence level to decide amt of $ to put in
- [x] In sim_engine trader: sell when predict down, then buy back
- [x] In sim_engine: add contour plots to visualize models, when 2 input vars (and hide plot when not 2 vars)
- [x] In predictoor_agent approach 1: stake up/down as 50/50
- [x] In predictoor_agent approach 3: stake up=prob_up*stake_amt, down=(1-prob_up)*stake_amt
- [x] rename predictoor approach 3 --> approach 2. Incl CLI
- [x] predictoor.md README updated:
  - [x] approach 3 -> 2 
  - [x] "sim engine does 2-sided predicting, and trading"
  - [x] "pdr bot need 2 txs because two-sided predicting; second tx must be a different wallet; therefore need PRIVATE_KEY2 (here's how to export it)"
- [x] create follow-up issue: k8s works w 2 keys. README updated. Issue: [#678](https://github.com/oceanprotocol/pdr-backend/issues/678)
- [x] create follow-up issue: blog post on two-sided prediction [#679](https://github.com/oceanprotocol/pdr-backend/issues/679)

## Background / Motivation

Currently, at each epoch, a predictoor bot picks a direction (eg "up") then stakes eg 100 OCEAN for that.

The idea here is: the predictoor bot calculates a probability of up (eg 60%) and down (100 - 60 = 40%), then stakes the 100 OCEAN pro-rata to each (60 OCEAN for up, 40 OCEAN for down).

**Expected benefits:**
- reduces risk - because you'll always be catching some of the slashing
- reduces variance - because you're staking pro-rata to probabilities on both sides, rather than one.
  - this in turn probably increases the average expected return per epoch
  - this should work especially well when probabilities are close, eg 60-40 (vs 95-5).
[...]

PR #625, opened 2024-02-16
#599 - Web3PP: Manage multiple private keys

trizin:
Towards #599

Changes proposed in this PR:

- The constructor smart contract classes accept an optional `pk_name` parameter. When provided, the contract uses the private key from `pk_name` env var.
- Function `web3_config_from_env` added to Web3PP that returns the web3 config object given the env variable name.

### Usage
`pk_name` variable is provided when a private key other than the default one needs to be used.

### DIFF
--- a/READMEs/get-rose-on-sapphire.md
+++ b/READMEs/get-rose-on-sapphire.md
@@ -1,19 +1,20 @@
 # Get ROSE on Sapphire Mainnet
 
-## 0. Steps
+Steps:
 
-- [1. Create a Wallet](#1-create-a-wallet)
-- [2. Purchase ROSE](#2-purchase-rose)
-- [3. Transfer ROSE from Emerald to Sapphire](#3-transfer-rose-from-emerald-to-sapphire)
+1. [Create a Wallet](#1-create-a-wallet)
+2. [Purchase ROSE](#2-purchase-rose)
+3. [Transfer ROSE from Oasis Wallet to Sapphire](#3-transfer-rose-from-oasis-wallet-to-sapphire)
 
 ## 1. Create a Wallet
 
 For optimal support and the ability to transfer ROSE between paratimes, we recommended to use Oasis Wallet extension for Chrome.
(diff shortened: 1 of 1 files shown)


pdr_backend/contract/base_contract.py
```diff
@@ -1,20 +1,26 @@
...
 
 from enforce_typing import enforce_types
 from sapphirepy import wrapper
 
 
 @enforce_types
 class BaseContract(ABC):
-    def __init__(self, web3_pp, address: str, contract_name: str):
+    def __init__(
+        self, web3_pp, address: str, contract_name: str, pk_name: Optional[str] = None
```
trentmc, line 11: A contract already sees web3_pp. It shouldn't need more info about where the private key comes from. That is, it shouldn't need the private key name specified. (Doing so adds a lot of extra code, as this PR shows.)

web3_pp already manages a private key.

Let's ask "what's the simplest thing that could possibly work" for the goal of [#599](https://github.com/oceanprotocol/pdr-backend/issues/599): "Enable staking on both sides".

First, more background.

base_predictoor agent currently makes this call to submit predictions
```python
self.feed_contract.submit_prediction(predval, stake, target_slot, wait_for_receipt=True)
```

The simplest-thing is a new predictoor approach 4 agent, where we JIT create a second feed on the second pk. (We don't even need Web3PP to manage <1 keys!) 

Here's how it would submit predictions;
```python
# given: stake_amt_up, stake_amt_down

# set up feeds
feed_up = self.feed_contract # uses PRIVATE_KEY, the default
feed_down = copy.deepcopy(feed_up)
pk_down = os.getenv("PRIVATE_KEY2")
feed_down.config = feed_b.web3_pp._web3_config = Web3Config(feed_a.web3_pp.rpc_url, pk2)

# submit txs
feed_up.submit_prediction(True, stake_amt_up, target_slot, wait_for_receipt=True)
feed_down.submit_prediction(False, stake_amt_down, target_slot, wait_for_receipt=True)
```

trentmc (changes requested):
I had a review. There's a much simpler way to implement all of #599. My comment describes how. (You don't actually need anything in this PR.)

Issue #545, opened 2024-01-16
[Sw Eng] Special Python types for Milliseconds and Seconds Unix Time


trizin:
### Background / motivation

We wish to avoid having ms where s were intended, and vice-versa. It's a huge source of bugs.

The opportunity: where we currently type-check for `int` for unix time variables, we can instead have special types, one for milliseconds (ms) and one for seconds (s).

Type-checking is a big win because the code is compact, self-documents, and is extremely good at finding / preventing errors.

(Further context / discussion on this idea is [this PR comment](https://github.com/oceanprotocol/pdr-backend/pull/539#issuecomment-1894991109), with ref to [this slack thread](https://oceanprotocol.slack.com/archives/C05P3TBHUAC/p1705337582348999))

### What label?

Candidate labels:
1. UnixTimeMilliseconds, UnixTimeSeconds
2. UnixTimeMs, UnixTimeS
3. UtMs, UtS

Pros & cons of each: see [comment below](https://github.com/oceanprotocol/pdr-backend/issues/545#issuecomment-1894983059)

### TODOs

- [ ] Agree on the label
- [ ] Create a type for unix time milliseconds, and a second one for unix time seconds. (They explicitly define whether the time value is in milliseconds or seconds.)
- [ ] Integrate the new types into rest of pdr-backend codebase. What helps: search for all variables labeled "ut"; for calls to "current_ut()", and any references to "time" or "datetime"

### Related

Follow-up issue that generalizes upon this: [#547](https://github.com/oceanprotocol/pdr-backend/issues/547) "Special Python Types for commonly-used int & str args"

trentmc:
Thanks for adding this issue!

"UnixTimeMilliseconds" and "UnixTimeSeconds" is a lot of characters. It will make argument lines quite long, and in turn because of black force multiple lines etc.

I recommend instead: "UnixTimeMs" and "UnixTimeS". Shorter, and just as clear.

PR #642, opened 2024-02-19
Towards #545: Use special time types in ppss.

calina-c:
Work on #545.

Pending TODOs:
- [ ] run all commands to make sure everything is correctly connected (tests may not be enough to demonstrate this)
- [x] figure out the pending mypy issue with floats. I presume it used to do some auto-converting to int previously


trentmc (approved):
Approving, subject to changing "UnixTimeSeconds" -> "UnixTimeS", and "UnixTimeMilliseconds" -> "UnixTimeMs"

PR #656, opened 2024-02-21
Fix #599 - [Predictoor bot] Enable staking on both sides

### LINKED ISSUE #599: [Pdr bot, sim] Implement two-sided prediction into sim_engine, pdr bot 1, pdr bot 3

trentmc:
WIP towards #599 

As of Feb 21, 2pm Berlin time: this is a demo of how to cleanly put "one-sided" and "two-sided" staking into base_predictoor_agent.

### DIFF
--- a/pdr_backend/classifiermodel/classifiermodel_data_factory.py
+++ b/pdr_backend/classifiermodel/classifiermodel_data_factory.py
@@ -1,0 +1,135 @@
+import sys
+from typing import Tuple
+
+import numpy as np
+import pandas as pd
+import polars as pl
+from enforce_typing import enforce_types
+
+from pdr_backend.ppss.predictoor_ss import PredictoorSS
+from pdr_backend.util.mathutil import fill_nans, has_nan
+
+
+@enforce_types
+class ClassifierModelDataFactory:
+    """
+    Roles:
+    - From mergedohlcv_df, create (X, y, x_df) for model building
+
+    Where
+      rawohlcv files -> rawohlcv_dfs -> mergedohlcv_df, via ohlcv_data_factory
+
+      X -- 2d array of [sample_i, var_i] : value -- inputs for model
+      y -- 1d array of [sample_i] -- target outputs for model
+
+      x_df -- *pandas* DataFrame with cols like:
+        "binanceus:ETH-USDT:open:t-3",
+        "binanceus:ETH-USDT:open:t-2",
+        "binanceus:ETH-USDT:open:t-1",
+        "binanceus:ETH-USDT:high:t-3",
+        "binanceus:ETH-USDT:high:t-2",
+        "binanceus:ETH-USDT:high:t-1",
+        ...
+        (no "timestamp" or "datetime" column)
+        (and index = 0, 1, .. -- nothing special)
+
+    Finally:
+       - "timestamp" values are ut: int is unix time, UTC, in ms (not s)
+    """
+
+    def __init__(self, ss: PredictoorSS):
+        self.ss = ss
+
+    def create_xy(
+        self,
+        mergedohlcv_df: pl.DataFrame,
+        testshift: int,
+        do_fill_nans: bool = True,
+    ) -> Tuple[np.ndarray, np.ndarray, pd.DataFrame]:
+        """
+        @arguments
+          mergedohlcv_df -- *polars* DataFrame. See class docstring
+          testshift -- to simulate across historical test data
+          do_fill_nans -- if any values are nan, fill them? (Via interpolation)
+            If you turn this off and mergedohlcv_df has nans, then X/y/etc gets nans
+
+        @return --
+          X -- 2d array of [sample_i, var_i] : value -- inputs for model
+          y -- 1d array of [sample_i] -- target outputs for model
+          x_df -- *pandas* DataFrame. See class docstring.
+        """
+        # preconditions
+        assert isinstance(mergedohlcv_df, pl.DataFrame), pl.__class__
+        assert "timestamp" in mergedohlcv_df.columns
+        assert "datetime" not in mergedohlcv_df.columns
+
+        # every column should be ordered with oldest first, youngest last.
+        # let's verify! The timestamps should be in ascending order
+        uts = mergedohlcv_df["timestamp"].to_list()
+        assert uts == sorted(uts, reverse=False)
+
+        # condition inputs
+        if do_fill_nans and has_nan(mergedohlcv_df):
+            mergedohlcv_df = fill_nans(mergedohlcv_df)
+        ss = self.ss.classifiermodel_ss
+
+        # main work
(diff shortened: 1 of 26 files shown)


pdr_backend/contract/base_contract.py
```diff
@@ -14,14 +14,17 @@ def __init__(self, web3_pp, address: str, contract_name: str):
...
         self.web3_pp = web3_pp
-        self.config = web3_pp.web3_config  # for convenience
         self.contract_address = self.config.w3.to_checksum_address(address)
         self.contract_instance = self.config.w3.eth.contract(
             address=self.config.w3.to_checksum_address(address),
             abi=web3_pp.get_contract_abi(contract_name),
         )
         self.contract_name = contract_name
 
+    @property
```
trentmc, line 24: @trizin we need to do this as a property, otherwise it's dangerous because we have two separate references to web3_config (one via web_3pp, one via config)

pdr_backend/predictoor/base_predictoor_agent.py
```diff
@@ -42,6 +42,13 @@ def __init__(self, ppss: PPSS):
         contracts = ppss.web3_pp.get_contracts([feed.address])
         self.feed_contract = sole_value(contracts)
 
+        pk2 = os.getenv("PRIVATE_KEY2")
+        self.feed_contract2 = None
+        if pk2 is not None:
+            self.feed_contract2 = copy.deepcopy(self.feed_contract)
+            self.feed_contract2.web3_pp.web3_config = \
+                Web3Config(self.feed_contract.web3_pp.rpc_url, pk2)
+
```
trentmc, line 51: The ^ is like @trizin version

pdr_backend/predictoor/base_predictoor_agent.py
```diff
@@ -82,22 +89,90 @@ def take_step(self):
         submit_epoch, target_slot = self.cur_epoch, self.target_slot
         logger.info("Predict for time slot = %s...", self.target_slot)
 
-        predval, stake = self.get_prediction(target_slot)
-        logger.info("-> Predict result: predval=%s, stake=%s", predval, stake)
-        if predval is None or stake <= 0:
-            logger.warning("Done: can't use predval/stake")
+        success = self.get_prediction_and_submit(target_slot)
+        if not success:
             return
```
trentmc, line 94: This is more manageable complexity than @trizin version, because it doesn't added 50+ more lines to `take_step()` to handle both one-sided and two-sided prediction. Rather, it pushes that functionality into a sub-method.

pdr_backend/predictoor/base_predictoor_agent.py
```diff
@@ -163,3 +238,6 @@ def get_prediction(
         timestamp: int,  # pylint: disable=unused-argument
     ) -> Tuple[bool, float]:
         pass
+
+def _tx_failed(tx):
+    return tx is None or tx["status"] != 1
```
trentmc, line 243: This avoided me having to do the test of "tx is None" and "tx["status"]" in three separate places.

BTW you always need `is None` not `== None`.

Issue #493, opened 2024-01-08
[Lake] Integrate dydx into fetch_ohlcv.py & ohlcv_data_factory.py


idiom-bytes:
### Motivation

Parent epic: [#492](https://github.com/oceanprotocol/pdr-backend/issues/492) "Explore DYDX towards making money: data_factory, trader, sim, and predictoor".

### TODOs

- [x] [Step 1] Rename fetch_ohlcv.py::safe_fetch_ohlcv() -> `safe_fetch_ohlcv_ccxt()`. Ensure all tests pass. Merge PR into main. Update: done in commit https://github.com/oceanprotocol/pdr-backend/commit/54b7038e1b68b0c1f39c49a1d5fb548a393f0b08
- [x] [Step 2] In fetch_ohlcv.py, add a new method `safe_fetch_ohlcv_dydx()`. Add unit test(s) in test_fetch_ohlcv.py. Update: done in commit https://github.com/oceanprotocol/pdr-backend/commit/188d0221581d2c3ff0e444de26ae41531a4e4239
  - [x] Build functionality && Test: Happy path of fetching data without problems
  - [x] Test: Gracefully handle nan values (vs floats). Update functionality as needed.
  - [x] Test: Handle when coin or pair doesn't exist. ""
  - [x] Test: Handle when bad start date or end date. ""
  - [x] Test: Is there a limit on # points at once? (eg 1000). ""
  - [x] Merge PR into main.
- [ ] [Step 3] In fetch_ohlcv.py, add a new method `safe_fetch_ohlcv()` which takes in a specified exchange name, and passes through to either `safe_fetch_ohlcv_ccxt()` or `safe_fetch_ohlcv_dydx()`, depending on the exchange name. Add unit test(s). Merge PR into main.
- [ ] [Step 4] Update ohclv_data_factory.py::OhlcvDataFactory._update_rawohlcv_files_at_feed() to use the (new) `safe_fetch_ohlcv()` interface, which means it will naturally support ccxt or dydx

graceful-coder:
Completed PR 1 of 4: #666

PR #673, opened 2024-02-24
Add safe_fetch_ohlcv_dydx() functionality

graceful-coder:
Changes proposed in this PR:

- In fetch_ohlcv.py, add a new method safe_fetch_ohlcv_dydx(). Add unit test(s) in test_fetch_ohlcv.py.
- This is the 2nd of 4 PRs towards solving issue #493


pdr_backend/lake/fetch_ohlcv.py
```diff
@@ -55,6 +60,45 @@ def safe_fetch_ohlcv_ccxt(
         return None
 
 
+@enforce_types
+def safe_fetch_ohlcv_dydx(
+    symbol: str,
+    resolution: str,
+    st_ut: UnixTimeMs,
```
trentmc, line 67: Key comment:

This method should have exactly the same interface / args as safe_fetch_ohlcv(). In particular, a "since" and "limit", and no "fin_ut".

This also implies that this routine has much less work to do. Eg it doesn't need to loop over and over to gather data. That's done by the data factory itself.

pdr_backend/lake/fetch_ohlcv.py
```diff
@@ -114,3 +158,136 @@ def _filter_within_timerange(
 ) -> list:
     uts = _ohlcv_to_uts(tohlcv_data)
     return [vec for ut, vec in zip(uts, tohlcv_data) if st_ut <= ut <= fin_ut]
+
+
+def fetch_dydx_data(
+    symbol: str, resolution: str, st_ut: UnixTimeMs, fin_ut: UnixTimeMs, limit: int
```
trentmc, line 164: Key comment: once you remove the complexity of no loop, and the conversion is done by a constant or smash separate method, the remaining work that this method does is small. So you can move all of its work into safe_fetch_ohlcv_dydx().

trentmc (changes requested):

Issue #547, opened 2024-01-17
[Sw eng] Special Python Types for commonly-used int & str args


trentmc:
### Background / motivation

[#545](https://github.com/oceanprotocol/pdr-backend/issues/545) is "Special Python types for Milliseconds and Seconds Unix Time".

Type-checking is a big win because the code is compact, self-documents, and is extremely good at finding / preventing errors.

That's the starting point.

We have other variables like unix time, which are basic types (int or str), repeated everywhere, and prone to mixing with other variables.

Which means an opportunity to further tighten our code for these other types.

### TODO

- [x] Inspect pdr-backend repo to find other opportunities. Document them (ideally here, by expanding the TODO checklist)
- [ ] For each opportunity: create a type, then integrate its usage throughout the pdr-backend codebase

calina-c:
I did a review of the entire app today, to determine possible such types. Along the way, I noticed some current issues such as the PredFilter being reverted, my guess is due to conflict resolution issues. Plus some other improvements I've already added in https://github.com/oceanprotocol/pdr-backend/pull/671.

Here is what I noted as possible improvements. Most but not all are related to special Python types:

- move timeframe.py from cli to utils, it should pe part of the args objects together with pair, exchange etc.
- payout.py -> should we objectify payout batches? I don’t think so
- base_ss -> exchange_pair_tups should probably be objectified
- lake_ss & others -> copy() methods of SS files just call copy. Previously they copied things from the corresponding PP but that is not the case anymore. Propose to remove.
- info725.py -> key and info 725 are good candidates for a class but I don’t think it’s warranted since they are very low level and stable
- in trueval and predictoor -> the bool/None value representing direction for predictoor and trueval can be made into a type but I don’t think it’s warranted, at least not right now
- some functions from mathutil.py are unused. E.g. intInStr, randunif. Should we remove them?
- Eth and Wei currency are **very good** candidates for becoming classes
- in strutil -> asCurrency should be moved to mathutil or a currency.py module should be created for math and str utils dealing with currency
- in strutil -> prettyBigNum is unused. Should I remove it?
- in signalstr.py -> signal is a good candidate for a class or at least an Enum. signalstr.py would retain most of its functionality but as methods for a Signal class/enum

trentmc:
In [https://github.com/oceanprotocol/pdr-backend/commit/c9fc5c352c91cde7654a2e29fa51214fd61af0b6] I just added tuned Eth/Weth currency_types: 
[...]

PR #676, opened 2024-02-26
Currency special types

calina-c:
Work on #547 - implement special currency types.

I will wrap up the rest of #547 in a different PR, let's get this through since it reaches many places in the code and fixes the pylint job as well.


trentmc (approved):

pdr_backend/predictoor/predictoor_agent.py
```diff
@@ -271,11 +273,11 @@ def calc_stakes1(self) -> Tuple[float, float]:
         """
         assert self.ppss.predictoor_ss.approach == 1
         tot_amt = self.ppss.predictoor_ss.stake_amount
-        stake_up, stake_down = 0.50 * tot_amt, 0.50 * tot_amt
+        stake_up, stake_down = tot_amt * 0.50 * tot_amt, tot_amt * 0.50
```
sylvester05, line 276: Why do we need to multiply tot_amt twice here?  @calina-c @trentmc

PR #689, opened 2024-02-28
Fix #688: PredictoorContract deepcopy() causes RecursionError

### LINKED ISSUE #688: [Bug, pdr bot] deepcopy of PredictoorContract causes RecursionError
### The issue

[Slack thread](https://oceanprotocol.slack.com/archives/C05P3TBHUAC/p1709106445386429) Feb 28, 2024:

**Norbert:** 

A user on discord is reporting the following error: `RecursionError: maximum recursion depth exceeded`

on the following line of code which I notice that was recently added: `self.feed_contract2 = copy.deepcopy(self.feed_contract)` from predictoor_agent.py 56th line

does anyone know what could be the issue?

Could be related to the system's recursion limit?

`ex: sys.setrecursionlimit(1500)`

Or maybe an issue around getting the feed_contract

Wondering if someone else encountered it

**Trent:**

We just merged that code in yesterday.
feed_contract likely references something circularly, which causes the recursion.

Candidate fixes:

1. find which attribute causes the circular reference, and change the design to not be circular
2. replace the deepcopy with a safer, custom [deep]copy routine for that class
  - variant 1: explicit copy all attributes
  - variant 2: temporarily remove the attribute with circular reference

The unit tests likely missed it because they have mocks of PredictoorContract.

I will dig further.

trentmc:
Fix #688

### DIFF
--- a/pdr_backend/contract/test/test_predictoor_contract.py
+++ b/pdr_backend/contract/test/test_predictoor_contract.py
@@ -1,3 +1,4 @@
+import copy
 from unittest.mock import Mock
 
 import pytest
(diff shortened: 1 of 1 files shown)


sylvester05 (commented):
Looks good. Good job!!!

PR #699, opened 2024-02-29
#681 - Update raw save-to-disk

### LINKED ISSUE #681: [Lake][ETL] Update raw save-to-disk
### Motivation
Parquet can append efficiently w/ a batch of records , but isnt' good for single-row/records. We want to make the process as efficient as possible for now, without adding any additional overhead (i.e. live processing + message queue => dump).

image

Rather than re-creating a parquet file, **simply append rows to the csv**.
image

### Outline
This was already partially implemented in the past (check git history).
**I do not believe that it was saving out different csvs.**
1 - Read last record from last csv
2 - Fetch all new records from subgraph
3 - Every value is validated
4 - *1000 lines per csv **(please review feature)**

### DoD
- [x] Update table.py **(and consumers)** to save/load to append/load from csv

### Final Comments
We reviewed updating ohlcv_factory to integrate CSVDataStore, however the CSV writing + OHLCV part is tightly integrated so it's going to require a bit more time. For now, we left ohlcv_factory as-is, and will create a separate ticket for updating it, such that we can continue separating concerns.

kdetry:
Fixes #681 

Changes proposed in this PR:

- CSV Data Store class is created
- implemented to the table.py
- tests are updated

PS: I didn't touch the OHLCV process because it looks like it has its own system and works properly

### DoD
Load from + Write to CSVs...Just append
- [ ] Please create a new function called `append`
- [ ] After fetching new data, users will append data to a table.
- [ ] Inside of append(), call a new function called `append_to_csv`. We just want to append a few rows at at time to a csv with as little latency as possible.
- [ ] Inside of append(), call a new function called `append_to_duckdb`. Comment this out. Address this in the next ticket.


pdr_backend/lake/csv_data_store.py
```diff
@@ -0,0 +1,286 @@
...
+        @args:
+            dataset_identifier: str - identifier of the dataset
+            start_time: int - start time of the data TIMESTAMP
+            end_time: int - end time of the data TIMESTAMP
+            row_count: int - number of rows in the data
+        """
+        start_time_str = self._fill_with_zero(start_time)
+        end_time_str = self._fill_with_zero(end_time)
+
+        return f"{dataset_identifier}_from_{start_time_str}_to_{end_time_str}_{row_count}.csv"
```
idiom-bytes, line 47: Thank you for reviewing and discussing these details @kdetry

Something that I noted is that {row_count} helps us understand whether the file is append-able from the name, but it doesn't help with indexing or other features. Also, because it's a csv file, it's very quick to index/readline and get len().

I believe we can simplify it to just this. Preferably, we're simply appending to the end of a file...
```f"{dataset_identifier}_from_{start_time_str}_to_{end_time_str}"```

If you don't know when it ends, and it can be appended, you could tag it like so (rather than having an end_time_str that is changing all the time) so you know what an open-ended file looks like:
```f"{dataset_identifier}_from_{start_time_str}_to_"```

File: pdr_backend/lake/table.py [hunk 2]
```diff
@@ -2,50 +2,42 @@
...
         self.load()
 
     @enforce_types
     def load(self):
         """
         Read the data from the Parquet file into a DataFrame object
         """
-        filename = self._parquet_filename()
+        print(f"Loading data for {self.table_name}")
+        self.csv_data_store = CSVDataStore(self.ppss.lake_ss.parquet_dir)
```
idiom-bytes, line 35: This is loading from parquet()
So, this function should be loading/querying from DuckDB PersistentStore()

pdr_backend/lake/table.py
```diff
@@ -2,50 +2,42 @@
...
-            )
-        else:
-            logger.info("Create initial df for %s", self.table_name)
-            df = pl.DataFrame(schema=self.df_schema)
-
-        # save data frame in memory
-        self.df = df
+        self.df = self.csv_data_store.read(
+            self.table_name, st_ut, fin_ut, schema=self.df_schema
+        )
```
idiom-bytes, line 40: This function used to do a polars query/return a dataframe from a parquet file.
This needs to be updated to do a SQL query from a DuckDB PersistentStore

Reading from CSV like this should be used for loading data into DuckDB..

pdr_backend/lake/table.py
```diff
@@ -63,22 +55,15 @@ def save(self):
...
-            # drop duplicates
-            self.df = self.df.filter(pl.struct("ID").is_unique())
-            self.df.write_parquet(filename)
-            n_new = self.df.shape[0] - cur_df.shape[0]
-            print(f"  Just appended {n_new} df rows to file {filename}")
-        else:  # write new file
-            self.df.write_parquet(filename)
-            print(
-                f"  Just saved df with {self.df.shape[0]} rows to new file {filename}"
-            )
```
idiom-bytes, line 81: This is the function that updates the parquet dataframe.
It's written this way because we can't append to parquet, so we're just saving everything

This function, now needs to update a DuckDB Table.
Because we haven't integrated DuckDB yet, we should not be modifying this function.

pdr_backend/lake/table.py
```diff
@@ -63,22 +55,15 @@ def save(self):
...
+        cur_df = self.csv_data_store.read_all(self.table_name, schema=self.df_schema)
+
+        self.df = pl.concat([cur_df, self.df])
+        self.df = self.df.filter(pl.struct("ID").is_unique())
+        self.csv_data_store.write(self.table_name, self.df, schema=self.df_schema)
+        n_new = self.df.shape[0] - cur_df.shape[0]
+        print(
+            f"  Just saved df with {n_new} df rows to the csv files of {self.table_name}"
+        )
```
idiom-bytes, line 66: **[Load from + Write to CSVs...Just append]**
We are not doing the same patterns with CSVs that we were doing with parquet.
This code will re-save everything using CSVs every-time.

Instead:
- [ ] Please create a new function called `append`
- [ ] After fetching new data, users will append data to a table.
- [ ] Inside of append(), call a new function called `append_to_csv`. We just want to append a few rows at at time to a csv with as little latency as possible.
- [ ] Inside of append(), call a new function called `append_to_duckdb`. Comment this out. Address this in the next ticket.

idiom-bytes:
This PR contains fixes & updates that are required for other PRs to move forward.

If your needs for CSVDataStore won't change @kdetry, then perhaps use this. Otherwise, I recommend you cherry-picking changes from as needed to continue with the other PRs and get the DuckDB Integration PR working.

I believe we should update CSVDataStore and have it "clean/simple".
There are updates required to clean up stuff in here that is part of "DuckDB Integration" PR.

pdr_backend/lake/table.py
idiom-bytes: I have updated the DoD of the task, please update there.

kdetry:
We will continue on the following PR

https://github.com/oceanprotocol/pdr-backend/pull/734

pdr_backend/lake/table.py
idiom-bytes: Fantastic, this has been updated into two functions
```
    def _append_to_csv(self, data: pl.DataFrame):
        """
        Append the data from the DataFrame object into the CSV file
        It only saves the new data that has been fetched

        @arguments:
            data - The Polars DataFrame to save.
        """
        self.csv_data_store.write(self.table_name, data, schema=self.df_schema)
        n_new = data.shape[0]
        print(
            f"  Just saved df with {n_new} df rows to the csv files of {self.table_name}"
        )

    def _append_to_db(self, data: pl.DataFrame):
        """
        Append the data from the DataFrame object into the database
        It only saves the new data that has been fetched

        @arguments:
            data - The Polars DataFrame to save.
        """
        self.persistent_data_store.insert_to_table(data, self.table_name)
        n_new = data.shape[0]
        print(
            f"  Just saved df with {n_new} df rows to the database of {self.table_name}"
        )
```

pdr_backend/lake/table.py
idiom-bytes: There are now two `load_into_cache` functions.
- one for csv
- one for duckdb persistent store

Resolving conversation

pdr_backend/lake/csv_data_store.py
idiom-bytes: Thank you for addressing this, and updating how files are saved.

This leads to a cleaner naming convention that is static (the name doesnt change after it's set), while letting you know which files can continue to be written to. Also, less logic.

We still have a problem where if we have +1000 events in a single timestamp, however this is a small concern at the moment.

Closing this feedback.

PR #713, opened 2024-03-01
Fix #702: Check balances before submitting a prediction

### LINKED ISSUE #702: [Pdr bot] Check balances of both wallets before submitting prediction
Check OCEAN and ROSEa balances of both wallets before submitting prediction. Do not submit if insufficient balance.

trizin:
Fixes #702

- Abort prediction submission if one of the wallets has insufficient balance


pdr_backend/predictoor/predictoor_agent.py
```diff
@@ -347,6 +360,34 @@ def get_ohlcv_data(self):
         return mergedohlcv_df
 
     @enforce_types
+    def check_balances(self) -> bool:
+        min_OCEAN_bal = self.ppss.predictoor_ss.stake_amount.to_wei()
```
trentmc, line 364: As discussed in standup, add two new properties: `OCEAN()` and `ROSE()`. From that, you'll be able to simplify `check_balances()` further. Including: you won't need local variables for balances or tokens.

trentmc (changes requested):
See new comment

PR #736, opened 2024-03-04
For #493 step 2/4 add a new method safe_fetch_ohlcv_dydx()

graceful-coder:
PR towards solving 2nd PR milestone in #493.

Changes proposed in this PR:

Improved typing: updated the safe_fetch_ohlcv_dydx() function to return a tuple format exactly like safe_fetch_ohlcv_ccxt() return format as revealed by mypy typing issues that showed the previous return format of raw json data from dydx was insufficient typing.

Comprehensive unit tests: added several unit tests to increase coverage beyond happy path test. 

- Added bad token format unit test
- Added nonexistent token unit test
- Added bad timeframe unit test
- Added bad date unit test
- Added bad limit unit test

### DIFF
--- a/pdr_backend/lake/test/test_fetch_ohlcv.py
+++ b/pdr_backend/lake/test/test_fetch_ohlcv.py
@@ -176,7 +176,7 @@ mock_bad_limit_dydx_response = {
             "5MINS",
             UnixTimeMs.from_timestr("2024-02-27"),
             1,
-            1709157000000,  # 1709139000000
+            1709139000000,
             61800,
             None,
             mock_dydx_response,


trentmc (changes requested):
The unit tests have many DRY violations. (Too much repeated code).

Please resolve the DRY violations, then ping me again for a review.

pdr_backend/lake/fetch_ohlcv.py
```diff
@@ -85,27 +83,52 @@ def safe_fetch_ohlcv_dydx(
...
-            + f"?resolution={timeframe}&fromISO={sinceIso}&limit={limit}",
-            headers=headers,
-            timeout=20,
-        )
-        response = requests.get(
-            "https://indexer.dydx.trade/v4/candles/perpetualMarkets/BTC-USD"
-            + "?resolution=5MINS&fromISO=2024-02-27T00:00:00.000Z&limit=1",
+            f"{BASE_URL_DYDX}/{symbol}?resolution={timeframe}"
+            f"&fromISO={since.to_iso_timestr()}&limit={limit}",
             headers=headers,
```
trentmc, line 90: The try/except loop should only be around the call to requests.get(), not to most of the method. Why: in general you want try/excepts to be as small and surgical as possible. Otherwise it gets really hard to debug.

pdr_backend/lake/fetch_ohlcv.py
```diff
@@ -85,27 +83,52 @@ def safe_fetch_ohlcv_dydx(
...
+
     except Exception as e:
         logger.warning("exchange: %s", e)
         return None
 
 
+def float_or_none(x: str) -> Optional[float]:
+    return float(x) if x is not None else None
+
+
```
trentmc, line 131: This method needs a decorator `@enforce_types` above it.

Then you will see that it will fail sometime, because the argument specification here says that `x` can only be a `str`, and not a None. What it _should_ do is allow for either `str` or `None`. How you do that is `x: Union[str, None]`. And in fact, because None is one of the types, there's an equivalent shorthand: `x: Optional[str]`.

Finally the logic to return the value can be switched around, so that you can say `is none` rather than `is not None`. It's cleaner.

I'll put it all together for you:
```python
@enforce_types
def float_or_none(x: Optional[str]) -> Optional[float]:
    return None if x is None else float(x)
```

pdr_backend/lake/fetch_ohlcv.py
```diff
@@ -85,27 +83,52 @@ def safe_fetch_ohlcv_dydx(
...
-            + "?resolution=5MINS&fromISO=2024-02-27T00:00:00.000Z&limit=1",
+            f"{BASE_URL_DYDX}/{symbol}?resolution={timeframe}"
+            f"&fromISO={since.to_iso_timestr()}&limit={limit}",
             headers=headers,
             timeout=20,
         )
         data = response.json()
-        return data
+
+        dydx_data = []
```
trentmc, line 95: `dydx_data` -> `raw_tohlcv_data` to be in line with method docstring (and to be correct).

pdr_backend/lake/test/test_fetch_ohlcv.py
```diff
@@ -43,6 +46,53 @@
     ]
 }
 
+mock_bad_token_dydx_response_1 = {
+    "errors": [
```
trentmc, line 50: Move these dydx test variables to be located right before `test_safe_fetch_ohlcv_dydx()`.

pdr_backend/lake/test/test_fetch_ohlcv.py
```diff
@@ -115,43 +165,109 @@ def test_safe_fetch_ohlcv_ccxt(exch):
...
+        (
+            "dydx",
+            "BTC-USD",
+            "5m",
+            UnixTimeMs.from_timestr("2024-02-27"),
+            1,
+            None,
+            None,
+            "resolution must be a valid Candle Resolution,"
+            + " one of 1MIN,5MINS,15MINS,30MINS,1HOUR,4HOURS,1DAY",
```
trentmc, line 214: Change `1MIN,5MINS,15MINS,30MINS,1HOUR,4HOURS,1DAY` -> `1MIN,5MINS,...`

trentmc (changes requested):
Almost there! Some small changes needed, then I will be happy to approve

trentmc (approved):

PR #757, opened 2024-03-06
Fix #742 - Cleanup duckdb integration

### LINKED ISSUE #742: [ETL][Lake] Support cleaning up DuckDB Integration
### Motivation
There are some updates to the DuckDB Integration that could use a touch. We need to continue to update Table and DF usage such that they aren't relying on parquet/lazy-evaluation any longer.

Example 1: Hanging functionality assuming we're always saving to parquet.

Example 2: Function names have not been adjusted to describe csv/db functionality... a generic "Table.load()" command does not make sense any longer.

### References to Parquet & Lazy Evaluation
**...I'm not sure how tests are passing at the moment...**

Remove code in ETL that used the parquet structure.
- We do not have the data loaded in parquet
- We do not have lazy evaluation and data pointers
- We are not storing data in parquet files

Example: `filename = table._parquet_filename()`
```

    def _update(self):
        """
        @description
            Iterate across all gql queries and update their parquet files:
            - Predictoors
            - Slots
            - Claims

            Improve this by:
            1. Break out raw data from any transformed/cleaned data
            2. Integrate other queries and summaries
            3. Integrate config/pp if needed
        @arguments
            fin_ut -- a timestamp, in ms, in UTC
        """

        for _, table in self.record_config["tables"].items():
            filename = table._parquet_filename()
            st_ut = table._calc_start_ut(filename)
            fin_ut = self.ppss.lake_ss.fin_timestamp
            print(f"      Aim to fetch data from start time: {st_ut.pretty_timestr()}")
            if st_ut > min(UnixTimeMs.now(), fin_ut):
                print("      Given start time, no data to gather. Exit.")

            # to satisfy mypy, get an explicit function pointer
            do_fetch: Callable[[str, int, int, int, int, Dict, str], pl.DataFrame] = (
                self._get_gql_df
            )

            # number of data at which we want to save to file
            save_backoff_limit = 5000
            # number of data fetched from the subgraph at a time
            pagination_limit = 1000

            print(f"Updating table {table.table_name}")
            do_fetch(
                self.record_config["tables"][table.table_name],
                self.record_config["fetch_functions"][table.table_name],
                self.ppss.web3_pp.network,
                st_ut,
                fin_ut,
                save_backoff_limit,
                pagination_limit,
                self.record_config["config"],
            )
```

### CSVDataSource & PersistentDataStore Fixes
- [x] `append_to_sources` is not a good name. A "source" is subgraph, an api, ethereum mainnet... use `append_to_storage`
- [x] Restructure `Table.load()` => `Table.load_from_db_to_cache` and `Table.load_from_csv_to_cache` and update code to reference these functions correctly.
- [x] Remove dependency from Table cache wherever possible (this was a lazy-polars pointer, and that will not work here)
- [x] Remove `df` cache from Table
- [x] Remove load() 

### End-to-End Checks
- [x] GQL Fetching + updating should be done by reading data from CSV
- [x] ETL steps should be done by querying data from Tables
- [x] Update all downstream objects to query from PersistentData

### Data Store Fixes
- [x] Remove generate_view()

### Tests
- [x] Update csvs, table, and gql tests

idiom-bytes:
Fixes #742


pdr_backend/lake/gql_data_factory.py
```diff
@@ -130,17 +205,16 @@ def _update(self):
             fin_ut -- a timestamp, in ms, in UTC
         """
 
-        for _, table in self.record_config["tables"].items():
-            filename = table._parquet_filename()
-            st_ut = table._calc_start_ut(filename)
+        for table_name, table in self.record_config["tables"].items():
+            st_ut = table.csv_data_store.get_first_timestamp(table_name)
```
kdetry, line 209: On the following PR, I changed the logic which is at this part. We have to decide which one that we have to use

https://github.com/oceanprotocol/pdr-backend/pull/751/files#

PR #759, opened 2024-03-06
Fix #756 - Removing generate_view_name and simplify db.query()

### LINKED ISSUE #756: [ETL][Lake] DuckDB BaseDataStore - Remove base_data_store._generate_view_name() - yagni
### Motivation
Table names are getting hashed, makes it impossible to write a basic query that joins across tables.

```
@enforce_types
    def _generate_view_name(self, base_path=str) -> str:
        """
        Generate a unique view name for a given base path.
        @arguments:
            base_path - The base path to generate a view name for.
        @returns:
            str - A unique view name.
        """

        path = f"{self.base_directory}/{base_path}"
        hash_object = md5(path.encode())
        return f"dataset_{hash_object.hexdigest()}"
```

**For ETL, we need multiple views + join across them.**

### Requirements
Table names should be explicit, as-is. No black-box/indirections.

If a table is named `pdr_predictions` the final/completed/up-to-date/clean results production table inside duckdb should be named `pdr_predictions`.

The name should not be `dataset_m41l24kj1` or `dataset_pdr_predictions`.

### DoD
- [x] yagni / leading to problems w/ writing basic queries
- [x] All `final/production` tables should use static names
- [x] All temporary or system-level tables should contain a `_prefix`. Do not try to handle this prematurely.

idiom-bytes:
Fixes #756 

Changes proposed in this PR:
- [x] Removed black-box functionality around generate_view_name
- [x] Updated tests and verified that this is working as expected, by being able to directly query table names by their natural names (not a hash)
- [x] Future temporary/system-level tables should manage their own needs, such as adding a `_etl_build` prefix while ETL is building new tables to be joined w/ the live/production tables (with their natural names).

### DIFF
--- a/pdr_backend/lake/base_data_store.py
+++ b/pdr_backend/lake/base_data_store.py
@@ -20,20 +20,6 @@ class BaseDataStore:
             database=f"{self.base_directory}/duckdb.db"
         )  # Keep a persistent connection
 
-    @enforce_types
-    def _generate_view_name(self, base_path=str) -> str:
-        """
-        Generate a unique view name for a given base path.
-        @arguments:
-            base_path - The base path to generate a view name for.
-        @returns:
-            str - A unique view name.
-        """
-
-        path = f"{self.base_directory}/{base_path}"
-        hash_object = md5(path.encode())
-        return f"dataset_{hash_object.hexdigest()}"
-
     @abstractmethod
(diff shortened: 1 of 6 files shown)


kdetry (approved):
It looks completely okay to me

PR #781, opened 2024-03-12
Fix #772 - TableRegistry Implementation

### LINKED ISSUE #772: [Lake][SWE] Cleanup tables. Query them directly via DataStore, or create registry.
### Motivation
GQLDataFactory + ETL should clean up Tables further.
Right now, they are instantiated and cached in a way that is not very clean and leads to poor SWE.

### Towards a solution
1. Remove Tables() from GQL + ETL such that they don't have to be cached in the object.
2. Move Tables() to a registry or keep abstracting them further such that they can be queried/accessed from anywhere.
3. Leverage `CSVDataStore` and `PersistentDataStore` as singletons to improve how this is handled

### DoD
- [ ] There is no more duplication of Tables inside of GQL + ETL
- [ ] Cleanup etl.py `do_sync_step`, and other objects that are using Tables

kdetry:
Fixes #772 

Changes proposed in this PR:

- ***TableRegistry Singleton:*** A new file `table_registry.py` has been added. This file defines the `TableRegistry` singleton, which manages access to the table objects.

- ***Refactor ETL Process:*** The `etl.py` file has been modified to use the `TableRegistry` singleton instead of directly accessing `record_config[tables]`. This centralizes the table management and makes the ETL process cleaner and more maintainable.

- ***Refactor GQLDataFactory:*** The `gql_data_factory.py` file has been modified to work with the new `TableRegistry` singleton.

- ***Refactor Bronze Table:*** The `table_bronze_pdr_predictions.py` file has been modified to register itself with the TableRegistry singleton.

- ***Unit Tests:*** New unit tests have been added in test_table_registry.py to ensure the correct functionality of the `TableRegistry` singleton. Existing tests in `test_gql_data_factory.py` have been updated to work with the new `TableRegistry` singleton.

### DIFF
--- a/pdr_backend/lake/table_registry.py
+++ b/pdr_backend/lake/table_registry.py
@@ -1,0 +1,50 @@
+from typing import Optional, Dict, Tuple, List
+from polars.type_aliases import SchemaDict
+from enforce_typing import enforce_types
+
+from pdr_backend.ppss.ppss import PPSS
+from pdr_backend.lake.table import Table
+
+
+class TableRegistry:
+    _instance: Optional["TableRegistry"] = None
+    _tables: Dict[str, Table] = {}
+
+    def __new__(cls):
+        if cls._instance is None:
+            cls._instance = super(TableRegistry, cls).__new__(cls)
+        return cls._instance
+
+    @enforce_types
+    def register_table(self, table_name: str, table_args: Tuple[str, SchemaDict, PPSS]):
+        if table_name in self._tables:
+            pass
+        self._tables[table_name] = Table(*table_args)
+        return self._tables[table_name]
+
+    @enforce_types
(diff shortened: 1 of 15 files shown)


pdr_backend/lake/table_registry.py
```diff
@@ -0,0 +1,55 @@
+from typing import Optional, Dict, Tuple, List
+from polars.type_aliases import SchemaDict
+from enforce_typing import enforce_types
+
+from pdr_backend.ppss.ppss import PPSS
+from pdr_backend.lake.table import Table
+
+
+class TableRegistry:
+    _instance: Optional["TableRegistry"] = None
```
idiom-bytes, line 10: Great work! Simple implementation, clean functions, well-written.

Issue #677, opened 2024-02-26
[Warning, CI] In test_mathutil, get 'FutureWarning: Downcasting object dtype arrays'


trentmc:
### The problem

```console
pytest pdr_backend/util/test_noganache/test_mathutil.py
```

All the tests pass. But we get two similar warnings:
```                                                                                                                                                                                                                                   
pdr_backend/util/test_noganache/test_mathutil.py::test_all_nan__or_None
  /Users/trentmc/code/pdr-backend/pdr_backend/util/mathutil.py:26: FutureWarning: Downcasting object dtype arrays on .fillna, .ffill, .bfill is deprecated and will change in a future version. Call result.infer_objects(copy=False) instead. To opt-in to the future behavior, set `pd.set_option('future.no_silent_downcasting', True)`
    x = x.fillna(value=np.nan, inplace=False)

pdr_backend/util/test_noganache/test_mathutil.py::test_all_nan__or_None
  /Users/trentmc/code/pdr-backend/pdr_backend/util/mathutil.py:30: FutureWarning: Downcasting object dtype arrays on .fillna, .ffill, .bfill is deprecated and will change in a future version. Call result.infer_objects(copy=False) instead. To opt-in to the future behavior, set `pd.set_option('future.no_silent_downcasting', True)`
    x = x.fillna(value=np.nan)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

### TODOs

To fix: likely the best approach is to follow the suggestion "To opt-in to the future behavior, set ...".

trentmc:
Fixed in PR https://github.com/oceanprotocol/pdr-backend/pull/783, commit https://github.com/oceanprotocol/pdr-backend/commit/e9795f529140e838dab68b96eaf05924f70f66f0

PR #783, opened 2024-03-13
Towards #779: [Sim, make $] Easier benchmarking, via new 'multisim' CLI

### LINKED ISSUE #779: [Sim, make $] Easier benchmarking, via new 'multisim' CLI command
### Background / motivation

The challenge: we want to know the effects of various ppss parameters in simulation. 

We want to run sweeps across >1 ppss params, and output key results ($ made, accuracy, other) into a csv table.

Candidate approaches: (a) user does each run manually, (b) user has their own custom setups, (c) put it inside 'sim' CLI command, and (d) new 'mu' CLI command.

We'll use approach (d). Best UX, and effort to build it isn't high.

Q: why 'multisim' and not eg 'simsweep'? 
A: it makes it easier for us to extend the meta-algorithm that's doing sim-in-loop, while retaining the same ppss setup that specifies what parameters to sweep across (and values to test)
- For _this_ issue, it's simply a grid search that outputs to csv. 
- Soon, we will use an optimizer to auto-discover the best-performing combination of params. 
- It also opens up to having "fast sweep" that uses model-building to adaptively sample 

### TODOs / DoD

- [x] First cut: most things end-to-end. Refactoring at cli level.
- [x] multisim_ss properly iterates through combos, via PointMeta as multibit counter
- [x] All end-to-end. Includes: Properly updates PPSS each time; this means maybe getting deeper into dot notation.
- [x] Support different multifeeds
- ~~Optimize: multiprocess~~ --> moved to [#804](https://github.com/oceanprotocol/pdr-backend/issues/804)

trentmc:
Towards #779: everything but multiprocess
- New `MultisimSS` module, plus test
- New `MultisimEngine` module, plus test
- PPSS can now optionally be instantiated directly with a constructor dict `d`
- New `PointMeta` module, to define a space of >1 `Point` (also new).
- New utils: `obj_in_objlist()`, `keyval()`, `flat_to_nested_args()`, ..
- Support different multifeeds. Commit https://github.com/oceanprotocol/pdr-backend/pull/783/commits/40a79c4edbfde432c50d984b243005b4bb02eb84

Also fixes:
- [#797](https://github.com/oceanprotocol/pdr-backend/issues/797) "[CLI, UX] CLI help is too long for new users". Commit https://github.com/oceanprotocol/pdr-backend/pull/783/commits/5a5f288ac2f48f5a24edba385dfa4247646c199c
- [#798](https://github.com/oceanprotocol/pdr-backend/issues/798) "[Logging, UX] I can't tell which logger is outputting which text". Commit https://github.com/oceanprotocol/pdr-backend/pull/783/commits/23f295ce70f0a0665b589f7e6a8b48603f76f527
- [#789](https://github.com/oceanprotocol/pdr-backend/issues/789) "[Warning, Aimodel] test_sim_engine has UndefinedMetricWarning in classification". Commit https://github.com/oceanprotocol/pdr-backend/pull/783/commits/3c5a7dc7d00a2ca89cc03974b3259cdba0cd5c49
- [#677](https://github.com/oceanprotocol/pdr-backend/issues/677) "[Warning, CI] In test_mathutil, get 'FutureWarning: Downcasting object dtype arrays'". Commit https://github.com/oceanprotocol/pdr-backend/pull/783/commits/c8a9c9e09c628bbbbefab3274ee57b16eae98bf8


pdr_backend/sim/sim_state.py
```diff
@@ -26,6 +26,22 @@ def update(self, acc_est, acc_l, acc_u, f1, precision, recall):
         self.precisions.append(precision)
         self.recalls.append(recall)
 
+    def recent_metrics_names(self) -> List[str]:
+        return ["acc_est", "acc_l", "acc_u", "f1", "precision", "recall"]
+
+    def recent_metrics(self) -> List[Union[int,float]]:
+        """Return most recent classifier metrics"""
+        assert self.acc_ests, "must have >0 entries to call this"
+        return (
```
calina-c, line 35: I would be more comfortable with this being a dict or an Enum class. Otherwise it seems very easy to switch up the order and difficult to add more keys in the future, if needed

ppss.yaml
```diff
@@ -61,6 +61,23 @@ sim_ss: # sim only
       createMarketBuyOrderRequiresPrice": False
       defaultType: spot
 
+multisim_ss:
+  approach: simplesweep # simplesweep | fastsweep (future) | ..
```
calina-c, line 65: are these options documented somewhere?

pdr_backend/sim/sim_state.py
trentmc: I get your point.

But, this is just for dumping into a csv. I want to keep it light for now.

ppss.yaml
trentmc: About as much as the other options are documented. Which means "lightly". Specifically, in:
- ppss.yaml like you see
- and CAND_APPROACHES at the top of pdr_backend/sim/multisim_engine.py

And that's ok for now. I want to keep it light so that we can evolve rapidly. "Make $" is the goal.

BTW this PR is still in draft form. Things will keep evolving.

Issue #701, opened 2024-02-29
[Epic][Pdr bot] New smart contract, so user needs just one tx for 2-sided predictions across >1 feeds


trizin:
Build a smart contract to be able to submit two sided predictions to multiple feeds in a single tx.

### TODO:
- [x] Write the smart contract
- [x] Write tests for the smart contract
- [x] Add PredictionManager module into pdr-backend
- [x] Add a command into pdr-backend to deploy the manager contract.
- [x] Add tests for PredictionManager module
- [x] Update Predictoor PP to support "train_on", "feeds" structure for feeds.
- [x] Implement PredictionManager into predictoor agent

### Flow:
- if doesn't exist, pdr-backend deploys the prediction manager contract using `PRIVATE_KEY` and saves the address of the deployed contract in a file.
- pdr-backend interacts with the smart contract to submit predictions and claim payouts.

### Appendix: Background, motivation, details

(From slack conversation between Berkay & Trent 20240229)

_Berkay:_

Some predictoors use smart contracts, submitting 20 predictions in a single tx.
- Pros: no risk of losing stake due to failed tx, single tx to submit all, single predictoor private key.
- Cons: predictoors gotta deploy & manage a smart contract 

_Trent:_ That's cool! Where did you hear about this?

_Berkay:_ I inspected the top addresses in autobotocean

_Trent:_ We could even consider making this easy for people to do, ie deploy a contract that does predictions across 20 feeds, two-sided

_Berkay:_

Probably yes! we can abstract the complexity. Users would have to not lose their private key and the deployed contract address, and input them to pdr-backend. However the predictoors I've seen were 50-50s, is it feasible to compute the prediction for 10 feeds in a single pdr-backend process?

**Contract interface would look like:**
- `submit([]contracts, []stakesup, []stakesdown, epoch)`**: submit predictions for given epoch to contracts**
- `payout([]contracts, []epochs)`**: claim payout and send funds back to owner**

_Trent:_

Yes, exactly!

Perhaps make a github issue and we can then decide when to do.

If you want to do it sooner rather than later, go for it 

Related : if that existed then we could easily modify predictoor_agent.py to work across >1 feeds.

_Berkay:_

[Issue created!]

It's a huge friction remover and ensures stakes submitted on each side. 

But there's a challenge, calculating prediction for multiple pairs before the epoch closes. We could leverage threading maybe to run the calculations in parallel. Do you think this could be a problem? 

[...]

PR #786, opened 2024-03-13
#701 - Prediction manager contract

### LINKED ISSUE #815: Replace ocean_token with OCEAN in ganache conftest
Replace variable names "ocean_token" with "OCEAN"

### LINKED ISSUE #816: Rename predictoor_contract in ganache conftest to predictoor_contract1
Rename `predictoor_contract` in ganache conftest to `predictoor_contract1`

trizin:
**Towards** #701
Fixes #815
Fixes #816

Changes proposed in this PR:

- PredictionManager contract and other helper contracts.
- Deployer module
- Compiler script
- pdr CLI `deploy_prediction_manager` command
- Contract tests

### DIFF
--- a/pdr_backend/contract/pred_submitter_mgr.py
+++ b/pdr_backend/contract/pred_submitter_mgr.py
@@ -88,7 +88,7 @@ class PredSubmitterMgr(BaseContract):
         stakes_up_wei = [s.amt_wei for s in stakes_up]
         stakes_down_wei = [s.amt_wei for s in stakes_down]
         if self.config.is_sapphire:
-            res, tx = self.send_encrypted_tx(
+            _, tx = self.send_encrypted_tx(
                 "submit", [stakes_up_wei, stakes_down_wei, feeds, epoch]
             )
         else:


pdr_backend/prediction_manager/contracts/PredictionManager.sol
```diff
@@ -0,0 +1,135 @@
+// SPDX-License-Identifier: None
```
trentmc, line 1: It's a bit confusing how PredictionManager & Predictoor relate. And maybe they're more appropriate to be merged? The labels need improving for sure.

What makes this worse is: we already have `have contracts/predictoor_contract.py` & `PredictoorContract` class. However, those labels are inappropriate, and need to be renamed. I just created [#812](https://github.com/oceanprotocol/pdr-backend/issues/812) to fix it. In light of this prediction mgr work, we should do that issue asap.

I recommend: 
1. Do the quick fixes to address my comments (fairly light), then merge this PR. ("Ratchet up" philosophy)
2. Do # 812 (see above)
3. Determine better labels for PredictionManager & Predictoor, OR merge them, OR both
4. Continue the rest of the work on _this_ github issue

Appendix: 
- you should "try on" what happens if you merge Predictoor & PredictionManager. You'll have some Solidity that's slightly redundant for up-vs-down. But then can see everything "flat", which I expect to be much easier for people to read / understand. (Maybe this doesn't make sense too! Your call. But please try it if you think it does)
- Or, if we end up going for better labels, here's a first cut: Predictoor -> OneDirectionFeeds, PredictionManager -> BothDirectionFeeds.

trentmc (approved):
See comment "It's a bit confusing how PredictionManager & Predictoor relate". That's the top level. 

It suggests to go through four specific steps (1)..(4). I'm approving here because you don't need my approval to do the little fixes.

pdr_backend/prediction_manager/contracts/PredictionManager.sol
trentmc: > What about Predictoor & PredictionManager -> PredSubmitter & PredSubmitterManager?

This is already much better. Let's use those for now. If we think of something better yet, we can do another search & replace.

trentmc (changes requested):
This looks great!

One final thing: 
- 
"mgr" is the common abbreviation for "manager". This is an opportunity for us.
- therefore: let's rename "PredSubmitterManager" -> "PredSubmitterMgr". And all similar things, with "manager" -> "mgr". It will help keep the code compact yet readable.

Sorry that I didn't think of this sooner

Once this is in, I'm happy to approve!

trentmc (approved):

Issue #749, opened 2024-03-06
[Sim] Output sim_engine plots -> streamlit/plotly -> webapp


trentmc:
### Background / motivation

We're on a path to output predictoor analytics plots to webapp, using streamlit + plotly.

Currently the sim engine plots are output on-the-fly directly to the desktop using matplotlib defaults. They're not going to the browser. This was a fine enough starting point.

But there are issues:

- Because it keeps updating the plot in realtime, and takes up so much screen real estate, it's always popping over everything else. Annoying when you want to do other things at the same time
- A user reported ^ as well, in [#744](https://github.com/oceanprotocol/pdr-backend/issues/744)
- It doesn't work for Calina, because she needs to run compute remotely. What could work for is that the py code runs remotely, and then she can see the results in her local browser.
- Hard to add more plots, because of limited real estate. Yet we have more plots we want to add.
- Haven't tried interactivity yet, but it'll be hard

The solution: use streamlit.

That will address the issues above, and, a bonus: it helps us prototype for the analytics dapp. (We could even have them all in the same dapp. Pros and cons to this, decision is still TBD.)

### Pre-requisites

- [x] "Refactor plots in sim_engine" [#750](https://github.com/oceanprotocol/pdr-backend/issues/750)

### TODO

- [x] Migrate sim flow plots to a streamlit architecture, like the one we're going for with analytics app
- [x] Complete pickle/separation of concerns/overall architecture

### Resources

Key ref:
https://plotly.com/python/creating-and-updating-figures/

Maybe useful:
https://github.com/Paulescu/real-time-data-pipelines-in-python/tree/main

idiom-bytes:
How to organize all of this towards updating plots in a backend process + serving this via FE webapp
```
process 1 - etl.py - fetch + etl + update plots
process 2 - sim.py - running sim + updating plots
process 3 - streamlit frontend w/ different pages serving data + artifacts
```

trentmc:
@calina-c fyi I have completed the pre-requisite (refactor sim plot code)

calina-c:
Current behaviour:
- Static full-size image generated with matplotlib, containing multiple subplots
![sim_plot](https://github.com/oceanprotocol/pdr-backend/assets/4101015/7f569300-12d7-4d39-a716-423abf9d9abe)

- Difficult to separate and analyse the graphs, everything is displayed in one single picture and the user must zoom in and out
- No possibility for real-time updates, i.e. see the changes instantaneously during the simulation
[...]

PR #793, opened 2024-03-14
Issue #749 - Use streamlit to display graphs in sim plotter

calina-c:
Work on #749. Still using native matplotlib plots, but in a streamlit canvas.


pdr_backend/sim/sim_plotter.py
```diff
@@ -31,28 +34,33 @@ def __init__(
...
+            "aimodel_response": c5.empty(),
+            "f1_precision_recall_vs_time": streamlit.empty(),
+        }
+
+        self.figs = {}
+
+        for k, _ in self.canvas.items():
+            fig, ax = plt.subplots()
+            self.figs[k] = fig
+            setattr(self, f"ax_{k}", ax)
```
idiom-bytes, line 63: Awesome to see you have created a structure for a page layout.
@KatunaNorbert, please take a look at this for direction as to how to setup a page layout

idiom-bytes, line 63: @calina-c I'm a bit confused as to why we would register the figs into self.figs[dict] and the axes[] into a `self.ax_` property. 

I.E. 1 object is being split into 2 different interfaces, is there a reason for this?

calina-c: it was just an intermediary way of dealing with multiple figures. Once we get rid of the plot figures and keep streamlit ones, there is no need to keep figures. In addition to that, the architecture will also change by implementing a data layer, so this intermediary crutch of an implementation is moot.

pdr_backend/aimodel/aimodel_plotter.py
```diff
@@ -163,113 +178,110 @@ def _plot_aimodel_contour(
...
+    df.sort_values(by=["importance"], inplace=True, ascending=True)
+
+    chart = (
+        alt.Chart(df)
+        .mark_bar()
+        .encode(
+            x=alt.X("importance", title="Relative importance (%)"),
+            y=alt.Y("label", title=None, sort="-x", axis=alt.Axis(labelLimit=200)),
+        )
+        .properties(title="Variable importances")
```
idiom-bytes, line 272: spent some extra time reviewing this code block, i think the logic checks out although i guess we're not truncating anymore based on importance

assuming this, i think it looks good

pdr_backend/sim/sim_plotter.py
```diff
@@ -25,277 +20,235 @@ class SimPlotter:
...
-@enforce_types
-def _set_title(ax, s: str):
-    ax.set_title(s, fontsize=FONTSIZE, fontweight="bold")
+        chart = alt.Chart(df, title=s).mark_circle().encode(x="prob(up)", y=y)
 
+        ref_line = (
+            alt.Chart(pd.DataFrame({y: [0]}))
+            .mark_rule(color="grey", strokeDash=[10, 10])
+            .encode(y=y)
+        )
```
idiom-bytes, line 247: I reviewed the plot render functions and they seem to check out

pdr_backend/sim/sim_plotter.py
```diff
@@ -25,277 +20,235 @@ class SimPlotter:
...
-        if do_save_plot:
-            img_filename = self.ppss.sim_ss.unique_final_img_filename()
-            savefig(img_filename)
-
-        # wrapup for reloop
-        if do_show_plot:
-            plt.pause(0.001)
-        self.computed_plot_before = True
-
-        return img_filename
```
idiom-bytes, line 125: I verified this moved to sim_plots and based on the plots rendered, I think it checks out

idiom-bytes (approved):
I reviewed every file, updates, and how the logic was restructured. Although there was a significant amount of cleanup, the core logic looks accurate.

Approving the PR.

pdr_backend/aimodel/aimodel_plotter.py
calina-c: not truncating anymore since there is no need for the graph to "fit" inside a small container. I can not guarantee it looks "perfect" for edge cases like e.g. an enormous amount of variables, but it won't overlap any text. Also, in the future we might be selecting for specific variables to plot, so this logic will probably be something else entirely.

trentmc:
It will be a common case to have an enormous number of variables. That's why it truncated in the original code.

@calina-c  can you please create another github issue to handle this? Or do another bit for this issue

PR #794, opened 2024-03-14
Fix #771 - Temp Table Logic for the build step and the bronze step start/finish parameter calculations

### LINKED ISSUE #771: [Lake][ETL] DuckDB - Implement ETL SQL build strategy w/ temp tables
### Motivation
Right now, ETL expects you to modify `st_ts` and `end_ts` in ppss.yaml to manage ETL incrementally.
This is a lot of effort and right now is not handled very cleanly.

Let's look at the following example so we can understand this a bit better.
In the first run we will have something like:

```
lake:
  st_ts: '06-01-2023'
  end_ts: '02-15-2024'
```

In the second run we want to have something like this, because we've already processed all historical data
```
lake:
  st_ts: '02-15-2024'
  end_ts: '02-16-2024'
```

However, the ux for the lake tries to be a bit more end-to-end by simply stating...
**This is a much better/cleaner way of doing things.**
```
lake:
  st_ts: '06-01-2023'
  end_ts: 'now'
```

**[What's the current challenge]**
Raw Lake Tables: In the lake, **ohlcv/subgraph raw tables** can have different max(timestamp). **So each source data has it's own checkpoint** by simply looking at it's local csv/data and getting the latest record/timestamp.

Build Tables: In the lake, **etl bronze/silver/gold tables** need to be enforced/guaranteed such that all records up-to **the max(timestamp) across all tables, have been processed**.

Because of this, we're currently addressing this by **recalculating all final records in the ETL step.**

### SQL Build Step
Here is a simple way to enforce a clean-build, with any "step" that we implement. Whether with OHCLV, GQL, or ETL. The simplest way to summarize it is:
1. Calculate the step from_timestamp and to_timestamp
2. Run the step logic, such that all records are being yielded to `_temp_table_name`
3. Once the logic completes, (1) append records from `_temp_table_name` to `table_name`

**In the case records need to be overwritten:**
3. Once the logic completes, (1) upsert records from `_temp_table_name` to `table_name`

**In case the records need to be swapped:**
3. Once the logic completes, (1) drop `table_name` (2) rename `_temp_table_name` to `table_name`

### Source/raw data vs. ETL SLA

**Source/Raw Data Fetching => Lake SLA**
1. All raw tables can be in different timestamps.
2. Data Factories reads from their CSVDataStores to find the max(timestamp), and resume fetching up-to-now.
3. Each Data Factory then tries to fetch remote records, and update their local data store

**Raw Data => ETL SLA**
1. In order to start the ETL, we need to get 2 values: `from_timestamp` and `to_timestamp`. This will provides us with the timerange to process all raw-data, and to update the ETL tables such that they are all updated to the latest `to_timestamp`
2. To get `to_timestamp`, we look across all source/raw Data Factory tables inside DuckDB, and find their common-denominated-timestamp: `min(source_tables[max_timestamp])`.
3. To get `from_timestamp`, we look across all ETL Tables inside DuckDB, and find the max common-timestamp between them: `max(etl_tables[max_timestamp]).
4. We can then start the ETL process by using the raw_data `from_timestamp` => `to_timestamp` to process the new bronze/silver/gold records.
5. All rows processed and yielded then make their ways to temporary `_build_etl_tables`
6. Some ETL queries may need to be updated to union data from existing data `etl_table_names` and temp build data `_build_etl_table_names`, such that the build step has access to all data.
7. Once all ETL queries have completed, all `_build_table_names` should be yielded and finalized.
8. We can then apply the process of appending/upserting/or replacing records in the live `etl_table_names`, such that the ETL step finalizes.

### DoD
- [x] Raw tables continue to update on their own and be at different timestamps.
- [x] ETL tables can only be updated if raw-tables are up-to-date.
- [x] ETL tables are kept in-sync with each other, such that they are all processed up-to-the-same timestamp.
- [x] Implement build step to use SQL and temp tables.
- [x] ETL logic should look to see if there are already `_build` tables that exist.
- [x] When ETL is starting. if `_build` tables already exist, drop them and start a fresh build.
- [x] ETL updates should use `from_timestamp` by calculating `max(etl_tables_max_timestamp)`. This lets us know all the old records not to be processed.
- [x] ETL updates should use `to_timestamp` by calculating `min(max(source_tables_max_timestamp))`. This lets us know all the latest downloaded records that need to be processed.

kdetry:
Fixes #771 

Changes proposed in this PR:

- The data fetched during the build process is written to build tables.
- The logic to move the DB table after the build process has been developed.
- Some helpers and tests are added

PS: broken sync process on the ETL is fixed

### DIFF
--- a/pdr_backend/lake/etl.py
+++ b/pdr_backend/lake/etl.py
@@ -14,6 +14,8 @@ from pdr_backend.lake.table_pdr_predictions import predictions_table_name
 from pdr_backend.lake.table_pdr_subscriptions import subscriptions_table_name
 from pdr_backend.lake.table_pdr_truevals import truevals_table_name
 from pdr_backend.lake.plutil import get_table_name
+
+
 class ETL:
     """
     @description
@@ -44,7 +46,7 @@ class ETL:
             payouts_table_name,
             predictions_table_name,
             subscriptions_table_name,
-            truevals_table_name
+            truevals_table_name,
         ]
 
     def _check_build_sql_tables(self):
--- a/pdr_backend/lake/etl.py
+++ b/pdr_backend/lake/etl.py
@@ -83,6 +83,10 @@ class ETL:
         print("do_etl - Start ETL.")
 
         try:
+            # Check if the SQL tables exist and drop them if they do
+            self._check_build_sql_tables()
+            print("do_etl - Checked build tables.")
+
             # Sync data
             self.gql_data_factory.get_gql_tables()
             print("do_etl - Synced data.")
--- a/pdr_backend/lake/etl.py
+++ b/pdr_backend/lake/etl.py
@@ -8,8 +8,12 @@ from pdr_backend.lake.table_bronze_pdr_predictions import (
     get_bronze_pdr_predictions_data_with_SQL,
 )
 from pdr_backend.lake.table_registry import TableRegistry
-
-
+from pdr_backend.lake.persistent_data_store import PersistentDataStore
+from pdr_backend.lake.table_pdr_payouts import payouts_table_name
+from pdr_backend.lake.table_pdr_predictions import predictions_table_name
+from pdr_backend.lake.table_pdr_subscriptions import subscriptions_table_name
+from pdr_backend.lake.table_pdr_truevals import truevals_table_name
+from pdr_backend.lake.plutil import get_table_name
 class ETL:
     """
     @description
@@ -35,6 +39,40 @@ class ETL:
             ),
         )
 
+        self.build_table_names = [
+            bronze_pdr_predictions_table_name,
+            payouts_table_name,
+            predictions_table_name,
+            subscriptions_table_name,
+            truevals_table_name
+        ]
+
+    def _check_build_sql_tables(self):
+        """
+        @description
+            Check if the SQL tables are built
+            If not, build them
+            If exists, drop them and rebuild
+        """
+        # drop the tables if it exists
+
+        pds = PersistentDataStore(self.ppss.lake_ss.parquet_dir)
+        for table_name in self.build_table_names:
+            pds.drop_table(get_table_name(table_name, build_mode=True))
+
+    def _move_build_tables_to_permanent(self):
+        """
+        @description
+            Move the build tables to permanent tables
+        """
+
+        pds = PersistentDataStore(self.ppss.lake_ss.parquet_dir)
+        for table_name in self.build_table_names:
+            pds.move_table_data(
+                get_table_name(table_name, build_mode=True),
+                table_name,
+            )
+
     def do_etl(self):
         """
         @description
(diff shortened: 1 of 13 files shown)


pdr_backend/lake/etl.py
```diff
@@ -80,5 +129,6 @@ def update_bronze_pdr_predictions(self):
         print("update_bronze_pdr_predictions - Update bronze_pdr_predictions table.")
         data = get_bronze_pdr_predictions_data_with_SQL(self.ppss)
         TableRegistry().get_table(bronze_pdr_predictions_table_name).append_to_storage(
-            data
+            data,
+            build_mode=True,
```
idiom-bytes, line 133: 1. While running ETL, append_to_db(build_mode = True).
2. After successfully building all tables, dump from tables => CSV (so that we can guarantee no errors)
3. Then, finally, move all records from build_tables => final_tables

These 3 steps should guarantee a clean build, clean CSV, and no hanging records.

pdr_backend/lake/etl.py
```diff
@@ -121,13 +129,80 @@ def do_bronze_step(self):
...
+        @returns
+            The max timestamp values from the tables
+        """
+
+        pds = PersistentDataStore(self.ppss.lake_ss.parquet_dir)
+        max_timestamp_query = "SELECT MAX(timestamp) as max_timestamp FROM {}"
+        values = {}
+        for table_name in table_names:
+            calc_table_name = get_table_name(table_name, build_mode)
+            result = pds.query_data(max_timestamp_query.format(calc_table_name))
```
idiom-bytes, line 151: To improve:
- Build querystr by iterating through tables
- Reduce query to 1 execution

pdr_backend/lake/etl.py
```diff
@@ -121,13 +129,80 @@ def do_bronze_step(self):
...
+            ETL updates should use from_timestamp by calculating
+            max(etl_tables_max_timestamp).
+            ETL updates should use to_timestamp by calculating
+            min(max(source_tables_max_timestamp)).
+        """
+        from_values = self._get_max_timestamp_values_from(
+            self.bronze_table_names
+        ).values()
+        from_timestamp = (
+            min(from_values)
```
idiom-bytes, line 172: When query is improved, min can be moved to it.

pdr_backend/lake/etl.py
```diff
@@ -121,13 +129,80 @@ def do_bronze_step(self):
...
+        )
+
+        to_values = self._get_max_timestamp_values_from(
+            self.raw_table_names, build_mode=True
+        ).values()
+        to_timestamp = (
+            min(to_values)
+            if None not in to_values
+            else datetime.strptime(self.ppss.lake_ss.fin_timestr, "%Y-%m-%d_%H:%M")
+        )
```
idiom-bytes, line 184: This is what I meant by us being able to leverage SQL + Views (not tables) to simplify this.

pseudo code 1 - create view that unions both tables so all data sits together
views should not take more storage/mem, it's "an alias for two different tables"
```
create view etl_bronze_pdr_predictions (
select *
from bronze_pdr_predictions
union etl_bronze_pdr_predictions
);
```

pseudocode 2 - reduce query across tables to 1 scan/execute
```
"SELECT MAX(timestamp) as max_timestamp FROM etl_bronze_pdr_predictions"
```

PR #819, opened 2024-03-19
Fix #809 - Clean entire table queries

### LINKED ISSUE #809: [Lake][ETL] DuckDB Integration - Cleanup any refs to `select * from tables`
### Background

As a result of DuckDB integration, there is some remaining code that may `select * from table` as a way to "get a lazy pointer".

### DoD
- [x] Clean up any final queries that fetch the whole database system. No `select * from table_name` anywhere.

kdetry:
Fixes #809 

Changes proposed in this PR:

- The `get_records` method of the `Table` class has been removed.
- The `PredFilter` class has been removed and replaced with SQL queries in places where the class was used.
- The tests have been fixed.
- The test methods that query the entire table have not been changed as it was deemed unnecessary.

### DIFF
--- a/pdr_backend/lake/table.py
+++ b/pdr_backend/lake/table.py
@@ -1,5 +1,4 @@
 import logging
-from typing import Optional
 import polars as pl
 from polars.type_aliases import SchemaDict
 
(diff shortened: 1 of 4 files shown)


idiom-bytes (approved):
lgtm

pdr_backend/lake/table.py
```diff
@@ -60,38 +59,3 @@ def _append_to_db(
...
-
-    def get_pds_last_record(self) -> Optional[pl.DataFrame]:
-        """
-        Get the last record from the persistent data store
-
-        @returns
-            pl.DataFrame
-        """
-
-        query = f"SELECT * FROM {self.table_name} ORDER BY timestamp DESC LIMIT 1"
```
idiom-bytes, line 72: Although this only returns 1 record, I agree that this should be removed because it's not being used anywhere.

idiom-bytes:
Looks good, although CICD isn't working, this checks out.
Merging.

PR #824, opened 2024-03-20
Fix #806: DRY violations in specifying exchange

### LINKED ISSUE #806: [UX, Sim, Trader] DRY violations in specifying exchange
### The issue

Here's a snippet from ppss.yaml:
```yaml
trader_ss:
  feed: binance BTC/USDT c 5m
  sim_only:
    buy_amt: 1000 USD # buy this amount in each epoch
    fee_percent: 0.0 # simulated % fee
    init_holdings:
      - 100000 USDT
      - 2 BTC
  bot_only:
    min_buffer: 30 # in s. only trade if there's > this time left
    max_tries: 10 # max no. attempts to process a feed
    position_size: 3 # buy/sell this amount in each epoch
  exchange_only:
    timeout: 30000
    options:
      createMarketBuyOrderRequiresPrice": False
      defaultType: spot

sim_ss: # sim only
  do_plot: True
  log_dir: logs
  final_img_filebase: final_img # for name log_dir/{final_img_filebase}_{NR}.png
  test_n: 5000 # number of epochs to simulate
  tradetype: histmock # histmock | livemock | livereal
  exchange_only:
    timeout: 30000
    options:
      createMarketBuyOrderRequiresPrice": False
      defaultType: spot
```

The problem: `exchange_only` and its sub-values are in two places.

This gets reflected further downstream too.

This is a DRY violation. And, it's annoying from a UX perspective too: the user should really only have to change one place.

### Candidate solutions

1. Create a new module ExchangeSS that has `exchange_only` and all the info below that. Then, both Trader and Simulator see this
2. Only have the info in TraderSS. Then, SimulatorSS also uses the info

### TODO

- [ ] Quick implementation experiment trying cand (1): will it be clean? 
- [ ] If yes to ^, build (1). Else build (2)

trentmc:
Fixes #806


calina-c:
@trentmc I took the liberty of continuing this and fixing the tests. I think the only change warranted is to have the exchange mgr functionality inside exchange ss. My implementation is a bit clunky where I define the manager directly in ppss, but maybe you have something else in mind?

Either way, it should just be minor changes from here. We can merge it now (after some code review from you and Roberto) and do a follow-up later, or we can keep it until you're back from holiday. I think it's your call.

trentmc:
Thanks for running with this, @calina-c 

I really prefer that none of the SS modules do any work themselves. They are meant to hold strategy parameters, not do work.

This includes : exchange manager should be its own class with its own module. 

Over time it will grow in scope as well, as we add in support for dydx and more.

Could you update to re-separate please?

calina-c:
> Thanks for running with this, @calina-c
> 
> I really prefer that none of the SS modules do any work themselves. They are meant to hold strategy parameters, not do work.
> 
> This includes : exchange manager should be its own class with its own module.
> 
> Over time it will grow in scope as well, as we add in support for dydx and more.
> 
> Could you update to re-separate please?

It is still separated, but I'm wondering if I should initialize the manager every time based on ppss? It seems to be like a singleton pattern, where the manager should be common to multiple dependencies. Right now, it's: `exchange = ppss.exchange_mgr.exchange()`. Should it be:

```
exchange_mgr = ExchangeManager(ppss.exchange_mgr_ss)
exchange = exchange_mgr.exchange()
```

instead?

calina-c:
@trentmc corrected :)

trentmc (commented):
@calina-c your changes lgtm

calina-c (approved):

PR #836, opened 2024-03-28
Fix #811 - Improve ETL failure handling and extend test coverage

### LINKED ISSUE #811: [Lake][ETL] Extend ETL test coverage to simulate multiple runs.
### Background / motivation

As laid out in #810, the pipeline is going to run into issues as it assumes that all data is going to exist in the same step. However, this is not the case and there will be existing data in `live_` table that needs to be processed.

To address this, we should extend test coverage such that it simulates multiple builds and covers this scenario.

**silver-predictions is a great example of this dependency in practice**

### TODOs / DoD
- [x] Update test_etl.py or add new test coverage that executes a multi-stage ETL pipeline
- [x] First run needs to contain data that will be required by the second run
- [x] Second run should depend on data that is already existent on `live_` tables that were yielded as a result of first run
- [x] Test coverage should validate that ETL is able to process data from runs that completed (1st run), in addition to data that was recently generated into `build_` tables as a result of the current run (2nd run).
- [ ] Test coverage should validate that `silver-prediction` table is working as expected

idiom-bytes:
Fixes #811

### DIFF
--- a/pdr_backend/lake/etl.py
+++ b/pdr_backend/lake/etl.py
@@ -217,10 +217,6 @@ class ETL:
             fin_ms=UnixTimeMs.from_dt(fin_timestamp),
         )
 
-        print(f">>>>> st_timestamp is {st_timestamp}")
-        print(f">>>>> fin_timestamp is {fin_timestamp}")
-        print(f">>>>> data is {data}")
-
         print(f"update_bronze_pdr_predictions - data: {data}")
         TableRegistry().get_table(bronze_pdr_predictions_table_name)._append_to_db(
             data,
--- a/pdr_backend/lake/etl.py
+++ b/pdr_backend/lake/etl.py
@@ -49,11 +49,11 @@ class ETL:
                     bronze_pdr_predictions_schema,
                     self.ppss,
                 ),
-                bronze_pdr_slots_table_name: (
-                    bronze_pdr_slots_table_name,
-                    bronze_pdr_slots_schema,
-                    self.ppss,
-                ),
+                # bronze_pdr_slots_table_name: (
+                #     bronze_pdr_slots_table_name,
+                #     bronze_pdr_slots_schema,
+                #     self.ppss,
+                # ),
             }
         )
 
@@ -67,12 +67,12 @@ class ETL:
 
         self.bronze_table_getters = {
             bronze_pdr_predictions_table_name: get_bronze_pdr_predictions_data_with_SQL,
-            bronze_pdr_slots_table_name: get_bronze_pdr_slots_data_with_SQL,
+            # bronze_pdr_slots_table_name: get_bronze_pdr_slots_data_with_SQL,
         }
 
         self.bronze_table_names = list(self.bronze_table_getters.keys())
 
-        self.temp_table_names = [*self.bronze_table_names, *self.raw_table_names]
+        self.temp_table_names = [*self.bronze_table_names]
 
     def _drop_temp_sql_tables(self):
         """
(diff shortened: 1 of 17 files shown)


pdr_backend/lake/etl.py
```diff
@@ -69,9 +70,11 @@ def __init__(self, ppss: PPSS, gql_data_factory: GQLDataFactory):
             bronze_pdr_slots_table_name: get_bronze_pdr_slots_data_with_SQL,
         }
 
+        print(f"self.bronze_table_getters: {self.bronze_table_getters}")
+
         self.bronze_table_names = list(self.bronze_table_getters.keys())
 
-        self.temp_table_names = [*self.bronze_table_names, *self.raw_table_names]
+        self.temp_table_names = [*self.bronze_table_names]
```
kdetry, line 77: you removed raw_table_names from the temp_table_names but we have move the temp data to the prod data tables when the etl process is completed

pdr_backend/lake/etl.py
```diff
@@ -214,10 +218,7 @@ def _calc_bronze_start_end_ts(self):
             else datetime.strptime(self.ppss.lake_ss.st_timestr, "%Y-%m-%d_%H:%M")
         )
 
-        to_values = self._get_max_timestamp_values_from(
-            self.raw_table_names, TableType.TEMP
-        ).values()
-
+        to_values = self._get_max_timestamp_values_from(self.raw_table_names).values()
```
kdetry, line 221: the raw data tables are not moved to the prod tables yet during the etl process, so we have to use temp tables

pdr_backend/lake/table_bronze_pdr_slots.py
```diff
@@ -35,20 +35,20 @@ def get_bronze_pdr_slots_data_with_SQL(
...
 
-    pdr_slots_table_name = get_table_name(slots_table_name, TableType.TEMP)
-    pdr_truevals_table_name = get_table_name(truevals_table_name, TableType.TEMP)
-    _bronze_pdr_predictions_table_name = get_table_name(
-        bronze_pdr_predictions_table_name, TableType.TEMP
+    pdr_slots_table_name = get_table_name(slots_table_name)
+    pdr_truevals_table_name = get_table_name(truevals_table_name)
+    pdr_payouts_table_name = get_table_name(payouts_table_name)
+    etl_bronze_pdr_predictions_table_name = get_table_name(
+        bronze_pdr_predictions_table_name, TableType.ETL
```
kdetry, line 42: are we sure?

kdetry (approved):
I have shared my concerns. If you think there is no issue with the points that I criticised, you can merge the PR

pdr_backend/lake/etl.py
idiom-bytes: Yes, please remember that raw_table_names will now have already moved from _temp -> prod inside of GQLDataFactory, before ETL kicks off.
image

So when ETL runs, the only temp_tables will be the bronze, silver, gold.

pdr_backend/lake/table_bronze_pdr_slots.py
idiom-bytes: yes, please review the diagram above.

pdr_backend/lake/etl.py
idiom-bytes: The raw data tables move to prod tables during GQLDataFactory.
I have updated the tests to make sure that all flows/tests are following this change.

Please see the diagram I shared.

PR #851, opened 2024-04-02
Adds snapshots slider for sim engine plots

calina-c:


sim_plots.py
```diff
@@ -98,8 +82,33 @@
...
+
+    load_canvas_on_state(new_ts)
     last_ts = new_ts
 
     if last_ts == "final":
-        title.title("Final sim state")
+        snapshots = SimPlotter.available_snapshots()
+        timestamp = inputs.select_slider("Go to snapshot", snapshots, value="final")
+        st, new_ts = sim_plotter.load_state(timestamp)
+        load_canvas_on_state(timestamp)
```
idiom-bytes, line 113: i might not have waited until the very end to see this come up.
This looks good, but i did not verify/test the slider.

pdr_backend/sim/sim_plotter.py
```diff
@@ -24,7 +25,20 @@ def __init__(
         self.st = None
         self.aimodel_plotdata = None
 
-    def load_state(self):
+    @staticmethod
+    def available_snapshots():
```
idiom-bytes, line 29: checked this against pickle/save code in sim_engine, and it looks good

i check test_sim_plotter.py and it's looking empty
we might want to just a dd a bit of coverage here in the future

idiom-bytes (approved):
tested, everything worked except i didn't wait for engine to end so slider would show up

calina-c:
> tested, everything worked except i didn't wait for engine to end so slider would show up

@idiom-bytes If you want to delve into this some more, you can run the sim for as much time as comfortable. Then go into the sim_state folder and rename the last state for `st_{timestamp}.pkl` and `aimodel_plotter_{timestamp}.pkl` as `st_final.pkl` and `aimodel_plotter_final.pkl` respectively. That will trick the streamlit runner into thinking the simulation has finished. I am merging this in the meantime, but just letting you know if you want to follow up on it.

PR #860, opened 2024-04-04
Use ETL to fetch data and calculate accuracy from duckDB

### LINKED ISSUE #854: Update Accuracy API to use ETL + DuckDB
On the ETL + DuckDB flow the Accuracy data endpoint that we use to serve accuracy data to pdr-web trough an API it's broken because of some slot fetching updates. Before merging the ETL + DuckDB we should make sure that the Accuracy endpoint is also working.
While on this we should also use the data from the ETL to calculate the Accuracy instead of fetching and storing the data separately.

Dod:
- [x] Fix Accuracy endpoint so it fetches and serves data correctly with the new subgraph fetching changes
- [ ] Integrate ETL + DuckDB so the accuracy endpoint uses ETL to fetch and store data

KatunaNorbert:
Fixes #854.

### DIFF
--- a/pdr_backend/accuracy/app.py
+++ b/pdr_backend/accuracy/app.py
@@ -1,7 +1,7 @@
 import logging
 import threading
 from datetime import datetime, timedelta
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Callable
 from enforce_typing import enforce_types
 from flask import Flask, jsonify
 from pdr_backend.lake.table_pdr_slots import slots_table_name
@@ -79,18 +79,13 @@ def process_single_slot(
         int(slot.ID.split("-")[1])
     )  # Using dot notation for attribute access
 
-    today = 0
-    yesterday = 0
-
     if (
         end_of_previous_day_timestamp - SECONDS_IN_A_DAY
         < timestamp
         < end_of_previous_day_timestamp
     ):
-        yesterday += 1
         staked_yesterday += float(slot.roundSumStakes)
     elif timestamp > end_of_previous_day_timestamp:
-        today += 1
         staked_today += float(slot.roundSumStakes)
 
     prediction_result = calculate_prediction_result(
@@ -267,8 +262,10 @@ def fetch_statistics_using_ETL():
         threading.Event().wait(300)  # Wait for 5 minutes (300 seconds)
 
 
-def filter_func(spe):
-    return lambda item: int(item.get("seconds_per_epoch", 0)) == spe
+@enforce_types
+def filter_func(seconds_per_epoch: int) -> Callable[[Any], bool]:
+    # filter out all the items that have seconds_per_epoch different that the specified parameter
+    return lambda item: int(item.get("seconds_per_epoch", 0)) == seconds_per_epoch
 
 
 @enforce_types
@@ -305,7 +302,7 @@ def transform_slots_to_statistics(all_slots: List[PredictSlot]):
         "0x4ac2e51f9b1b0ca9e000dfe6032b24639b172703", network_param
     )
 
-    contracts_list_unfiltered = fetch_contract_id_and_spe(
+    contracts_list_unfiltered: List[ContractIdAndSPE] = fetch_contract_id_and_spe(
         contract_addresses,
         network_param,
     )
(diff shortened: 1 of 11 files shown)


pdr_backend/accuracy/app.py
```diff
@@ -77,13 +79,18 @@ def process_single_slot(
         int(slot.ID.split("-")[1])
     )  # Using dot notation for attribute access
 
+    today = 0
```
calina-c, line 82: these variables are named like they represent times, but they are ints. What are they supposed to hold?

pdr_backend/accuracy/app.py
```diff
@@ -239,12 +248,31 @@ def calculate_timeframe_timestamps(
         # else timedelta(days=1)
     )
     start_ts = UnixTimeS(int((datetime.utcnow() - time_delta).timestamp()))
-
     return start_ts, end_ts
 
 
 @enforce_types
-def save_statistics_to_file():
+def fetch_statistics_using_ETL():
```
calina-c, line 255: shouldn't the ppss be customisable?

pdr_backend/accuracy/app.py
```diff
@@ -239,12 +248,31 @@ def calculate_timeframe_timestamps(
...
+    )
+
+    gql_data_factory = GQLDataFactory(ppss)
+    etl = ETL(ppss, gql_data_factory)
+    while True:
+        etl.do_etl()
+        threading.Event().wait(300)  # Wait for 5 minutes (300 seconds)
+
+
+def filter_func(spe):
```
calina-c, line 270: please document and type hint/enforce

pdr_backend/accuracy/app.py
```diff
@@ -342,12 +352,37 @@ def serve_statistics_from_file():
...
+    end_ts = UnixTimeS(int(datetime.utcnow().timestamp()))
     try:
-        with open(JSON_FILE_PATH, "r") as f:
-            data = json.load(f)
-            response = jsonify(data)
-            response.headers.add("Access-Control-Allow-Origin", "*")  # Allow any origin
-            return response
+        db_conn = PersistentDataStore("./lake_data", read_only=True)
+        slots_table = db_conn.query_data(
+            f"""
```
calina-c, line 360: please align opening and close quotes

pdr_backend/accuracy/app.py
```diff
@@ -342,12 +352,37 @@ def serve_statistics_from_file():
     If the file cannot be read or another error occurs, it returns a 500 Internal Server Error.
     """
 
+    start_ts = UnixTimeS(int((datetime.utcnow() - timedelta(weeks=4)).timestamp()))
```
calina-c, line 355: for better readability, I would separate the date times. I.e.
four_weeks_ago = datetime.utcnow() - timedelta(weeks=4)
start_ts = UnixTimeS(int(four_weeks_ago.timestamp()))

and same for end

pdr_backend/lake/etl.py
```diff
@@ -168,30 +168,46 @@ def _get_max_timestamp_values_from(
...
+        all_db_tables = PersistentDataStore(
+            self.ppss.lake_ss.lake_dir
+        ).get_table_names()
+
+        queries = []
+
+        for table_name in table_names:
+            table_name_with_type = get_table_name(table_name, table_type)
+            if table_name_with_type not in all_db_tables:
+                print(
```
calina-c, line 180: please replace all print statements with logging

pdr_backend/lake/etl.py
```diff
@@ -218,11 +234,17 @@ def _calc_bronze_start_end_ts(self):
             else datetime.strptime(self.ppss.lake_ss.st_timestr, "%Y-%m-%d_%H:%M")
         )
 
+        fin_timestamp = (
```
calina-c, line 237: this is very confusing. It seems in one case, you assign this variable the result of strptime, and in another case the value of strftime. But those are not the same data type. Should this be a string or a date time?

ppss.yaml
```diff
@@ -5,7 +5,7 @@ lake_ss:
     - binance BTC/USDT ETH/USDT 5m
   #    - binance BTC/USDT ETH/USDT BNB/USDT XRP/USDT ADA/USDT DOGE/USDT SOL/USDT LTC/USDT TRX/USDT DOT/USDT 5m
   #    - kraken BTC/USDT 5m
-  st_timestr: 2023-06-01_00:00 # starting date for data
+  st_timestr: 2024-04-10_00:00 # starting date for data
```
calina-c, line 8: should this change be committed?

calina-c (changes requested):

pdr_backend/accuracy/app.py
KatunaNorbert: I think it's good as it is for now, the scope of the API is quite specific

pdr_backend/lake/etl.py
KatunaNorbert: I guess strptime, some of the canges are from other merges in the base PR

ppss.yaml
KatunaNorbert: no, it shouldn't

PR #866, opened 2024-04-08
Fix #859 - ETL + DuckDB encountered issues

### LINKED ISSUE #859: ETL + DuckDB encountered issues
While testing the ETL + DuckDB integration from this branch I run into the following issues:

- [ ] The ETL fetched all the predictions until today and then started fetching all predictions again from the start. Check image:
- [x] Looks like the 'now' value of the fin_timestr of lake_ss can't be interpreted and breaks the flow:
- [x] If the fetched data length is 0 the table is not created and the ETL flow crashes. Run into this with the subscriptions table:

-------
- [ ]  The logic to retrieve the last timestamp from the CSV data store is not working correctly. It has to be fixed

kdetry:
Fixes #859 

Changes proposed in this PR:

- the empty table issue is fixed
- the "now" value issue on the etl is fixed

[idiom-bytes updates]
- [x] merged my larger PR into this PR
- [x] resolved conflicts, iterated on the logic, and (2) cleaned up a bunch of code.
- [x] i noticed that there were (2) a bunch of [variables, function args, return values] called timestamp but weren't... they were datetime str, timestr, and other values.
- [x] added a bunch of type-checks, asserts, and updated this everywhere so that if something is called a timestamp, I can expect it to be a UnixTimeS or UnixTimeMs
- [x] updated tests so dummy_tables use INT64 (actual unix timestamp) rather than a TIMESTAMP (which is a datetime) and was causing conflicts with the code, introducing a mixture-of-concerns that should not be there

### DIFF
--- a/pdr_backend/lake/etl.py
+++ b/pdr_backend/lake/etl.py
@@ -164,13 +164,31 @@ class ETL:
             "SELECT '{}' as table_name, MAX(timestamp) as max_timestamp FROM {}"
         )
 
-        queries = [
-            max_timestamp_query.format(
-                get_table_name(table_name, table_type),
-                get_table_name(table_name, table_type),
+        all_db_tables = PersistentDataStore(self.ppss.lake_ss.lake_dir).get_table_names()
+
+        queries = []
+
+        for table_name in table_names:
+            table_name_with_type = get_table_name(table_name, table_type)
+            if table_name_with_type not in all_db_tables:
+                print(
+                    f"_get_max_timestamp_values_from - Table {table_name} does not exist."
+                )
+                continue
+
+            queries.append(
+                max_timestamp_query.format(
+                    table_name_with_type, table_name_with_type
+                )
             )
-            for table_name in table_names
-        ]
+
+        none_values: Dict[str, Optional[datetime]] = {
+            table_name: None for table_name in table_names
+        }
+
+        if len(queries) == 0:
+            return none_values
+
         final_query = " UNION ALL ".join(queries)
 
         print(f"_get_max_timestamp_values_from - final_query: {final_query}")
@@ -178,15 +196,13 @@ class ETL:
         result = PersistentDataStore(self.ppss.lake_ss.lake_dir).query_data(final_query)
 
         print(f"_get_max_timestamp_values_from - result: {result}")
-        values: Any = {}
 
         if result is None:
-            for table_name in table_names:
-                values[table_name] = None
-
-            return values
+            return none_values
 
         # print(f"_get_max_timestamp_values_from - result: {result}")
+
+        values = {}
 
         for row in result.rows(named=True):
             table_name = row["table_name"]
@@ -195,7 +211,7 @@ class ETL:
 
         return values
 
-    def _calc_bronze_start_end_ts(self):
+    def _calc_bronze_start_end_ts(self, tabletype: TableType = TableType.TEMP):
         """
         @description
             Calculate the start and end timestamps for the bronze tables
@@ -218,10 +234,14 @@ class ETL:
             self.raw_table_names, TableType.TEMP
         ).values()
 
+        fin_timestamp = self.ppss.lake_ss.fin_timestr if self.ppss.lake_ss.fin_timestr != "now" else datetime.now().strftime(
+                "%Y-%m-%d_%H:%M"
+            )
+
         to_timestamp = (
             min(to_values)
             if None not in to_values
-            else datetime.strptime(self.ppss.lake_ss.fin_timestr, "%Y-%m-%d_%H:%M")
+            else datetime.strptime(fin_timestamp, "%Y-%m-%d_%H:%M")
         )
 
         return from_timestamp, to_timestamp
--- a/pdr_backend/lake/etl.py
+++ b/pdr_backend/lake/etl.py
@@ -200,6 +200,22 @@ class ETL:
 
         return values
 
+    def get_timestamp_values(self, table_names: str, default_timestr: str) -> UnixTimeMs:
+        max_timestamp_values = self._get_max_timestamp_values_from(
+            [NamedTable(tb, TableType.NORMAL) for tb in table_names]
+        )
+        values = []
+        if len(max_timestamp_values) > 0:
+            values = [
+                value[1] for value in max_timestamp_values if value is not None
+            ]
+        timestamp = (
+            min(values)
+            if len(values) > 0
+            else UnixTimeMs.from_timestr(default_timestr)
+        )
+        return timestamp
+
     def _calc_bronze_start_end_ts(self) -> Tuple[UnixTimeMs, UnixTimeMs]:
         """
         @description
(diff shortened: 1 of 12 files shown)


pdr_backend/lake/etl.py
```diff
@@ -164,38 +164,54 @@ def _get_max_timestamp_values_from(
...
+
         for row in result.rows(named=True):
             table_name = row["table_name"]
             max_timestamp = row["max_timestamp"]
             values[table_name] = max_timestamp
 
         return values
 
-    def _calc_bronze_start_end_ts(self):
+    def _calc_bronze_start_end_ts(self, tabletype: TableType = TableType.TEMP):
```
idiom-bytes, line 214: you added this parameter but it doesn't get used.

pdr_backend/lake/etl.py
```diff
@@ -218,10 +234,16 @@ def _calc_bronze_start_end_ts(self):
             self.raw_table_names, TableType.TEMP
         ).values()
 
+        fin_timestamp = (
+            self.ppss.lake_ss.fin_timestr
+            if self.ppss.lake_ss.fin_timestr != "now"
+            else datetime.now().strftime("%Y-%m-%d_%H:%M")
```
idiom-bytes, line 240: i don't think this is right, we need to query the raw_tables to figure out how far the ETL can run

pdr_backend/lake/gql_data_factory.py
```diff
@@ -156,6 +156,10 @@ def _do_subgraph_fetch(
 
         buffer_df = pl.DataFrame([], schema=table.df_schema)
 
+        PersistentDataStore(self.ppss.lake_ss.lake_dir).create_table_if_not_exists(
+            get_table_name(table.table_name, TableType.TEMP), table.df_schema
+        )
```
idiom-bytes, line 161: This isn't needed.
- fetch will return and table.append_to_storage() will be called
- this in returns call _append_to_db()
- this calls `PersistentDataStore(self.base_path).insert_to_table(data, table_name)`
- insert_to_table check if the table doesn't exist, and creates it if needed

pdr_backend/lake/persistent_data_store.py
```diff
@@ -27,6 +28,28 @@ def __init__(self, base_path: str):
             database=f"{self.base_path}/duckdb.db"
         )  # Keep a persistent connection
 
+    @enforce_types
+    def create_table_if_not_exists(self, table_name: str, schema: SchemaDict):
```
idiom-bytes, line 32: this looks like a good addition

pdr_backend/lake/etl.py
```diff
@@ -164,38 +164,54 @@ def _get_max_timestamp_values_from(
             "SELECT '{}' as table_name, MAX(timestamp) as max_timestamp FROM {}"
         )
 
-        queries = [
-            max_timestamp_query.format(
-                get_table_name(table_name, table_type),
-                get_table_name(table_name, table_type),
+        all_db_tables = PersistentDataStore(
```
idiom-bytes, line 167: please review my PR before working through this
i believe it addresses a lot of this stuff

idiom-bytes:
Please review + merge my PR first, before considering submitting this code
I believe i addressed a lot of issues and more in my update

pdr_backend/lake/gql_data_factory.py
kdetry: I added this logic because if the data is empty, we do not create the table and it causes error on the bronze process

pdr_backend/lake/test/conftest.py
```diff
@@ -411,6 +411,20 @@ def fetch_function(
     return fetch_function
 
 
+@pytest.fixture()
+def _mock_fetch_empty_gql():
```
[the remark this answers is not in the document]
kdetry: I couldn't see but if it exists you can remove 👍

pdr_backend/lake/etl.py
kdetry: As I see the `_get_max_timestamp_values_from` method does not have any changes on your PR

pdr_backend/lake/etl.py
kdetry: OKay I got the issue now. I have changed the `_get_max_timestamp_values_from` method and it does not query the table if it does not exists. So I am going to remove it

pdr_backend/lake/gql_data_factory.py
idiom-bytes: thanks for sharing this, it makes sense
let's make sure we integrate this

pdr_backend/lake/test/conftest.py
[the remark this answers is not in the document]
idiom-bytes: let's keep it here and review when we merge this against my PR

idiom-bytes:
I'm updating tests at the moment.

There are many places where I'm seeing variables/labels as `timestamp`, but in reality it's not a timestamp, sometimes it's a `timestr`, and sometimes it's a `datetime` str.

I'm starting to clean this up a bit and beginning to enforce types, enforce returns, in addition to adding assert in key places such as checking that `st_ts < fin_ts`.

image

idiom-bytes:
i merged my larger pr + mustafa had done some updates...

i (1) merged my pr into his update now, sorted conflicts, iterated on the logic, and (2) cleaned up a bunch of code.

when it comes to (2) i noticed there were a bunch of [variables, function args, return values] called timestamp but weren't... they were datetime str, timestr, and other values.

i added a bunch of type-checks, asserts, and updated this everywhere so that if something is called a timestamp, I can expect it to be a UnixTimeS or UnixTimeMs

pdr_backend/lake/etl.py
```diff
@@ -153,7 +152,7 @@ def do_bronze_step(self):
 
     def _get_max_timestamp_values_from(
         self, table_names: List[str], table_type: TableType = TableType.NORMAL
```
calina-c, line 154: I don't really agree with having the same table type for a list of table_names, it looks like it could cause troubles in the future and is not flexible for mixed queries.

calina-c: my best suggestion would be a Table (or NamedTable...?) class which would hold the table name and type. It would support direct access to the attributes and make for better separation of concerns.

i.e. tables[List[NamedTable]] instead of table_names, get rid of table_type param altogether
then "if table.name not in all_db_tables"...

The NamedTable class would of course hold functionality to get_table_with_name and the initialisation would happen outside this function.

Such a change would greatly improve other parts of the code as well, where such concerns are mixed and prone to bugs.

pdr_backend/lake/etl.py
```diff
@@ -208,21 +212,35 @@ def _calc_bronze_start_end_ts(self):
...
-            if None not in from_values
-            else datetime.strptime(self.ppss.lake_ss.st_timestr, "%Y-%m-%d_%H:%M")
+            if len(from_values) > 0
+            else UnixTimeMs.from_timestr(self.ppss.lake_ss.st_timestr)
         )
 
-        to_values = self._get_max_timestamp_values_from(self.raw_table_names).values()
+        max_timestamp_values = self._get_max_timestamp_values_from(self.raw_table_names)
+        to_values = []
+        if len(max_timestamp_values) > 0:
```
calina-c, line 231: DRY issue, but seems minor.

calina-c (dismissed):

idiom-bytes (approved):
iterated on feedback and created new tickets

PR #895, opened 2024-04-15
Fix #894 - Memory leak issue

### LINKED ISSUE #894: [Bug][Lake] Memory issue while creating bronze tables with big datasets
We get a memory issue while creating bronze tables
image

image

kdetry:
Fixes #894 

Changes proposed in this PR:

- the split part logic on queries moved to raw tables

### DIFF
--- a/pdr_backend/lake/table_bronze_pdr_predictions.py
+++ b/pdr_backend/lake/table_bronze_pdr_predictions.py
@@ -46,8 +46,7 @@ def get_bronze_pdr_predictions_data_with_SQL(
         f"""
         SELECT
             {pdr_predictions_table_name}.ID as ID,
-            SPLIT_PART({pdr_predictions_table_name}.ID, '-', 1)
-                || '-' || SPLIT_PART({pdr_predictions_table_name}.ID, '-', 2) AS slot_id,
+            {pdr_predictions_table_name}.pdr_slot_slot_id as slot_id,
             {pdr_predictions_table_name}.contract as contract,
             {pdr_predictions_table_name}.slot as slot,
             {pdr_predictions_table_name}.user as user,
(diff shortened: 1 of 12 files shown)


idiom-bytes (approved):
This looks ok, does the new memory limits work on your setup @KatunaNorbert

pdr_backend/lake/table_bronze_pdr_predictions.py
```diff
@@ -46,8 +46,7 @@ def get_bronze_pdr_predictions_data_with_SQL(
         f"""
         SELECT
             {pdr_predictions_table_name}.ID as ID,
-            SPLIT_PART({pdr_predictions_table_name}.ID, '-', 1)
-                || '-' || SPLIT_PART({pdr_predictions_table_name}.ID, '-', 2) AS slot_id,
+            {pdr_predictions_table_name}.pdr_slot_slot_id as slot_id,
```
KatunaNorbert, line 49: This line is trowing me an error. We try to access "pdr_slot_slot_id" from `pdr_predictions` but that field it's not set

pdr_backend/subgraph/subgraph_payout.py
```diff
@@ -163,6 +163,7 @@ def fetch_payouts(
                 ),
                 "roundSumStakes": float(payout["prediction"]["slot"]["roundSumStakes"]),
                 "stake": float(payout["prediction"]["stake"]),
+                "pdr_slot_slot_id": f"{payout['id'].split('-')[1]}-{payout['id'].split('-')[2]}",
```
KatunaNorbert, line 166: The sot id format is `contract-slot` so the actual value is `"pdr_slot_slot_id": f"{payout['id'].split('-')[0]}-{payout['id'].split('-')[1]}"`

KatunaNorbert:
Did some fixes and tested it. The issue still happens and it's not solved.
The reason it worked was because the slot_id values stored were wrong and when computing the bronze tables the computation load was load because values were not found.
After fixing the slot id value the issues still occurs, feel free to test it

Issue #901, opened 2024-04-16
[Warning] datetime.utcfromtimestamp() is deprecated, scheduled for removal


trentmc:
### To reproduce

Ensure your system is up-to-date. Incl python 3.12.

Then:
```bash
pytest pdr_backend/util/test_noganache/test_time_types.py 
```

It outputs:
```text
...
====================================================================== warnings summary =======================================================================
pdr_backend/util/test_noganache/test_time_types.py: 31 warnings
  /Users/trentmc/code/pdr-backend/pdr_backend/util/time_types.py:83: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
    dt = datetime.utcfromtimestamp(int(self) / 1000)
```

### TODOs

- [ ] Update based on the recommendation

PR #903, opened 2024-04-16
Fix #907: Opportunity: solve #901 deprecation warning, while working in older py versions

### LINKED ISSUE #907: Opportunity: solve #901 deprecation warning, while working in older py versions
### Background / motivation

[#901](https://github.com/oceanprotocol/pdr-backend/issues/901) is "[Warning] datetime.utcfromtimestamp() is deprecated, scheduled for removal". This was fixed in Apr 16 am.

Then it raised a UTC import issue for at least one user (@w1kke - Robin). 

Robin then created a PR to fix it: [PR#903](https://github.com/oceanprotocol/pdr-backend/pull/903)

(The user has an older version of py.)

### TODOs 

Test and merge Robin's PR. Ensure that it solves #901 deprecation warning. And ideally while work in older py versions

w1kke:
Fixes #907

UTC does not exist in the datetime library for older versions of python. Changed the use to timezone.utc like in the line after the use of UTC


trentmc:
@w1kke update to the most recent version of Python

Then it works

trentmc (changes requested):
Do not merge. See my comment. This works on the most recent version of Python

w1kke:
My servers are on Ubuntu 22.04 LTS and the current version for it is Python 3.10 - but I will go down the deadsnake ppa path to update to Python 3.12

I was wondering as in the next line of the code UTC is used via timezone.utc - why not use the same way to access it in both lines?

https://github.com/oceanprotocol/pdr-backend/blob/96c0da83ab9df98b3f2262f17cf7e2c477dc21b6/pdr_backend/util/time_types.py#L83

Here I changed it to the same way as in the next line - this does not require Python3.12 but also works with Python3.10:
https://github.com/oceanprotocol/pdr-backend/blob/96c0da83ab9df98b3f2262f17cf7e2c477dc21b6/pdr_backend/util/time_types.py#L84

trentmc:
> I was wondering as in the next line of the code UTC is used via timezone.utc - why not use the same way to access it in both lines?

Good idea. OK then. If the tests pass, I'll approve and we can merge it.

trentmc:
> Good idea. OK then. If the tests pass, I'll approve and we can merge it.

Everything passed, including the main `pytest` tests. Exception: code coverage failed, because a missing key. I'm guessing that's because of an external contributor.

PR #914, opened 2024-04-17
Fix #908 - ETL unnecessary rows issue

### LINKED ISSUE #908: [Bug] Bronze prediction ETL is generating more rows than it should
**Describe the bug**
There is a bad join somewhere.

ETL is generating more rows than it should. There are only 10k predictions submitted, but 90k bronze_prediction rows.

**To Reproduce**
1. Edit ppss.yaml to simply use Nov-15-2023 => Nov-18-2023
2. Run `pdr analytics` to fetch all the data
3. Notice cli command generates an error in the output 
4. Run latest lake describe command from Calina's PR
5. Notice that there are only 10k predictions
Screenshot from 2024-04-16 08-49-22
6. Notice how there are somehow 90k rows in bronze_predictions table.
Screenshot from 2024-04-16 08-50-23

**Expected behavior**
There should only be 10k bronze predictions.

**Config:**
Edit ppss.yaml to simply use Nov-15-2023 => Nov-18-2023

**Additional context**
- [ ] There should only be 10k bronze_predictions after everything is said and done

kdetry:
Fixes #908 

Changes proposed in this PR:

- SQL query is fixed

### DIFF
--- a/pdr_backend/lake/table_bronze_pdr_predictions.py
+++ b/pdr_backend/lake/table_bronze_pdr_predictions.py
@@ -66,7 +66,8 @@ def get_bronze_pdr_predictions_data_with_SQL(
             {pdr_predictions_table_name}
         LEFT JOIN {pdr_truevals_table_name}
             ON {pdr_predictions_table_name}.slot = {pdr_truevals_table_name}.slot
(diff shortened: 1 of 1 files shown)


idiom-bytes (approved):
Fantastic.

PR #919, opened 2024-04-19
Issue 804 - Thread multisim

### LINKED ISSUE #804: [Sim, make $] Make benchmarking parallel, via threading
### Background / motivation

Previous issue: [#779](https://github.com/oceanprotocol/pdr-backend/issues/779) "[Sim, make $] Easier benchmarking, via new 'multisim' CLI command". [PR](https://github.com/oceanprotocol/pdr-backend/pull/783). [Commit](https://github.com/oceanprotocol/pdr-backend/commit/e9795f529140e838dab68b96eaf05924f70f66f0).

We can make benchmarking run faster if we make run across >1 processes or threads.

Don't bother with k8s, we don't need that level of speedup. Just multi-threading or multi-process is fine.

Priority (as of 20240318): not urgent, we're not bottlenecked by this yet

### How / TODOs

- [ ] Change `MultisimEngine.run()` to be parallel. Details below

### Details: what to change

Below is the code from [`multisim_engine.py::MultisimEngine.run()`](https://github.com/oceanprotocol/pdr-backend/blob/main/pdr_backend/sim/multisim_engine.py). Each iteration of the `for` loop runs a different benchmark. To make it parallel, make each iteration be on a different process or thread.
```python
    def run(self):
        ss = self.ss
        logger.info("Multisim engine: start. # runs = %s", ss.n_runs)
        self.initialize_csv_with_header()
        for run_i in range(ss.n_runs):
            point_i = self.ss.point_i(run_i)
            logger.info("Multisim run_i=%s: start. Vals=%s", run_i, point_i)
            ppss = self.ppss_from_point(point_i)
            sim_engine = SimEngine(ppss)
            sim_engine.run()
            run_metrics = sim_engine.st.recent_metrics()
            self.update_csv(run_i, run_metrics, point_i)
            logger.info("Multisim run_i=%s: done", run_i)
```

calina-c:
Closes #804


trentmc (changes requested):
The current changes look fine. 

However they don't do anything to address potential contention for working with the csv file. They should, no?

calina-c:
> The current changes look fine.
> 
> However they don't do anything to address potential contention for working with the csv file. They should, no?

There's a lock placed on the csv file to prevent simultaneous access. I can also make different csv files based on the multisim unique run id, but I'm not sure if we want to pursue that direction.

trentmc:
> > The current changes look fine.
> > However they don't do anything to address potential contention for working with the csv file. They should, no?
> 
> There's a lock placed on the csv file to prevent simultaneous access. I can also make different csv files based on the multisim unique run id, but I'm not sure if we want to pursue that direction.

What if:
- it tries to update csv
- if no lock, then good, all done 
- if a lock, then wait 1 s and try again

?

trentmc:
What's the status of this PR?

calina-c:
> What's the status of this PR?

Not sure how I can test the lock system and whether asyncio runs any waits. I'll come back to this task today.

calina-c:
@trentmc I double checked and that seems to be the behaviour of the statement:
https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock

waiting is therefore implicit.

trentmc:
> @trentmc I double checked and that seems to be the behaviour of the statement: https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock
> 
> waiting is therefore implicit.

Great! Thanks for checking.

trentmc (approved):

PR #948, opened 2024-04-29
Fix subgraph data fetching stop in loop

### LINKED ISSUE #904: [ETL][Lake] Fetching data stuck in loop
I run into this issue mainly on the Predictions and Payouts data fetching, and when it happens, after fetching the latest data it keeps on fetching from the beginning again and creates a loop. It is usually fixed when I stop the process and run it again. I suspect that something is off with handling the last date offset.

KatunaNorbert:
Fixes #904.


kdetry (approved):
LGTM

PR #949, opened 2024-04-29
Fix #941 - Thread test coverage issue

### LINKED ISSUE #941: [Lake][Tests] Thread test coverage is failing and throwing errors, although it technically passes CI/CD.
**Describe the bug**

The threading test inside of test_persistent_data_store is throwing a warning.
Tests are "passing" but technically should not as there is a type-error being handled as a warning.

**To Reproduce**
Try running pytest against test, notice it will pass although a type error is encountered. The test expects len(results) == 3, but results is None.
`pytest pdr_backend/lake/test/test_persistent_data_store.py::test_multiple_thread_table_updates -s`

**Expected behavior**
Test should pass without typerror, and with len(results) == 3.

**Screenshots**
image

### DoD:
- [ ] Fix `test_persistent_data_store.py::test_multiple_thread_table_updates`

kdetry:
Fixes #941 

Changes proposed in this PR:

- We passed a list to the other thread (it works with reference) and we moved the check step to the main thread

### DIFF
--- a/pdr_backend/lake/test/test_persistent_data_store.py
+++ b/pdr_backend/lake/test/test_persistent_data_store.py
@@ -2,6 +2,7 @@ import os
 import threading
 import polars as pl
 import duckdb
+import time
 
 from pdr_backend.lake.table import TableType, get_table_name, TempTable
 from pdr_backend.lake.persistent_data_store import PersistentDataStore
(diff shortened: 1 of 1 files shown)


idiom-bytes (approved):
ran tests and played with it to make sure it looks good

PR #972, opened 2024-04-30
#971 - Organize payout slots and contract addresses based on shared slots

### LINKED ISSUE #971: Payout Bug: claiming payouts for invalid slots
**Describe the bug**
The current payout function generates a list of slots and contract addresses but fails to verify if a prediction slot exists for each contract address. It incorrectly assumes that a slot has been submitted for every listed contract address.

**To Reproduce**
1. Submit a prediction for one contract.
2. Update the list of contract addresses.
3. The system attempts to claim payouts for the new contract, resulting in a failure.

**Expected behavior**
The payout function should ensure that a prediction slot exists for a contract before attempting to claim payouts and organize the slots & contract addresses to ensure validity.

trizin:
Fixes #971

Changes proposed in this PR:

- Organize payout slots and contract addresses based on shared slots
- Added `find_shared_slots` function and tests.

### DIFF
--- a/pdr_backend/cli/cli_arguments.py
+++ b/pdr_backend/cli/cli_arguments.py
@@ -1,7 +1,8 @@
 import argparse
+from argparse import Namespace
 import logging
 import sys
-from argparse import Namespace
+from typing import List
 
 from enforce_typing import enforce_types
 from eth_utils import to_checksum_address
@@ -52,7 +53,7 @@ Utilities:
   pdr get_traction_info ST END PQDIR PPSS_FILE NETWORK --FEEDS
   pdr check_network PPSS_FILE NETWORK --LOOKBACK_HOURS
   pdr create_accounts NUM PPSS_FILE NETWORK
-  pdr view_accounts ACCOUNTS PPSS_FILE NETWORK
+  pdr print_balances ACCOUNT PPSS_FILE NETWORK
   pdr fund_accounts TOKEN_AMOUNT ACCOUNTS PPSS_FILE NETWORK --NATIVE_TOKEN
   pdr deploy_pred_submitter_mgr PPSS_FILE NETWORK
 
@@ -120,23 +121,36 @@ class NETWORK_Mixin:
 
 
 @enforce_types
-def check_addresses(value):
+def check_addresses(value) -> List[str]:
     """
     @description
-        validates that all addressses is a comma-separated list of strings
-        each string, will be a valid Ethereum address
+        Validates that all addresses is a comma-separated list of strings
+        Each string must be a valid Ethereum address
+
+    @return
+        checksummed list of addresses
     """
     if not value:
         return []
+    addrs = value.split(",")
+    return [check_address(addr) for addr in addrs]
 
-    addresses = value.split(",")
-    checksummed_addresses = []
-    for address in addresses:
-        try:
-            checksummed_addresses.append(to_checksum_address(address.lower()))
-        except Exception as exc:
-            raise TypeError(f"{address} is not a valid Ethereum address") from exc
-    return checksummed_addresses
+
+@enforce_types
+def check_address(addr) -> str:
+    """
+    @description
+        Validates the input address a string, with a valid eth address
+
+    @return
+        checksummed version of the address
+    """
+    try:
+        addr2 = to_checksum_address(addr.lower())
+    except Exception as exc:
+        raise TypeError(f"{addr} is not a valid Ethereum address") from exc
+
+    return addr2
 
 
 @enforce_types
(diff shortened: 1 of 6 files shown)


pdr_backend/predictoor/predictoor_agent.py
```diff
@@ -481,22 +482,23 @@ def get_payout(self):
...
-        contracts_checksummed = [
-            self.ppss.web3_pp.web3_config.w3.to_checksum_address(addr)
-            for addr in contracts
-        ]
-        slots = list(pending_slots.values())
-        slots_flat = [item for sublist in slots for item in sublist]
-        slots_unique_flat = list(set(slots_flat))
-        if len(slots_unique_flat) == 0:
+        shared_slots = find_shared_slots(pending_slots)
+        if len(shared_slots) == 0:
```
trentmc, line 486: Replace

```python
if len(shared_slots) == 0:
```

with

```python
if not shared_slots:
```

Why: cleaner! :)

pdr_backend/predictoor/predictoor_agent.py
```diff
@@ -481,22 +482,23 @@ def get_payout(self):
...
             logger.info("No payouts available")
             return
 
-        tx = self.pred_submitter_mgr.get_payout(
-            slots_unique_flat, contracts_checksummed
-        )
-        logger.info("Payout tx: %s", tx["transactionHash"].hex())
+        for slot_tuple in shared_slots:
+            contract_addresses = slot_tuple[0]
```
trentmc, line 491: Replace 

```python
 contract_addresses = slot_tuple[0]
```

and (a few lines below)
```python
slots = slot_tuple[1]
```

with

```
contract_addresses, slots = slot_tuple
```

----

Also: replace `contract_addresses` -> `contract_addrs`

Why: always replace `address` -> `addr`, and `addresses` -> `addrs`, because it's just as descriptive / understandable, and more compact and readable.

pdr_backend/predictoor/predictoor_agent.py
```diff
@@ -481,22 +482,23 @@ def get_payout(self):
...
+            contracts_checksummed = [
```
trentmc, line 492: Rename `contracts_checksummed` -> `contract_addrs`, in all locations in this method. 

Why: `contracts_checksummed` is not contracts per se, but contract addresses. The variable name should reflect this. And going through a checksum filter doesn't make it worth the extra complexity of the variable name to include `checksummed`.

----
And in fact the whole 4 lines starting with `contracts_checksummed` can be converted to a single line using self's `_to_checksum()` method.

That is, from this:

```python
            contracts_checksummed = [
                self.ppss.web3_pp.web3_config.w3.to_checksum_address(addr)
                for addr in contract_addresses
            ]
```

To this:
```python
contract_addrs = self._to_checksum(contract_addrs)
```

pdr_backend/predictoor/predictoor_agent.py
```diff
@@ -481,22 +482,23 @@ def get_payout(self):
         pending_slots = query_pending_payouts(
             self.ppss.web3_pp.subgraph_url, up_pred_addr
         )
```
trentmc, line 484: Is `self.pred_submitter_mgr.predictoor_up_address()` (a) address of a prediction manager contract itself, (b) address of a particular prediction feed?
- The documentation of PredSubmitterMgr::predictoor_up_address() isn't helpful. It says "@return: address -- str, address of upward predictoor contract".
- In general calling something a "predictoor x address" is unhelpful, because so many things are "predictoor". And the main thing that's a predictoor is the person running the bots.

Please change the name of method `predictoor_up_address()`, and `predictoor_up_address()` to make it more precise.

Then, please  `up_pred_addr` to correspond.

Second: there was no call to get / use `predictoor_down_address()` in this method. Was that intentional, or an inadvertent omission? If the latter, it could be a bug?

trentmc (changes requested):
Changes requested, mostly to reduce code complexity. Please ping me when ready for review again.

pdr_backend/predictoor/predictoor_agent.py
trentmc: > We might rename it to `pred_submitter_up_address`, does that sound good?

OK. 

And, clarify the docstring for `pred_submitter_up_address()`.

And, same for the "down" version of course (name, and docstring).

> `up_pred_addr`

Just change this to `up_addr`. This is the lowest-complexity option, and we can get away with it since the variable is local and the method is short.

>> [T] there was no call to get / use predictoor_down_address() in this method. Was that intentional, or an inadvertent omission? If the latter, it could be a bug?
> [B] Since both predict on the same slots it's sufficient to query for just one of them

OK. Let's clarify in the code. My recommendation: above the line that would now read `up_addr = self.pred_submitter_mgr.pred_submitter_up_address()`, write a comment such as `# we only need to query in one direction, since both predict on the same slots`.

trentmc (approved):

PR #976, opened 2024-04-30
Fix #975 - Bronze PDR Slots Query Bug

### LINKED ISSUE #975: [Lake][Bug] The bronze PDR slots table has more than it needs to include
The bronze PDR slots table has more than it needs to include. Reported by @KatunaNorbert on slack

The timestamp value is also weird, some of them are just "1"

**DoD**

- [ ] Inspect and fix the SQL
- [ ] Test manually sure about the data result

kdetry:
Fixes #975 

Changes proposed in this PR:

- The SQL query has been fixed

### DIFF
--- a/pdr_backend/lake/table_bronze_pdr_slots.py
+++ b/pdr_backend/lake/table_bronze_pdr_slots.py
@@ -54,46 +54,50 @@ def get_bronze_pdr_slots_data_with_SQL(
 
     query = f"""
             INSERT INTO {temp_bronze_pdr_slots_table_name}
-            SELECT
-                {pdr_slots_table_name}.ID as ID,
-                SPLIT_PART({pdr_slots_table_name}.ID, '-', 1) as contract,
-                filtered_{etl_bronze_pdr_predictions_table_name}.pair as pair,
-                filtered_{etl_bronze_pdr_predictions_table_name}.timeframe as timeframe,
-                filtered_{etl_bronze_pdr_predictions_table_name}.source as source,
-                {pdr_slots_table_name}.slot as slot,
-                filtered_{pdr_payouts_table_name}.roundSumStakesUp as roundSumStakesUp,
-                filtered_{pdr_payouts_table_name}.roundSumStakes as roundSumStakes,
-                {pdr_truevals_table_name}.truevalue as truevalue,
-                {pdr_slots_table_name}.timestamp as timestamp,
+            SELECT DISTINCT
+                {pdr_slots_table_name}.ID AS ID,
+                SPLIT_PART({pdr_slots_table_name}.ID, '-', 1) AS contract,
+                {etl_bronze_pdr_predictions_table_name}.pair AS pair,
+                {etl_bronze_pdr_predictions_table_name}.timeframe AS timeframe,
+                {etl_bronze_pdr_predictions_table_name}.source AS source,
+                {pdr_slots_table_name}.slot AS slot,
+                {pdr_payouts_table_name}.roundSumStakesUp AS roundSumStakesUp,
+                {pdr_payouts_table_name}.roundSumStakes AS roundSumStakes,
+                {pdr_truevals_table_name}.truevalue AS truevalue,
+                {pdr_slots_table_name}.timestamp AS timestamp,
                 GREATEST(
                     {pdr_slots_table_name}.timestamp,
-                    filtered_{pdr_payouts_table_name}.timestamp,
-                    filtered_{etl_bronze_pdr_predictions_table_name}.timestamp,
-                    {pdr_truevals_table_name}.timestamp
-                ) as last_event_timestamp,
+                    {pdr_truevals_table_name}.timestamp,
+                    {pdr_payouts_table_name}.timestamp,
+                    {etl_bronze_pdr_predictions_table_name}.timestamp
+                ) AS last_event_timestamp
             FROM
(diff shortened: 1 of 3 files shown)


KatunaNorbert:
Looks like the new query it's not working, test is failing, also the flow fails on local

pdr_backend/lake/table_bronze_pdr_slots.py
```diff
@@ -54,46 +54,50 @@ def get_bronze_pdr_slots_data_with_SQL(
...
-            ON {pdr_slots_table_name}.ID = filtered_{pdr_payouts_table_name}.match_id
-            AND filtered_{pdr_payouts_table_name}.rn = 1
+                SELECT 
+                    {pdr_slots_table_name}.ID AS slot_id,
+                    MAX({etl_bronze_pdr_predictions_table_name}.pair) AS pair,
+                    MAX({etl_bronze_pdr_predictions_table_name}.timeframe) AS timeframe,
+                    MAX({etl_bronze_pdr_predictions_table_name}.source) AS source,
+                    MAX({etl_bronze_pdr_predictions_table_name}.timestamp) AS timestamp
+                FROM {pdr_slots_table_name}
+                LEFT JOIN {etl_bronze_pdr_predictions_table_name} ON SPLIT_PART({pdr_slots_table_name}.ID, '-', 1) = {etl_bronze_pdr_predictions_table_name}.contract
```
kdetry, line 85: why do we use "join" again?

pdr_backend/lake/table_bronze_pdr_slots.py
```diff
@@ -54,46 +54,50 @@ def get_bronze_pdr_slots_data_with_SQL(
...
+                GROUP BY {pdr_slots_table_name}.ID
+            ) AS {etl_bronze_pdr_predictions_table_name} ON {pdr_slots_table_name}.ID = {etl_bronze_pdr_predictions_table_name}.slot_id
```
kdetry, line 87: we should not use a JOIN result instead of an existing table's name, here `{etl_bronze_pdr_predictions_table_name}`. I think we have to use an alias.

pdr_backend/lake/table_bronze_pdr_slots.py
```diff
@@ -54,46 +54,50 @@ def get_bronze_pdr_slots_data_with_SQL(
...
-                ROW_NUMBER() OVER (PARTITION BY match_id) as rn,
-                FROM {pdr_payouts_table_name} 
-            ) AS filtered_{pdr_payouts_table_name} 
-            ON {pdr_slots_table_name}.ID = filtered_{pdr_payouts_table_name}.match_id
-            AND filtered_{pdr_payouts_table_name}.rn = 1
+                SELECT 
+                    {pdr_slots_table_name}.ID AS slot_id,
+                    MAX({etl_bronze_pdr_predictions_table_name}.pair) AS pair,
+                    MAX({etl_bronze_pdr_predictions_table_name}.timeframe) AS timeframe,
+                    MAX({etl_bronze_pdr_predictions_table_name}.source) AS source,
```
kdetry, line 82: the pair, timeframe and source columns are string values, I think using `MAX` does not make sense

kdetry (commented):
As I see you changed the "ROW_NUMBER" logic with group_by, it makes sense but I have some concerns and shared them

pdr_backend/lake/table_bronze_pdr_slots.py
KatunaNorbert: I want to get all the data based on contract address, before it used slot_id and if there was a mismatch the contract related values were null. We should be able to only get the data based on contractId. We might be able to further improve this

pdr_backend/lake/table_bronze_pdr_slots.py
KatunaNorbert: MAX() will return the string value that appears last in the collation order.

pdr_backend/lake/table_bronze_pdr_slots.py
```diff
@@ -54,35 +54,50 @@ def get_bronze_pdr_slots_data_with_SQL(
...
+                FROM {pdr_payouts_table_name}
+                GROUP BY slot_id
+            ) AS joined_{pdr_payouts_table_name} ON {pdr_slots_table_name}.ID = joined_{pdr_payouts_table_name}.slot_id
             WHERE
-                {pdr_slots_table_name}.timestamp >= {st_ms}
-                AND {pdr_slots_table_name}.timestamp <= {fin_ms}
+                pdr_slots.timestamp >= {st_ms}
+                AND pdr_slots.timestamp <= {fin_ms}
+            ORDER BY timestamp
         """
```
idiom-bytes, line 101: 1. please dump the data at the end o the DF
2. please copy this data into the test to make it easy to see
3. use the data from test to be processed by this function, such that we can:
a) see the output from this query
b) see the data being processed correctly

pdr_backend/lake/table_bronze_pdr_slots.py
idiom-bytes: As I wrote here... these should be retrieved from subgraph so the data is just there from the initial fetch from the subgraph... No need to execute a join or a max.
https://github.com/oceanprotocol/pdr-backend/issues/989

And... if you just need one record in SQL use `first()` as the scan/cost of the call will be cheaper or `last()`

PR #993, opened 2024-05-05
Fix #982: active checkpoint system

### LINKED ISSUE #982: [Lake][Incremental Pipeline] Use bronze_pdr_predictions to manage active checkpoint
### Background / motivation

Inferring where to resume the pipeline from is having issues.

It's not about "removing subscriptions" from this logic, we need to think about the system and what it's trying t do.

Rather than querying all tables to find the timestamp from where to resume the pipeline, just use the following for now:
- raw_predictions for GQLDF
- bronze_pdr_predictions for ETL

### Macro

let's break down what's happening here

**we're querying the data tables from duckdb, in order to understand how to resume the jobs pipeline incrementally**

basically, resuming from where it left off

i had mentioned in the past updating my_ppss.yaml such that st_ts is mutated w/ the last_run_timestamp and the pipeline enforces being incremental rather than trying to do it through data-inference (like ohlcv data factory)... but, this kind of breaks the pattern for how the yaml file is being used.

it also doesn't provide a way to track the etl/workflow runs, such that they can be rolledback in a systemic way

so, let's do this:
- [x] document this (so it's clear how this is expected to work) and how we plan to solve it properly
1. use raw_predictions/bronze_predictions max_timestamps to resume GQLDF/ETL.
1. ship duckdb integration
1. start tracking the runs in duckdb, and using these as the checkpoint for how to run the next GQLDF/ETL run

### TODOs / DoD

**Checkpoints & Incremental Pipeline**
============
Update GQLDF and ETL logic, to resume from their equivalent prediction tables.

**To run GQLDF**
- We're going to use the st_ts from the last(raw_pdr_predictions) in duck_db
- We're going to use the end_ts from ppss.lake_ss.end_ts = now

**To run ETL**
- We're going to use the st_ts from the last(raw_pdr_predictions) in duck_db
- We're going to use the end_ts from the last(temp_raw_pdr_predictions) in duck_db (it could also be ppss.lake_ss.end_ts = now)

This will KISS and let us ship a simple incremental pipeline that works reliably and gives us the data quality we need.
We'll worry about implementing a job object to abstract this later.

This means your ppss can be:
st_ts: 01-01-2024
end_ts: now

...and it should just work

Tasks:
- [x] Update GQLDF to start/end from the correct spot
- [x] Update ETL to start/end from the correct spot

kdetry:
Fixes #982 

Changes proposed in this PR:

- the move query is changed
- UNIQUE constraints are added
- the start-end timestamp logic is changed for the ETL and GQLDF

### DIFF
--- a/pdr_backend/lake/gql_data_factory.py
+++ b/pdr_backend/lake/gql_data_factory.py
@@ -106,6 +106,9 @@ class GQLDataFactory:
             }
         )
 
+        self._etl_st_ms = None
+        self._etl_fin_ms = None
+
     @enforce_types
     def get_gql_tables(self) -> Dict[str, Table]:
         """
@@ -158,6 +161,37 @@ class GQLDataFactory:
             logger.info("  Table exists. Insert pending %s csv data", table_name)
             data = CSVDataStore(table.base_path).read(table_name, st_ut, fin_ut, schema)
             table._append_to_db(data, TableType.TEMP)
+
+    @enforce_types
+    def _calc_start_ut_from_predictions(self) -> UnixTimeMs:
+        """
+        @description
+            Calculate start timestamp, reconciling whether file exists and where
+            its data starts. If file exists, you can only append to end.
+
+        @return
+            start_ut - timestamp (ut) to start grabbing data for (in ms)
+        """
+
+        # check if the predictions table exists in the database
+        if not PersistentDataStore(self.ppss.lake_ss.lake_dir).table_exists(
+            predictions_table_name
+        ):
+            return UnixTimeMs(self.ppss.lake_ss.st_timestamp + 1000)
+
+        # get last timestamp of the predictions table from the database
+
+        last_timestamp = PersistentDataStore(self.ppss.lake_ss.lake_dir).query_data(
+            f"SELECT MAX(timestamp) FROM {predictions_table_name}"
+        )
+
+        start_ut = (
+            last_timestamp['max("timestamp")'][0]
+            if last_timestamp is not None
+            else self.ppss.lake_ss.st_timestamp
+        )
+
+        return UnixTimeMs(start_ut + 1000)
 
     @enforce_types
     def _calc_start_ut(self, table: Table) -> UnixTimeMs:
@@ -275,6 +309,17 @@ class GQLDataFactory:
             Move the records from our ETL temporary build tables to live, in-production tables
         """
 
+        if not PersistentDataStore(self.ppss.lake_ss.lake_dir).table_exists(
+            predictions_table_name
+        ):
+            self._etl_st_ms = self.ppss.lake_ss.st_timestamp
+        else:
+            self._etl_st_ms = UnixTimeMs(
+                PersistentDataStore(self.ppss.lake_ss.lake_dir).query_data(
+                    f"SELECT MAX(timestamp) FROM {predictions_table_name}"
+                )['max("timestamp")'][0]
+            )
+
         pds = PersistentDataStore(self.ppss.lake_ss.lake_dir)
         for table_name in self.record_config["gql_tables"]:
             pds.move_table_data(
@@ -283,6 +328,15 @@ class GQLDataFactory:
             )
 
             pds.drop_table(get_table_name(table_name, TableType.TEMP))
+
+        self._etl_fin_ms = UnixTimeMs(
+            PersistentDataStore(self.ppss.lake_ss.lake_dir).query_data(
+                f"SELECT MAX(timestamp) FROM {predictions_table_name}"
+            )['max("timestamp")'][0]
+        )
+
+    def get_etl_st_fin(self):
+        return self._etl_st_ms, self._etl_fin_ms
 
     @enforce_types
     def _update(self):
@@ -301,12 +355,13 @@ class GQLDataFactory:
             fin_ut -- a timestamp, in ms, in UTC
         """
 
+        st_ut = self._calc_start_ut_from_predictions()
+        fin_ut = self.ppss.lake_ss.fin_timestamp
+
         for table in (
             TableRegistry().get_tables(self.record_config["gql_tables"]).values()
         ):
             # calculate start and end timestamps
-            st_ut = self._calc_start_ut(table)
-            fin_ut = self.ppss.lake_ss.fin_timestamp
             logger.info(
                 "      Aim to fetch data from start time: %s", st_ut.pretty_timestr()
             )
--- a/pdr_backend/lake/gql_data_factory.py
+++ b/pdr_backend/lake/gql_data_factory.py
@@ -231,9 +231,11 @@ class GQLDataFactory:
             # convert predictions to df and transform timestamp into ms
             df = _object_list_to_df(data, table.df_schema)
             df = _transform_timestamp_to_ms(df)
-            df = df.filter(pl.col("timestamp").is_between(st_ut, fin_ut)).sort(
-                "timestamp"
-            )
+            df = df.filter(pl.col("timestamp").is_between(st_ut, fin_ut))
+
+            if len(df) > 0:
+                if df["timestamp"][0] > df["timestamp"][len(df) - 1]:
+                    df = df.filter(pl.col("timestamp") >= df["timestamp"][0])
 
             if len(buffer_df) == 0:
                 buffer_df = df
@@ -254,6 +256,8 @@ class GQLDataFactory:
 
                 buffer_df = pl.DataFrame([], schema=table.df_schema)
                 save_backoff_count = 0
+                if df["timestamp"][0] > df["timestamp"][len(df) - 1]:
+                    return
 
             # avoids doing next fetch if we've reached the end
             if len(df) < pagination_limit:
--- a/pdr_backend/lake/gql_data_factory.py
+++ b/pdr_backend/lake/gql_data_factory.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Callable, Dict
+from typing import Callable, Dict, Optional
 from enforce_typing import enforce_types
 import polars as pl
 from pdr_backend.lake.table import Table, TableType, get_table_name, TempTable
@@ -133,7 +133,7 @@ class GQLDataFactory:
 
         return TableRegistry().get_tables(self.record_config["gql_tables"])
 
-    def _prepare_temp_table(self, table_name, st_ut, fin_ut, schema):
+    def _prepare_temp_table(self, table_name, st_ut, fin_ut, schema) -> Optional[UnixTimeMs]:
         """
         @description
             # 1. get last timestamp from database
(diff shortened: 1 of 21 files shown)


pdr_backend/lake/gql_data_factory.py
```diff
@@ -159,6 +162,37 @@ def _prepare_temp_table(self, table_name, st_ut, fin_ut, schema):
...
+
+        @return
+            start_ut - timestamp (ut) to start grabbing data for (in ms)
+        """
+
+        # check if the predictions table exists in the database
+        if not PersistentDataStore(self.ppss.lake_ss.lake_dir).table_exists(
+            predictions_table_name
+        ):
+            return UnixTimeMs(self.ppss.lake_ss.st_timestamp + 1000)
```
KatunaNorbert, line 180: Why are we adding 1000 to the timestamp?

KatunaNorbert:
deleted lake_data and run ETL, stopped it at the subscriptions fetching step and run into the following error when run it again:
Screenshot 2024-05-07 at 15 08 59

KatunaNorbert:
I think I see what is happening, if the gql update is stopped the temp table and csv are updated but live table is not created. Then when restarting the process we take the start date from live table but that doesn't exist so we take it from lake_ss but there is data inside cvs and temp table and when trying to insert it gives an error

KatunaNorbert:
A fix could be to add a try catch around the `self._do_subgraph_fetch` so whenever there the process is stopped we catch the error and move the existing tables to production. This was the production tables are reflecting what's inside the temp tables and there are not conflicts.
I tested it and seems to be working, I could push the changes if you agree with this approach

idiom-bytes:
```
A fix could be to add a try catch around the self._do_subgraph_fetch so whenever there the process is stopped we catch the error and move the existing tables to production. This was the production tables are reflecting what's inside the temp tables and there are not conflicts.
```

The point is that there was an error and that updates need to be atomic.
They either complete successfully, or they don't.

What should happen here is...
1. GQL Data Factory should resume fetching from where it left off.
2. When all fetching is completed, all data (from st_ts => to end_ts) is loaded from CSV onto the DuckDB tables.
3. When all of this completes end-to-end the GQLDF is up-to-date.

pdr_backend/lake/persistent_data_store.py
```diff
@@ -53,6 +88,29 @@ def create_table_if_not_exists(self, table_name: str, schema: SchemaDict):
             empty_df = pl.DataFrame([], schema=schema)
             self._create_and_fill_table(empty_df, table_name)
 
+    def _create_sql_table_schema(self, schema: SchemaDict) -> str:
```
idiom-bytes, line 91: why is this having to be created now?

pdr_backend/lake/persistent_data_store.py
```diff
@@ -196,15 +258,45 @@ def move_table_data(self, temp_table, permanent_table_name: str):
...
+                f"SELECT * FROM {temp_table.fullname} LIMIT 1"
+            ).pl()
+
+            # create the sql schema
+            sql_table_schema = self._create_sql_table_schema(temp_table_df.schema)
+
+            # create the permanent table
+            self.execute_sql(
+                f"CREATE TABLE {permanent_table_name} ({sql_table_schema})"
+            )
```
idiom-bytes, line 272: can we put all of this into a function?

pdr_backend/lake/persistent_data_store.py
```diff
@@ -196,15 +258,45 @@ def move_table_data(self, temp_table, permanent_table_name: str):
...
+        temp_table_columns = temp_table_columns.columns
+
+        # generate on conflict clause
+        on_conflict_columns = ", ".join(
+            [
+                f"{column} = EXCLUDED.{column}"
+                for column in temp_table_columns
+                if column != "ID"
+            ]
+        )
```
idiom-bytes, line 294: I don't think this is correct...
1. I don't think the pds should handle the conflict. I.e. it should just handle merging from one to the other.
2. You are trying to handle stuff ahead of time.

What should happen right now....
1. Only new events are being processed... **there should be no conflicts.**
2. There will be gaps in the data "null points", do not try to join/fx for these at the moment.

pdr_backend/lake/persistent_data_store.py
```diff
@@ -196,15 +258,45 @@ def move_table_data(self, temp_table, permanent_table_name: str):
...
         # Move the data from the temporary table to the permanent table
         self.execute_sql(
-            f"INSERT INTO {permanent_table_name} SELECT * FROM {temp_table.fullname}"
+            f"""INSERT INTO {permanent_table_name}
+            SELECT * FROM {temp_table.fullname}
+            ON CONFLICT (ID) DO UPDATE SET {on_conflict_columns}"""
```
idiom-bytes, line 299: Same here...

pdr_backend/lake/persistent_data_store.py
```diff
@@ -31,6 +32,40 @@ def __init__(self, base_path: str, read_only: bool = False):
...
+            pl.Time: "TIME",  # TIME only for time data.
+            pl.Duration: "INTERVAL",  # INTERVAL for time durations.
+            pl.List: "ARRAY",  # ARRAY for list data types, ensure compatibility.
+            pl.Object: "BLOB",  # BLOB for binary large objects, check usage context.
+            pl.Struct: "STRUCT",  # STRUCT for nested data, might need specific handling in SQL.
+            # Additional types like dictionaries, etc.,
+            # can be added based on specific needs and compatibility.
+        }
+
+    def _get_sql_column_type(self, column_type) -> str:
```
idiom-bytes, line 59: why does this have to be created now, as part of this PR?

KatunaNorbert:
> We do not wan to update the live table if there was an error...
> 1. GQL Data Factory should resume fetching from where it left off.
> 2. When all fetching is completed, all data (from st_ts => to end_ts) is loaded from CSV onto the DuckDB tables.
> 3. When all of this completes end-to-end the GQLDF is up-to-date.

I agree with this, If we move to production and there is missing data then we can't trust the production

pdr_backend/lake/persistent_data_store.py
kdetry: This logic was created because we were using auto-casting for datatypes in the database. However, we now require at least one unique column, and making a column unique by altering it does not work. Therefore, we must specify it while creating the table. As a solution, I created this logic to cast column types manually.


I am open for different suggestions

pdr_backend/lake/persistent_data_store.py
kdetry: https://github.com/oceanprotocol/pdr-backend/pull/993#discussion_r1593838809

pdr_backend/lake/gql_data_factory.py
kdetry: It is unnecessary to add 1000 to the timestamp not retrieved from the DB. I agree

kdetry: removed

pdr_backend/lake/persistent_data_store.py
idiom-bytes: Can we just use the id column, which should be unique and every object has one at the moment?

idiom-bytes:
At the Data Fetching level..
Use CSV as stream + checkpoint -> Should keep writing and updating. This will work as the checkpoint for fetching new data.
DuckDB as output -> Will write at the very end, when CSV is done updating. To get the st_ts, fetch the max(timestamp) from `pdr_predictions` from duck_db.... This should give us the `last_update_from_csv_to_duckdb` and provide us with the starting point to get data from CSV -> DuckDB.

pdr_backend/lake/gql_data_factory.py
```diff
@@ -373,14 +373,18 @@ def _update(self):
...
+                    table,
+                    self.record_config["fetch_functions"][table.table_name],
+                    self.ppss.web3_pp.network,
+                    st_ut,
+                    fin_ut,
+                    self.record_config["config"],
+                )
+            except Exception as e:
+                self._move_from_temp_tables_to_live()
+                logger.error("Error on fetching data from %s: %s", table.table_name, e)
```
idiom-bytes: I think we need to review this...

pdr_backend/lake/test/test_gql_data_factory.py
```diff
@@ -198,7 +198,7 @@ def test_calc_start_ut_from_predictions(_gql_datafactory_etl_predictions_df, tmp
     gql_data_factory = GQLDataFactory(ppss)
 
     st_ut = gql_data_factory._calc_start_ut_from_predictions()
-    assert st_ut.to_seconds() == 1701561601
+    assert st_ut.to_seconds() == 1701561600
```
idiom-bytes, line 201: I think I understand this update, thank you.

pdr_backend/lake/gql_data_factory.py
KatunaNorbert: removed that part after adding it, realised we don't want this

KatunaNorbert:
Tested it and identified an issue:
When running gql update to fetch the data then stopping it along the way, for example after fetching predictions, slots and trueval, then running it again it fetches again all the data. I think it's looking at the live pdr_predictions table which it's only created at the end of the etl update, which is fine, but if the fetching process is stopped we should take the existing data from csvs and only fetch what is left instead of fetching everything.

KatunaNorbert:
The resume part looks fine now, but when stopping the fetch predictions process and restarting it at timestamp_x that timestamp is going to be used in 
Screenshot 2024-05-10 at 11 38 39
all the other tables instead of the start date from the lake

kdetry:
> The resume part looks fine now, but when stopping the fetch predictions process and restarting it at timestamp_x that timestamp is going to be used in all the other tables instead of the start date from the lake Screenshot 2024-05-10 at 11 38 39

I tried to explain this situation while discussing the logic, we shouldn't rely on the pdr_predictions table for GQLDF, all other tables should manage their own situations. Only the ETL side look at pdr_predictions, if the bronze_pdr_predictions table does not have a record.

idiom-bytes:
@kdetry @KatunaNorbert you tagged the wrong ticket...
https://github.com/oceanprotocol/pdr-backend/issues/982

#982 is or incremental updates... which is not what we're doing

```
the start-end timestamp logic is changed for the ETL and GQLDF
```
please read the readme, the epic, and ticket #1000 
incremental updates are for AFTER we get the first build ready.

The only thing that we care about right now is putting raw-records into GQLDF.
ETL is not part of #1000.

pdr_backend/lake/test/test_gql_data_factory.py
```diff
@@ -298,3 +331,129 @@ def test_do_subgraph_fetch_stop_loop_when_restarting_fetch(
...
+    )
+
+    gql_data_factory = GQLDataFactory(ppss)
+    gql_data_factory.record_config["gql_tables"] = ["pdr_predictions"]
+
+    pds = PersistentDataStore(ppss.lake_ss.lake_dir)
+    csvds = CSVDataStore(ppss.lake_ss.lake_dir)
+    # Add some predictions
+    pds.drop_table(get_table_name("pdr_predictions", TableType.TEMP))
+    csvds.delete("pdr_predictions")
```
idiom-bytes, line 393: i don't get this...
1. pytest creates a different tmpdir for every test
2. you don't need to delete anything, there is nothing there
3. there is now a function to delete csvs which is only used by this test

idiom-bytes:
We've discussed the issues in here, reviewed implementations and agreed that:
1. Much of the work being done originally and the order-of-events were indeed correct.
2. We need to look across all tables to get the right timestamp, we can't just use the predictions table
3. There were a lot of wrong assumptions in this implementation. Example: GQLDF SHOULD look at CSV records for where to resume not the DuckDB tables.

Closing this PR

PR #999, opened 2024-05-06
Fix #990: Add predictoor log line.

### LINKED ISSUE #990: Issue with Logging Small Stake Amounts Due to Formatting
When formatting stake amounts using ":.2f", values less than 0.01 are not displayed. This behavior mirrors the issue addressed in [#952](https://github.com/oceanprotocol/pdr-backend/issues/952). Implement a solution that ensures these small stake amounts are also logged appropriately.

Lines: https://github.com/oceanprotocol/pdr-backend/blob/main/pdr_backend/predictoor/predictoor_agent.py#L153-L154

calina-c:
Closes #990


pdr_backend/util/logger_base.py
```diff
@@ -0,0 +1,21 @@
...
+            else attr_or_amount
+        )
+
+        if use_eth:
+            x = x.amt_eth
+
+        if x < 0.01:
+            return f"{x:6.2e}"
+
+        return f"{x:6.2f}"
```
trentmc, line 21: We already have `util/strutil.py::prettyBigNum()`. It's a super useful routine. And doesn't have OO overkill.'

I strongly suggest that here too.

To be precise: I recommend to delete `logger_base.py` along with `_compactNum()` there. 

And instead add a new function in `util/strutil.py` as follows:
```python
def compactSmallNum(x: Union[float, int]) -> str:
    """
    @description
      Prints numbers with two decimal places precision, using e-notation for small numbers if needed. 
      Removes unneeded zeroes.
    """
    if x == 0:
        return "0"
    if x >= 0.01:
        return f"{x:6.2f}"
    s = f"{x:6.2e}"
    s = s.replace("e-0", "e-")
    return s
```

I put a few things in there intentionally:
- Handles special case of x = 0
- Removes the "0" right after the "e-"
- Doesn't take in `Eth`. (That's a higher level of abstraction. Shouldn't muck up this level)

Also, be sure to add a unit test, next to the unit test for `prettyBigNum()`.

trentmc (changes requested):

pdr_backend/util/logger_base.py
trentmc: > should I use the same logic in SimLoggerLine?

Yes. And I see that you have. Great!

Issue #1009, opened 2024-05-09
ARIMA-style feed analysis: for {BoxCox=T/F, diff=0,1,2}, plot the seasonality and ACF/PACF


trentmc:
## Background / motivation

Parent epic: [#1006](https://github.com/oceanprotocol/pdr-backend/issues/1006) "Explore SARIMAX modeling"

**Top-level goal: Help user answer which of which transform to use {0, B0, B1, B2}, to (a) achieve stationarity; and (b) see how well the signal can decompose into trend / seasonal / residual.** 

**Specific transforms:**
  - 0: differencing = 0
  - B0: Box-Cox transform + differencing=0
  - B1: Box-Cox transform + differencing=1
  - B2: Box-Cox transform + differencing=1

**Sub-goals: We want it easy to see these:**
- decomposition into trend / seasonal / residual
- relative strengths of trend vs seasonality vs resid
- ACF & PACF. That is, autocorrelation function & partial autocorrelation function, for insight into best AR coefficient etc
- ADF for each transform {0, B0, B1, B2}. Insight: the lowest # that has ADF < 0.05 is the best
- Then, make it easy for the user to select a transform and see the plots ^ for that given transform

**Best ref:** jahnvisikligar Master's thesis https://github.com/jahnvisikligar/Masters-Thesis/blob/main/Final_Thesis_code.ipynb

## TODOs / DoD

- [x] Add real crypto data, ready for use by unit tests [PR #1016](https://github.com/oceanprotocol/pdr-backend/pull/1016)
- [ ] Unit test with plots for each sub-component:
  - [x] decomposition into trend / seasonal / residual [PR #1016](https://github.com/oceanprotocol/pdr-backend/pull/1016)
  - [x] relative strengths of trend vs seasonality vs resid [PR #1025](https://github.com/oceanprotocol/pdr-backend/pull/1025)
  - [x] ACF & PACF [PR #1026](https://github.com/oceanprotocol/pdr-backend/pull/1026)
  - [x] ADF for each transform {0, B0, B1, B2}. (Calc'd it as part of above line. Plots are being done in another issue).

## Main prototype: with everything

For a given choice of differencing {0,B0,B1,B2}, show
- 5 decomposition plots: energy per signal, plus 4 signals: (i) observed feed (ii) "trend" (iii) "seasonal", "residual".
- autocorrelation (ACF) & partial autocorrelation (PACF) plots

[...]

PR #1026, opened 2024-05-11
Towards #1009: plot ACF & PACF

trentmc:
Towards #1009


trentmc:
### BoxCox = F, diff=0
bc=f diff=0

### BoxCox = T, diff=0
bc=t diff=0

### BoxCox = T, diff=1
bc=t diff=1

PR #1034, opened 2024-05-13
#1033 - slack bot warn when subgraph gets stuck

### LINKED ISSUE #1033: Slack bot: warn when subgraph gets stuck
Update slack bot to warn when subgraph gets stuck

trizin:
Fixes #1033

Changes proposed in this PR:

- Check subgraph last sync block and notify slack if it's behind.
- Replace trivia questions with random jokes.
- Update grep to also notify for errors from python.
- Update GH action to use main branch and failing fixed tests & type issues.


trentmc (approved):
Outside code that I know well. But at a glance, looks sane. And if there are issues then we will know soon enough, without much risk

Issue #973, opened 2024-04-30
[Lake][Data Quality] Add pydantic to validate schema at the lake-level


idiom-bytes:
### Background / motivation

The current table schemas use types but are poorly validated. We also can't leverage validation wrappers because it's just using a generic type.

Rather: let's implement a validation tool like pydantic to validate data through our system, and in our lake

### DoD
- [ ] Update table schema definitions to use pydantic
- [ ] Update `lake validate` cli to validate lake schema vs. expected schema

idiom-bytes:
We're not going to do the pydantic working any longer, since we updated our schema and tables recently

PR #1035, opened 2024-05-13
Add schema cleanup.

calina-c:
Work on #973


pdr_backend/lake/etl.py
```diff
@@ -1,28 +1,27 @@
...
-from pdr_backend.lake.table_pdr_predictions import predictions_table_name
-from pdr_backend.lake.table_pdr_subscriptions import subscriptions_table_name
-from pdr_backend.lake.table_pdr_slots import slots_table_name
-from pdr_backend.lake.table_pdr_truevals import truevals_table_name
+from pdr_backend.ppss.ppss import PPSS
+from pdr_backend.subgraph.payout import Payout
+from pdr_backend.subgraph.prediction import Prediction
+from pdr_backend.subgraph.slot import Slot
+from pdr_backend.subgraph.subscription import Subscription
+from pdr_backend.subgraph.trueval import Trueval
```
idiom-bytes, line 24: I believe the tables should exist in `lake/` not `subgraph/`

pdr_backend/lake/table.py
```diff
@@ -85,10 +85,10 @@ def drop_tables_from_st(pds: PersistentDataStore, type_filter: str, st):
 
 @enforce_types
 class Table:
-    def __init__(self, table_name: str, df_schema: SchemaDict, ppss: PPSS):
+    def __init__(self, dataclass: type, ppss: PPSS):
         self.ppss = ppss
-        self.table_name = table_name
-        self.df_schema = df_schema
+        self.dataclass = dataclass
```
idiom-bytes, line 90: lets go!

idiom-bytes (changes requested):
My request is to:
1. Move the table objects from subgraph/ => lake/
2. Move the test utilities from subgraph/ => `lake/test/` & `subgraph/test/` (3) there might be some considerations for some cleanup here

(1)(2) were good opportunities as parts of this cleanup.
(3) I think most test utilities should make it to `lake/test/` but there might be some sharing between `lake/` and `subgraph/` that requires consideration.

pdr_backend/lake/etl.py
calina-c: for etl models I agree. For raw models, the fetch functions return these object types, as defined in subgraph. Essentially, ETL tables copy subgraph objects at a higher level. They exist in subgraph not because I put them there, but because they are at a more basic level. They were already there, I just added the functions to retrieve lake schema and lake table name on them.

calina-c: TLDR: raw tables are not ETL tables, they are reflections of subgraph

calina-c: update: I've checked. Actually they are pretty "movable" primitives and there are even name clashes in subgraph. I've moved them and I'll re-request a review once the tests pass.

calina-c:
@idiom-bytes it seems all tests have passed, but I'm still not seeing the green dots. Must be something off with Github Actions today :)

idiom-bytes (approved):
thank you! looks good

PR #1039, opened 2024-05-14
fix prepare temp tables fails if production table has no values

### LINKED ISSUE #1038: [Lake][DuckDB] Raw data update fails if there is CSV data for a table but no rows in production table
Ho to reproduce:
- go trough etl raw update flow
- use the drop raw etl command to remove data from the a raw table
- try to run etl raw update again and you should get the following error:

When checking if there is a need for data to be loaded from CSVs, it tries to get timestamp value from table, but because the table is empty the mentioned error is thrown

KatunaNorbert:
Fixes #1038.

- [x] Fixed `lake raw drop` and `lake etl drop` so that they both work with `pdr lake raw drop ppss.yaml sapphire-network 2023-01-01`
- [x] You can now drop rows from raw, or etl, and then resume building from where you left off, getting all the data loaded from the CSV
- [x] You can very quickly execute different loops of: 
1. update raw and view results, then drop part of raw: `raw update -> describe -> raw drop partial`
2. update raw again and verify it went to full: `raw update (to full) -> describe (see full)`
3. now update etl (to build full bronze tables) -> describe (see full bronze tables)
4. now run `lake raw update` to get latest from gql and sync raw data to duckdb
5. then run `lake etl update` to update the bronze tables based on the latest data you just fetched... only processing the new rows that have been fetched.
6. or you can `lake etl drop` everything, then run `lake etl update` and have your bronze tables rebuild in a second.
- [x] ETL is using the right etl_tables (start_ts) and raw_tables (end_ts) mixture to calculate how to process data, and only process the new data required. Both bronze SQL queries also enforce this.

image

### DIFF
--- a/pdr_backend/lake/test/test_gql_data_factory.py
+++ b/pdr_backend/lake/test/test_gql_data_factory.py
@@ -3,7 +3,7 @@ from unittest.mock import MagicMock, patch
 from pdr_backend.lake.gql_data_factory import GQLDataFactory
 from pdr_backend.lake.persistent_data_store import PersistentDataStore
 from pdr_backend.lake.prediction import mock_daily_predictions
-from pdr_backend.lake.table import TableType, get_table_name
+from pdr_backend.lake.table import TableType, TempTable, get_table_name
 from pdr_backend.lake.table_registry import TableRegistry
 from pdr_backend.ppss.ppss import mock_ppss
 from pdr_backend.util.time_types import UnixTimeMs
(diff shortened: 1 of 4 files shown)


pdr_backend/lake/test/test_gql_data_factory.py
```diff
@@ -318,3 +318,71 @@ def test_do_subgraph_fetch_stop_loop_when_restarting_fetch(
...
+    # if production table doesn't have any rows doesn't break the preparation
+    gql_data_factory._prepare_temp_table(
+        table.table_name,
+        UnixTimeMs(1699300800000),
+        UnixTimeMs(1699300800000),
+        table.dataclass.get_lake_schema(),
+    )
+    pds.move_table_data(TempTable(table.table_name), table.table_name)
+
+    assert len(pds.query_data("SELECT * FROM {}".format(table.table_name))) == 6
```
idiom-bytes, line 388: Very happy to see this test written @KatunaNorbert!!! This covers the use-case bang on, and gives me confidence that things are working like we expected them to.

Note: Please look at the fixtures inside `test_etl.py`. Perhaps try re-writing this test after to use the existing setup and reduce the amount of code you need to write.

idiom-bytes (approved):
Fantastic work @KatunaNorbert 

Thank you for isolating everything in here, writing test coverage, and writing enough details throughout the ticket/PR for me to review and complete everything with ease

PR #1043, opened 2024-05-15
Csv datastore simplification

calina-c:
To follow-up on the booleans issue (ref #1015 and maybe #973)

Added rechunks and simplified data management for identifiers. Adds utility to create CSV identifiers from tables directly, but keep CSVDataStore for inspection purposes as well.

### DIFF
--- a/pdr_backend/lake/csv_data_store.py
+++ b/pdr_backend/lake/csv_data_store.py
@@ -1,81 +1,14 @@
 import os
-from typing import List, Optional, Union
+from typing import Optional, Union
+
 import polars as pl
+from enforce_typing import enforce_types
 from polars.type_aliases import SchemaDict
 
-from enforce_typing import enforce_types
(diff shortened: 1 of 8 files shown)


KatunaNorbert (approved):
LGTM

calina-c:
Added the last commit based on suggestions from Mustafa. @kdetry if it looks good to you without the instances part, there was also no point in keeping the high level data store, the new identifiers manage just fine.

idiom-bytes:
Hi, thanks for simplifying the process such that CSV signatures can be built from tables.

But, why did the class have to be renamed?

Not only that doesn't seem needed, but it's unfortunate... I've been trying to deprecate the name "dataset_identifier" for a long time... Now I'm seeing it being used even more... We should just call it a key, or a signature and **avoid creating new terminology unless needed.**

PR #1047, opened 2024-05-15
Fix bronze tables st and end date calculation

### LINKED ISSUE #1046: [Lake][DuckDB] Fix bronze tables start and end date filter calculation
In the current implementation the start and end date for the bronze tables is wrong, and here is why:
- the calculation is done once and it is the same for all the table which is wrong, the calculation should be done for each bronze table just like for each raw table
- the start time is the lowest timestamp from all the highest timestamp values from each **etl table**
- the end time is the lowest timestamp from all the highest timestamp values from each **raw table**

Proposed solution:
- calculate start and end date for each bronze table individually
- start time should be the highest timestamp from the current bronze table
- end time should be the highest timestamp from the corresponding raw table

KatunaNorbert:
Fixes #1046.

### DIFF
--- a/pdr_backend/lake/etl.py
+++ b/pdr_backend/lake/etl.py
@@ -10,7 +10,7 @@ from pdr_backend.lake.persistent_data_store import PersistentDataStore
 from pdr_backend.lake.prediction import Prediction
 from pdr_backend.lake.slot import Slot
 from pdr_backend.lake.subscription import Subscription
-from pdr_backend.lake.table import ETLTable, NamedTable, TableType, TempTable
+from pdr_backend.lake.table import ETLTable, NamedTable, TempTable
 from pdr_backend.lake.table_bronze_pdr_predictions import (
     BronzePrediction,
     get_bronze_pdr_predictions_data_with_SQL,
@@ -206,11 +206,9 @@ class ETL:
         return values
 
     @enforce_types
-    def get_timestamp_values(
-        self, table_names: List[str], default_timestr: str
-    ) -> UnixTimeMs:
+    def get_timestamp_values(self, table_name: str, default_timestr: str) -> UnixTimeMs:
         max_timestamp_values = self._get_max_timestamp_values_from(
-            [NamedTable(tb, TableType.NORMAL) for tb in table_names]
+            [NamedTable(table_name)]
         )
 
         logger.info(
@@ -229,12 +227,14 @@ class ETL:
 
         logger.info("get_timestamp_values - values: %s", values)
         timestamp = (
-            min(values) if len(values) > 0 else UnixTimeMs.from_timestr(default_timestr)
+            values[0] if len(values) > 0 else UnixTimeMs.from_timestr(default_timestr)
         )
         return timestamp
 
     @enforce_types
-    def _calc_bronze_start_end_ts(self) -> Tuple[UnixTimeMs, UnixTimeMs]:
+    def _calc_bronze_start_end_ts(
+        self, table_name: str
+    ) -> Tuple[UnixTimeMs, UnixTimeMs]:
         """
         @description
             Calculate the start and end timestamps for the bronze tables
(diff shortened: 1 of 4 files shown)


kdetry:
LGTM

kdetry (approved):

pdr_backend/lake/etl.py
```diff
@@ -244,11 +244,11 @@ def _calc_bronze_start_end_ts(self) -> Tuple[UnixTimeMs, UnixTimeMs]:
             min(max(source_tables_max_timestamp)).
         """
         from_timestamp = self.get_timestamp_values(
-            self.bronze_table_names, self.ppss.lake_ss.st_timestr
+            table_name, self.ppss.lake_ss.st_timestr
         )
 
         to_timestamp = self.get_timestamp_values(
-            self.raw_table_names, self.ppss.lake_ss.fin_timestr
+            table_name[len("bronze_") :], self.ppss.lake_ss.fin_timestr
```
calina-c, line 251: table.base_name?

pdr_backend/lake/etl.py
```diff
@@ -314,10 +314,10 @@ def update_bronze_pdr(self):
         """
         logger.info("update_bronze_pdr - Update bronze tables.")
 
-        # st_timestamp and fin_timestamp should be valid UnixTimeMS
-        st_timestamp, fin_timestamp = self._calc_bronze_start_end_ts()
-
         for table_name, get_data_func in self.bronze_table_getters.items():
+            # st_timestamp and fin_timestamp should be valid UnixTimeMS
+            st_timestamp, fin_timestamp = self._calc_bronze_start_end_ts(table_name)
```
idiom-bytes, line 319: like we discussed, this should not only be using a single table

idiom-bytes, line 319: This is wrong.
1. All tables should have already been processed up-to-the-same end-ts
2. We do not need to calculate st_ts and en_ts for each table.
3. We can get the max(table1, table2, table3) across all raw_tables as end_ts
4. And the max(table1, table2) across all bronze_table as st_ts

pdr_backend/lake/table_bronze_pdr_predictions.py
```diff
@@ -89,7 +89,7 @@ def get_bronze_pdr_predictions_data_with_SQL(
                 SPLIT_PART({pdr_truevals_table_name}.ID, '-', 1)
         LEFT JOIN {pdr_payouts_table_name}
             ON {pdr_predictions_table_name}.ID = {pdr_payouts_table_name}.ID
-        WHERE {pdr_predictions_table_name}.timestamp >= {st_ms}
+        WHERE {pdr_predictions_table_name}.timestamp > {st_ms}
```
idiom-bytes, line 92: This looks correct and good

pdr_backend/lake/table_bronze_pdr_slots.py
```diff
@@ -101,7 +101,7 @@ def get_bronze_pdr_slots_data_with_SQL(
                 GROUP BY slot_id
             ) AS joined_{pdr_payouts_table_name} ON {pdr_slots_table_name}.ID = joined_{pdr_payouts_table_name}.slot_id
             WHERE
-                pdr_slots.timestamp >= {st_ms}
+                pdr_slots.timestamp > {st_ms}
```
idiom-bytes, line 104: This looks correct and good

pdr_backend/lake/etl.py
```diff
@@ -229,12 +227,14 @@ def get_timestamp_values(
 
         logger.info("get_timestamp_values - values: %s", values)
         timestamp = (
-            min(values) if len(values) > 0 else UnixTimeMs.from_timestr(default_timestr)
+            values[0] if len(values) > 0 else UnixTimeMs.from_timestr(default_timestr)
```
idiom-bytes, line 230: not correct

idiom-bytes:
The original issue is not a bug.

I will be updating the Readme and eventually, visualize this.

PR #1050, opened 2024-05-15
Fix 1046 - Improvements to harden duckdb

### LINKED ISSUE #1046: [Lake][DuckDB] Fix bronze tables start and end date filter calculation

idiom-bytes:
Fixes #1046 

Changes proposed in this PR:
- Fixing ETL st_ts calculation by getting the max(values) rather than min(values)
- Adjusting Bronze SQL queries st_ts condition to be precise
- Adjusting ETL for loop to retrieve table names using the latest implementation

### DIFF
--- a/pdr_backend/lake/etl.py
+++ b/pdr_backend/lake/etl.py
@@ -10,7 +10,7 @@ from pdr_backend.lake.persistent_data_store import PersistentDataStore
 from pdr_backend.lake.prediction import Prediction
 from pdr_backend.lake.slot import Slot
 from pdr_backend.lake.subscription import Subscription
-from pdr_backend.lake.table import ETLTable, NamedTable, TableType, TempTable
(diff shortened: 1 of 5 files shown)


idiom-bytes:
All of these fixes were implemented as part of 1047 - https://github.com/oceanprotocol/pdr-backend/pull/1047
And discussions w/ @kdetry and @KatunaNorbert about fixing ETL by using `max(values)`

I am now merging it into duckdb

PR #1053, opened 2024-05-16
Enforce schema types for lake and bronze mappers

calina-c:
Works on #973.


calina-c:
@idiom-bytes I left one comment in the code in the form of a question. That's causing the pylint failure until we discuss.

idiom-bytes (approved):
Looks good, thank you!

Issue #1058, opened 2024-05-17
[DuckDB][Lake] Expand validation tool - Duplicate rows


idiom-bytes:
### Background / motivation

As part of completing the duckdb + #1000 epic and continuing to aim for high-data quality, we should likely have a better way to validate issues in the lake. Here are some recommendations I made:
- Updating `lake validate` to report duplicate rows on tables
- Updating `lake describe` to: report st_ts, end_ts, also in datestr format (st_ts_datestr, end_ts_datestr)
- Updating `lake describe` to provide a single summary across all tables so we can eventually put it on a dashboard (rather than outputting 12x different results that are hard to stitch together)
- Updating "lake validate gappiness" to work on other raw tables (slots and truevals)

### TODOs / DoD

Expand validation tool such that it tests for duplicate rows, and provides a report about duplicate rows/validations.

Tasks:
- [ ] Write duplicate row queries
- [ ] Aggregate results so we can write a report
- [ ] Generate a report based on tables & output
- [ ] Write duplicate validation errors & messages
- [ ] Write tests showing this validation is working as intended
- [ ] User is able to easily get a duplicate report at the lake level, across all tables
- [ ] User is able to easily get a report at the row level, across all tables

### Related github issues
https://github.com/oceanprotocol/pdr-backend/issues/618
https://github.com/oceanprotocol/pdr-backend/issues/1000

Issue #1000, opened 2024-05-06
[Lake][DuckDB] Verify lake functionality and behavior is working as expected


idiom-bytes:
### Motivation

To verify that
- The lake is working reliably, as expected
- The basic flows and tables are working reliably, as expected
- The commands and lake behavior is working as expected

We need to improve basic reliability and stability of the lake.
The basic duckdb behavior needs to be working as expected.

We should verify things are working by keeping it simple, and focusing on the `bronze_pdr_predictions table`.
I am recommending that we **ignore pdr-subcriptions, pdr-slots, and possibly other tables** so we can validate that the lake is behaving as expected.

![image](https://github.com/oceanprotocol/pdr-backend/assets/69865342/b75a447f-9b42-4400-9007-7a4d90f1b923)

### Verification - Inserting data into the lake and manipulating it
When you first start interacting with the lake, there will be a large a fetch/update step that will try to build everything into the lake. As these records are processed, we begin inserting them into our DB.
1. User should edit ppss such that the st_ts and end_ts to define their lake start & end time.
1. User should run `lake update` command to start fetching data, and fill the whole lake.
1. At any point, the user should be able to pause/cancel and resume w/o any errors
1. Show user how to set end_ts to "Now" so lake continues to update forever.

Once the lake is built, it's very likely that many records will have `null` entries as they are initially inserted into the database. **We are not worried about this for the moment.**

### Test - Cutting off the lake (dropping)

Let's first consider how our lake works. 
A certain amount of data and events arrive that need to be processed. Each time we do a run, we update a certain amount of records. 

|        | Run 1 | Run 2 | Run 3 |
|--------|-------|-------|-------|
| Time   | 1:00  | 2:00  | 3:00  |

Let's say we wanted to drop everything since Run 1. We would call our cli drop command, and get rid of that data.

`pdr lake drop 10000001 my_ppss.yaml sapphire-mainnet`

Which might be the equivalent of dropping all records since Run 1 -> End.
This would include the data from `[Run 2, Run 3]`.

The user would continue updating the lake by calling `pdr lake update`... which would refetch and rebuild `[Run 2, Run 3]`, getting the system up-to-date, and then continuing on from there,

### Verifying
We could consider that by dropping/cutting off part of the lake, all tables would have the same data cut-off/rows-dropped like below. Such that the data pipeline can resume from here, and all tables can be updated/resumed from the same "height".
```
WITH 
max_pdr_predictions AS (
[...]

PR #1060, opened 2024-05-17
Add HTML dashboard for describe/validation report

calina-c:
Work on #1058.

### DIFF
--- a/pdr_backend/lake/lake_info.py
+++ b/pdr_backend/lake/lake_info.py
@@ -1,4 +1,6 @@
+from dash import Dash, html
 from typing import Dict, List
+import dash_bootstrap_components as dbc
 
 import polars as pl
 from polars.dataframe.frame import DataFrame
@@ -10,12 +12,13 @@ pl.Config.set_tbl_hide_dataframe_shape(True)
 
 
 class LakeInfo:
-    def __init__(self, ppss: PPSS):
-        self.pds = PersistentDataStore(ppss.lake_ss.lake_dir)
+    def __init__(self, ppss: PPSS, html: bool = False):
+        self.pds = PersistentDataStore(ppss.lake_ss.lake_dir, read_only=True)
         self.all_table_names: List[str] = []
         self.table_info: Dict[str, DataFrame] = {}
         self.all_view_names: List[str] = []
         self.view_info: Dict[str, DataFrame] = {}
+        self.html = html
 
     def generate(self):
         self.all_table_names = self.pds.get_table_names(all_schemas=True)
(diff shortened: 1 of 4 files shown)


idiom-bytes:
Hi @calina-c, I need to review this a bit more... my only comment and I didn't think about this enough is that we're still focusing on #1000 and the epic.

Improving validation + describe CLI commands are needed for us to debug + fix the issues.
Moving this stuff to the app, is a nice-to have.

The key thing we need to focus on is fixing data quality.
This means debugging the data pipeline and solving issues that can cause duplication.

pdr_backend/lake/lake_info.py
```diff
@@ -1,22 +1,45 @@
...
+from pdr_backend.lake.gql_data_factory import GQLDataFactory
 from pdr_backend.lake.persistent_data_store import PersistentDataStore
+from pdr_backend.lake.renderers.cli import CliRenderer
+from pdr_backend.lake.renderers.html import HtmlRenderer
+from pdr_backend.lake.table_bronze_pdr_predictions import BronzePrediction
 from pdr_backend.ppss.ppss import PPSS
 
 pl.Config.set_tbl_hide_dataframe_shape(True)
 
+logger = logging.getLogger("LakeValidate")
```
idiom-bytes, line 18: LakeInfo

pdr_backend/lake/lake_validate.py
```diff
@@ -17,171 +12,6 @@
 
 @enforce_types
 class LakeValidate(LakeInfo):
```
idiom-bytes, line 14: should we delete this?

idiom-bytes (changes requested):
looks good, but could use a couple of tweaks

calina-c:
> Hi @calina-c, I need to review this a bit more... my only comment is that I didn't think about this enough, and that we're still focusing on #1000 and the epic...

Sure, take your time. I will present some arguments here, point by point.

> this flows here, is not what we're trying to do here... which is all on the CLI.

Firstly, this PR does not impact other work done on the CLI. This is just data presentation, not data manipulation, as can be observed through the read_only argument. Actually, that part is also a bug fix. During development, I encountered a serious testing issue where the read_only argument was mixed in tests and never corrected. Please consider at least merging some bugfixes related to that param.

> Thus i feel that: CLI/Improvements PR I prepared (that's now merged here) is needed for us to debug + fix the issues. Serving this data to the app right now, is a nice-to have.

Secondly, while this is a nice to have, it is also a debugging tool in itself. Visualising the data is difficult on the CLI and I think this code can help the debugging process in #1000 by clearly showing the content and validation issues that Lake has. As a person with some attention-related issues, I found it difficult to debug without a proper visualisation tool. The CLI results fill a lot of space and are difficult to browse.

Btw. my first thought was just to publish a basic html report instead of a dashboard, where I could at least Ctrl+F through it, but given that the dependency for Dash already exists, it was actually faster to develop this way.

> I just want to still raise the focus on the engine-level work. More so, we need to stop adding code to duckdb-pr that's not required to go in there and get it out the door.

Thirdly, regarding the code that "is ready", this is just an experimental debugging tool and not meant as a serious data inspection DBeaver alternative. It doesn't necessarily have to be perfect, just functional enough for us to be able to debug. This is also in reference to my first point, where it is not breaking anything related to the ETL code itself. The "experimental" nature only applies to the dashboard itself, where (in my opinion) we can afford the risk of being bold and experimental.

Fourthly, I agree there are structure changes as well, where parts of the validation was moved to LakeInfo. I agree that might seem like an architecture change but since you inherited LakeInfo on LakeValidate the changes make a lot of sense.

> Fixing the data pipeline and SQL queries is the name of the game right now. 🙌

I do think the info presented as such will help with debugging the pipeline and it will be also easy to add other specific information we need for debugging, whereas the CLI result itself has reached a bottleneck of UX (or, rather, developer X... 😅 ) right now.

pdr_backend/lake/lake_validate.py
calina-c: I don't think so. We still have the same CLI validation independently of the dashboard. CLI describe command still works exactly as before (+ your timestr changes). It's only the dashboard that combines Info + Validation as a full report on the ETL.

idiom-bytes (approved):
Approving PR

idiom-bytes:
I have read everything you said, and I agree with it all.

I want to be clear: "I love the work you're doing"

I'm just particularly focused on getting the duckdb update out the door in a very good, reliable state.

idiom-bytes:
Ran everything locally
image

One other thing I noticed is that you need to type `--html true` rather than `--HTML` like the other implementations.
Screenshot from 2024-05-20 09-05-44

I documented how to run describe --html in the comments

idiom-bytes:
I updated the code to use `--HTML` as the parameter to trigger the html/dash site

PR #1067, opened 2024-05-20
Feature #1061 - Set minimum slot threshold for payouts

### LINKED ISSUE #1061: Set Minimum Slot Threshold for Payouts
The feature allows users to define a minimum number of slots that must be available before pdr-backend claims payouts **to save gas**.

trizin:
Fixes #1061

Changes proposed in this PR:

- `count_unique_slots` to count the unique slots returned from `find_shared_slots` function.
- New predictoor_ss bot_only flag, `min_payout_slots`
- Check `min_payout_slots` before claiming OCEAN payouts, if total number of slots claimable smaller than that, skip payouts.

### DIFF
--- a/pdr_backend/payout/payout.py
+++ b/pdr_backend/payout/payout.py
@@ -132,24 +132,32 @@ def do_rose_payout(ppss: PPSS, check_network: bool = True):
                 receipt["transactionHash"],
             )
             return
-        logger.info("Transfering wROSE to owner")
         time.sleep(4)
-        balance = wROSE.balanceOf(pred_submitter_mgr.contract_address)
-        receipt = pred_submitter_mgr.transfer_erc20(
-            wROSE_addr, web3_config.owner, balance, True
-        )
-        if receipt["status"] != 1:
-            logger.warning(
-                "Failed to transfer wROSE tokens to the owner, tx: %s",
-                receipt["transactionHash"],
-            )
-            return
     else:
         logger.warning("No rewards available to claim")
-        return
+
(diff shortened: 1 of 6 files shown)


pdr_backend/ppss/predictoor_ss.py
```diff
@@ -90,6 +90,12 @@ def s_until_epoch_end(self) -> int:
     def pred_submitter_mgr(self) -> str:
         return self.d["bot_only"]["pred_submitter_mgr"]
 
+    @property
+    def min_payout_slots(self) -> int:
+        if "min_payout_slots" not in self.d["bot_only"]:
```
calina-c, line 95: I would suggest: if "bot_only" not in self.d: return 0
return d["bot_only"].get("min_payout_slots", 0)

or a one-liner:
self.d.get('bot_only', {}).get('min_payout_slots', 0)

pdr_backend/predictoor/util.py
```diff
@@ -40,6 +40,23 @@ def find_shared_slots(
...
+    Calculates the number of unique slots from the output of the `find_shared_slots`.
+
+    @param shared_slots: List[Tuple[List[str], List[int]]]
+        The output of the find_shared_slots function.
+
+    @return
+    int: The number of unique slots.
+    """
+    all_slots = []
+    for _, slots in shared_slots:
```
calina-c, line 54: return len({all_slots.extend(sl) for sl in shared_slots.values()})
I think {} will be enough for it to be a set, but if not you can use the `set` syntax as well

calina-c (approved):

PR #1069, opened 2024-05-20
[DuckDB][GQL] Fix Issue1068 - Keep fixed end_ts across all table updates + fetches

### LINKED ISSUE #1068: [DuckDB] Fix Checkpoint Logic - Drifting "Now"
**Describe the bug**

### Problem 1
We want all raw_tables to be updated to the same end_ts
image

And since the loop in GQLDF uses self.ppss.lake_ss.fin_timestamp (below), under-the-hood, the end_ts will keep shifting as we keep fetching records from subgraph... So we need to move it up, outside the _update loop.

image

**To Reproduce**
- Use lake_ss.end_ts: "now"
- Fetch raw lake data...
- Notice how `end_ts` for tables has shifted significantly as tables were fetched.

### DoD
- [x] Fix problem 1 such that end_ts is consistent across all tables in GQLDF
- [x] Verify that tables are syncing to the same `end_ts`

idiom-bytes:
Fixes #1068 

image

Changes proposed in this PR:
- [x] Keep end_ts fixed across all tables during GQLDF procedure

### DIFF
--- a/pdr_backend/lake/gql_data_factory.py
+++ b/pdr_backend/lake/gql_data_factory.py
@@ -285,12 +285,13 @@ class GQLDataFactory:
         @arguments
             fin_ut -- a timestamp, in ms, in UTC
         """
+        fin_ut = self.ppss.lake_ss.fin_timestamp
+
         for table in (
             TableRegistry().get_tables(self.record_config["gql_tables"]).values()
         ):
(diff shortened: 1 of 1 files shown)


KatunaNorbert (approved):
Tested and it's looking good, end_ts is the same

PR #1071, opened 2024-05-20
Fix Issue1070 - Unix time natural_language: "now" and other commands are generating local time

### LINKED ISSUE #1070: [DuckDB][UnixTimeMs & UnixTimeS] - Time functions are returning local timestamp, not UTC
**Describe the bug**

Hi @calina-c, I have tested the issue further and the current UTC flags & implementations are not returning an UTC timestamp, but rather, a local timestamp. This is leading to a variety of data problems.

Problem (1) many of our UnixTImeS + UnixTImeMs functions are returning local timestamp results.

TLDR; when asked GPT about it... "The replace method in Python's datetime module is used to replace specified date attributes from a datetime object. However, it doesn't perform any timezone conversions. It simply replaces the tzinfo attribute with the specified timezone, in this case timezone.utc."

I wrote a simple test to see if the unix timestamp was close to the UTC values that should have been yielded.. It verifies that how we use "natural language" to get "now" isn't the same as UnixTimeS.now()

Problem (2) as you can see, using "now" will return a local timestamp now, rather than a unix (or GMT/UTC) now.

Problem (3) is that @kdetry or perhaps others, are using it wrongly I believe in parts of the code.
```
# from app.py line 237
ms_end_ts_from_utc = UnixTimeMs(int(datetime.utcnow().timestamp())) #in seconds
s_end_ts_from_utc = UnixTimeS(int(datetime.utcnow().timestamp())) #in seconds
```

It also checks other implementations (like what's going on in app.py, and points out that it's wrong) @kdetry.
image

We can see that all of the unix tmestamp now() commands worked... except for the first, natural language one. The first one, is giving us local timestamp. (UTC-7)
```
- ms_end_ts_from_timestr_now 1716213309546
- ms_end_ts_from_utc 1716263709
- s_end_ts_from_utc 1716263709
- time.time() 1716238509
- UnixTimeMs.now() 1716213309546
- UnixTimeS.now() 1716213309
```
image

We're trying to sync to "now" but subscriptions subgraph is fetching from 7 hours in the past... 
image

What I want, is lake_ss.end_ts("now") to return time.time(), such that we can fetch the right data from subgraph and beyond.
image

### DoD:
- [x] UnixTimeS and UnixTimeMs shouldn't operate off local timestampimplement
- [x] "now" should return UTC time "now", not local time "now"

idiom-bytes:
Fixes #1070

We're running into issues where UnixTimeS and UnixTimeMs respond w/ timestamps that should be in UTC, but in reality are in local time. This is causing all sorts of issues in the data.

Changes proposed in this PR:
- [x] Not yet fixed, I have created a test to show the problem
- [x] Using "now" should generate the correct timestamp (not local time, but UTC)

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -62,6 +62,9 @@ ignore_missing_imports = True
 [mypy-pytest]
 ignore_missing_imports = True
 
+[mypy-pytz.*]
+ignore_missing_imports = True
+
 [mypy-sapphirepy.*]
 ignore_missing_imports = True
 
--- a/pdr_backend/lake/gql_data_factory.py
+++ b/pdr_backend/lake/gql_data_factory.py
@@ -295,7 +295,7 @@ class GQLDataFactory:
             logger.info(
                 "      Aim to fetch data from start_time: [%s] to end_time: [%s]",
                 st_ut.pretty_timestr(),
-                fin_ut.pretty_timestr()
+                fin_ut.pretty_timestr(),
             )
             if st_ut > min(UnixTimeMs.now(), fin_ut):
                 logger.info("      Given start time, no data to gather. Exit.")
--- a/pdr_backend/lake/gql_data_factory.py
+++ b/pdr_backend/lake/gql_data_factory.py
@@ -285,12 +285,13 @@ class GQLDataFactory:
         @arguments
             fin_ut -- a timestamp, in ms, in UTC
         """
+        fin_ut = self.ppss.lake_ss.fin_timestamp
+
         for table in (
             TableRegistry().get_tables(self.record_config["gql_tables"]).values()
         ):
             # calculate start and end timestamps
             st_ut = self._calc_start_ut(table)
-            fin_ut = self.ppss.lake_ss.fin_timestamp
             logger.info(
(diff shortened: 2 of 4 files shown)


calina-c:
Ref the failing CI tests, @idiom-bytes do you want to keep/adjust this test in particular, or will it suffice to add the timezone? I can (and probably will) add some tests using freezegun, which is much safer wrt. defining the time in tests (I think it already is a dependency)

KatunaNorbert:
> @KatunaNorbert is time.time() or time.gmtime() used in any relevant way for the ETL? If so, I think we need to reconsider those as well.

From what I see they are not used inside the ETL

idiom-bytes:
@Norbert @calina-c 

```From what I see they are not used inside the ETL```

time.time() isn't used but we're trying to use "now"

```
Tested it and the end date it's fixed now and it uses the local time to fetch from subgraph. Both of the bronze tables are now updating, but bronze_predictions table it's a little bit behind the pdr_predictions table. Slots tables are up to date.
```

I'm confused, I thought that we were making it so that it would use unix time?

PR #1077, opened 2024-05-21
Fix Issue1076 - Scope down queries, Fix drop, Improve DRY, Stabilize e2e ETL

### LINKED ISSUE #1076: [DuckDB][ETL] Scope down Queries/ETL + Increase base quality
**Describe the bug**
Like I had proposed before... we're still having issues in our queries/ETL that are causing a bunch of distractions. These are queries that extend beyond our core focus and we can increase in complexity after the basics are working well (example: slots).
image

Our release candidate only needs 4: predictions, truevals, payouts + bronze_predictions
Screenshot from 2024-05-21 08-15-24

Once our pipeline is working end-to-end reliably, then, increase in complexity and add those queries back in. As you can see from the image above, by reducing the problem we can see that our bronze table `min_datestr` is not behaving as expected.

Isolate our target, get the whole thing working correctly end-to-end, then expand.

### DoD
- Scope down queries
- Increase quality of ETL

### Tasks
- [x] Disable slots and subscriptions raw_queries
- [x] Disable bronze_slots table
- [x] Fix issues across all 4 tables such that they are working as expected, end-to-end

idiom-bytes:
Fixes #1076 

bronze_predictions and other tables are starting + ending like how we would expect it to, at the end of the first run...
image

I had to...:
- [x] Fix drop logic [#1079](https://github.com/oceanprotocol/pdr-backend/issues/1079)

**...I have tested it multiple runs and other variations and it remained stable...**
image

Changes proposed in this PR:
- [x] Removed slots, and subscriptions queries from GQLDF
- [x] Removed bronze_slots query from ETL
- [x] Updated tests to verify expected tables
- [x] Run pipeline e2e
- [x] Verify that expected tables and results are being yielded
- [x] Try a variety of drops/updates/validate/describes, and verify that the pipeline is running reliably, without generating gaps or dupes
- [x] Fixed "temp_tables" util functionality inside ETL that wasn't correctly implemented. This bug would potentially keep temp tables hanging around, although they should be cleaning up on their own.
- [x] Implemented Calina's feedback to reduce DRY violations WRT GQLDF + ETL configuration
- [x] Improved may tests that are hard-coded, to instead validate against the expected number of tables registered, or records in the dataframe, in such a way where we don't have to modify numbers manually/use fewer magic numbers.

### DIFF
--- a/pdr_backend/cli/cli_module_lake.py
+++ b/pdr_backend/cli/cli_module_lake.py
@@ -10,7 +10,6 @@ from pdr_backend.lake.lake_validate import LakeValidate
 from pdr_backend.lake.persistent_data_store import PersistentDataStore
 from pdr_backend.lake.table import drop_tables_from_st
 from pdr_backend.ppss.ppss import PPSS
-from pdr_backend.util.time_types import UnixTimeMs
 
 logger = logging.getLogger("cli")
 
@@ -64,7 +63,7 @@ def do_lake_query(args, ppss):
 @enforce_types
 def do_lake_raw_drop(args, ppss):
     pds = PersistentDataStore(ppss.lake_ss.lake_dir, read_only=False)
-    drop_tables_from_st(pds, "raw", UnixTimeMs(args.ST))
+    drop_tables_from_st(pds, "raw", args.ST)
 
 
 @enforce_types
@@ -88,7 +87,7 @@ def do_lake_raw_update(_, ppss):
 @enforce_types
 def do_lake_etl_drop(args, ppss):
     pds = PersistentDataStore(ppss.lake_ss.lake_dir, read_only=False)
-    drop_tables_from_st(pds, "etl", UnixTimeMs(args.ST))
+    drop_tables_from_st(pds, "etl", args.ST)
 
 
 @enforce_types
--- a/pdr_backend/cli/cli_module_lake.py
+++ b/pdr_backend/cli/cli_module_lake.py
@@ -10,6 +10,7 @@ from pdr_backend.lake.lake_validate import LakeValidate
 from pdr_backend.lake.persistent_data_store import PersistentDataStore
 from pdr_backend.lake.table import drop_tables_from_st
 from pdr_backend.ppss.ppss import PPSS
+from pdr_backend.util.time_types import UnixTimeMs
 
 logger = logging.getLogger("cli")
 
@@ -63,7 +64,7 @@ def do_lake_query(args, ppss):
 @enforce_types
 def do_lake_raw_drop(args, ppss):
     pds = PersistentDataStore(ppss.lake_ss.lake_dir, read_only=False)
-    drop_tables_from_st(pds, "raw", args.ST)
+    drop_tables_from_st(pds, "raw", UnixTimeMs(args.ST))
 
 
 @enforce_types
@@ -87,7 +88,7 @@ def do_lake_raw_update(_, ppss):
 @enforce_types
 def do_lake_etl_drop(args, ppss):
     pds = PersistentDataStore(ppss.lake_ss.lake_dir, read_only=False)
-    drop_tables_from_st(pds, "etl", args.ST)
+    drop_tables_from_st(pds, "etl", UnixTimeMs(args.ST))
 
 
 @enforce_types
--- a/pdr_backend/cli/cli_module_lake.py
+++ b/pdr_backend/cli/cli_module_lake.py
@@ -7,7 +7,7 @@ from pdr_backend.lake.etl import ETL
 from pdr_backend.lake.gql_data_factory import GQLDataFactory
 from pdr_backend.lake.lake_info import LakeInfo
 from pdr_backend.lake.lake_validate import LakeValidate
-from pdr_backend.lake.persistent_data_store import PersistentDataStore
+from pdr_backend.lake.duckdb_data_store import DuckDBDataStore
 from pdr_backend.lake.table import drop_tables_from_st
 from pdr_backend.ppss.ppss import PPSS
 
@@ -51,9 +51,9 @@ def do_lake_query(args, ppss):
     @description
         Query the lake for a table or view
     """
-    pds = PersistentDataStore(ppss.lake_ss.lake_dir, read_only=True)
+    db = DuckDBDataStore(ppss.lake_ss.lake_dir, read_only=True)
     try:
-        df = pds.query_data(args.QUERY)
+        df = db.query_data(args.QUERY)
         print(df)
     except Exception as e:
         logger.error("Error querying lake: %s", e)
@@ -62,8 +62,8 @@ def do_lake_query(args, ppss):
 
 @enforce_types
 def do_lake_raw_drop(args, ppss):
-    pds = PersistentDataStore(ppss.lake_ss.lake_dir, read_only=False)
-    drop_tables_from_st(pds, "raw", args.ST)
+    db = DuckDBDataStore(ppss.lake_ss.lake_dir, read_only=False)
+    drop_tables_from_st(db, "raw", args.ST)
 
 
 @enforce_types
@@ -86,8 +86,8 @@ def do_lake_raw_update(_, ppss):
 
 @enforce_types
 def do_lake_etl_drop(args, ppss):
-    pds = PersistentDataStore(ppss.lake_ss.lake_dir, read_only=False)
-    drop_tables_from_st(pds, "etl", args.ST)
+    db = DuckDBDataStore(ppss.lake_ss.lake_dir, read_only=False)
+    drop_tables_from_st(db, "etl", args.ST)
 
 
 @enforce_types
--- a/pdr_backend/lake/gql_data_factory.py
+++ b/pdr_backend/lake/gql_data_factory.py
@@ -9,8 +9,6 @@ from pdr_backend.lake.payout import Payout
 from pdr_backend.lake.persistent_data_store import PersistentDataStore
 from pdr_backend.lake.plutil import _object_list_to_df
 from pdr_backend.lake.prediction import Prediction
-from pdr_backend.lake.slot import Slot
-from pdr_backend.lake.subscription import Subscription
 from pdr_backend.lake.table import NamedTable, Table, TableType, TempTable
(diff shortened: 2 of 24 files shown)


idiom-bytes:
After playing around for a while and feeling confident, i ran the validation pipeline and noticed errors.

image

This is bad because I only noticed duplicate errors when i run `validation` although I've been running `describe` for a while... again, feels like some basic validation on describe would be useful.

However, after reflecting a bit... I realized that I was testing deleting CSV files from my hardrive and then rescuinig the lake... which meant that the duplicate records that I was getting, were inside the duckdb table...

So, I dropped the Raw tables, knowing they should be rebuilt very quickly from the CSV data, and then rebuilt the ETL... both Raw + ETL came up quickly, and right back to where they should be.

image

pdr_backend/lake/gql_data_factory.py
```diff
@@ -59,23 +56,19 @@ def __init__(self, ppss: PPSS):
...
                 Payout: fetch_payouts,
-                Slot: fetch_slots,
             },
             "config": {
                 "contract_list": contract_list,
             },
             "gql_tables": [
                 dn.get_lake_table_name()  # type: ignore[attr-defined]
-                for dn in [Prediction, Subscription, Trueval, Payout, Slot]
+                for dn in [Prediction, Trueval, Payout]
```
calina-c, line 67: I would suggest adding a constant in lake_mapper called REGISTERED_LAKE_TYPES or something similar. It seems this list of types is a repeating thing. In all fairness, I did notice this while working on the ETL but I was lazy to do it myself :)

pdr_backend/cli/cli_module_lake.py
```diff
@@ -63,7 +64,7 @@ def do_lake_query(args, ppss):
 @enforce_types
 def do_lake_raw_drop(args, ppss):
     pds = PersistentDataStore(ppss.lake_ss.lake_dir, read_only=False)
-    drop_tables_from_st(pds, "raw", args.ST)
+    drop_tables_from_st(pds, "raw", UnixTimeMs(args.ST))
```
calina-c, line 67: please remember to remove this conversion here, as well. See comments in #1080

pdr_backend/lake/gql_data_factory.py
idiom-bytes: I mean... we are looking inside the `self.record_config` param from the GQLDataFactory object... This gets used inside the ETL to create the lake.

I feel like it's pretty self-explanatory but I'll improve this so we only have to define it once, and the nomenclature is better.

idiom-bytes:
**[Fix ETL - Move Temp Table to Live + Drop Temp Tables]**
There were some bugs/bad assumptions with how the "Temp Table" logic works inside etl.py... I believe i have fixed that now, and updated the tests to match this.

**[Table Registry]**
Due to the new NamedTables, I think we can get rid of the TableRegistry.

pdr_backend/lake/test/test_gql_data_factory.py
```diff
@@ -173,7 +173,7 @@ def test_get_gql_tables(mock_update):
 
     gql_dfs = gql_data_factory.get_gql_tables()
 
-    assert len(gql_dfs.items()) == 5
+    assert len(gql_dfs.items()) == 3
```
idiom-bytes, line 176: This changed due to the 2 raw tables we stopped processing (slots and subscrptions)

pdr_backend/lake/test/test_etl.py
```diff
@@ -222,21 +217,31 @@ def test_move_from_temp_tables_to_live(setup_data):
...
+    # check all temp tables are dropped, and regular bronze_tables exist
     table_names = pds.get_table_names()
-    assert len(table_names) == 8
-    assert "c" in table_names
-    assert "a" in table_names
-    assert "b" in table_names
+    assert all(
+        table in table_names for table in etl.bronze_table_names
+    ), "Not all temporary bronze tables were moved to live tables successfully"
+    assert len(table_names) == 6
```
idiom-bytes, line 244: It's not every table that gets moved ("a", "b", "c"), it's the bronze temp tables for the ETL that do.

pdr_backend/lake/gql_data_factory.py
idiom-bytes: Done, please review.

pdr_backend/cli/cli_module_lake.py
idiom-bytes: Done, please review.

KatunaNorbert:
I'm getting the following error when running the describe cli command:
Screenshot 2024-05-23 at 10 01 56

pdr_backend/lake/gql_data_factory.py
```diff
@@ -55,28 +59,26 @@ def __init__(self, ppss: PPSS):
...
             "gql_tables": [
                 dn.get_lake_table_name()  # type: ignore[attr-defined]
-                for dn in [Prediction, Subscription, Trueval, Payout, Slot]
+                for dn in _GQLDF_REGISTERED_LAKE_TABLES
             ],
         }
 
         TableRegistry().register_tables(
-            [Prediction, Subscription, Trueval, Payout, Slot], self.ppss
+            list(_GQLDF_REGISTERED_LAKE_TABLES.keys()), self.ppss
```
calina-c, line 75: I don't think you need the list cast, but I'm not sure.

pdr_backend/lake/test/test_etl.py
```diff
@@ -41,7 +41,7 @@ def test_etl_tables(
     assert len(pdr_payouts_df) == 4
     assert len(pdr_predictions_df) == 5
     assert len(pdr_truevals_df) == 5
-    assert len(TableRegistry().get_tables()) == 7
+    assert len(TableRegistry().get_tables()) == 4
```
calina-c, line 44: instead of hardcoding the tests, I think we should assert lengths like either:
- length of bronze table types
- length of etl table types
- combinations of the two

this will make it s.t. we don't have to modify tests when we add/remove tables from the defined list

pdr_backend/lake/test/test_etl.py
```diff
@@ -189,28 +183,29 @@ def test_drop_temp_sql_tables(setup_data):
     etl, pds, _ = setup_data
 
     # SELECT ALL TABLES FROM DB
-    table_names = pds.get_table_names()
+    original_tables = pds.get_table_names()
+    assert len(original_tables) == 5
```
calina-c, line 187: and especially here!
thoughout the implementation, asserts should be in relation to something else in the code, not a hardcoded number

calina-c (changes requested):

KatunaNorbert:
The cli command description for describing the lake misses the network parameter:
`pdr lake describe PPSS_FILE --HTML` -> `pdr lake describe PPSS_FILE NETWORK --HTML`

KatunaNorbert:
I run the ETL a few times and the number of rows within `pdr_predictions` table has always been different that the number of rows in the `bronze_pdr_predictions` table
Screenshot 2024-05-23 at 10 13 47
Screenshot 2024-05-23 at 10 17 58
Screenshot 2024-05-23 at 10 32 33

KatunaNorbert:
I deleted the lake with all the tables and I'm using:
start_d: 1d ago
end_d: now
I'll try to rerun and share the whole data

idiom-bytes:
@KatunaNorbert i'm pretty sure the reason for the problem is the st_ts: "1d ago" and configuring a sliding window for the lake st_ts and end_ts params.

KatunaNorbert:
Yes, that's it, I retested with a fixed date and it's looking good

idiom-bytes:
```
Found the issue and pushed a fix, etl.raw_table_name property was accessed but that doesn't exist. I removed it from the validation function and replaced etl.raw_table_name + etl.bronze_table_names with etl.all_table_names. Also removed that from etl tests
```
Reviewed fix and implementation. ty.

pdr_backend/lake/gql_data_factory.py
idiom-bytes: "Given the function register_tables that you've shown, which expects a list of data classes (List[Type[LakeMapper]]), you will indeed need to convert the keys from _GQLDF_REGISTERED_LAKE_TABLES.keys() to a list if those keys represent data class types that are compatible with the register_table method within register_tables."

More simply
- we're not defining the variable and the type
- keys() does not enforce a list
- because we're using type enforce, we should define the type or cast to a type, in order to enforce it

pdr_backend/lake/test/test_etl.py
idiom-bytes: Yes, I've been trying to update some of these. I'll do it here too.

pdr_backend/lake/test/test_etl.py
idiom-bytes: yeah, i tried to do it throughout this whole test... i skipped this one because there are default tables in duckdb "main" schema... i'll fix this.

idiom-bytes:
I have noticed that `subscription` table was half implemented into the fixture/coverage, and am removing it from the ETL fixture such that we implement it properly when subscriptions are enabled.
https://github.com/oceanprotocol/pdr-backend/issues/1085

This should happen after duckdb PR first-candidate.

I'm simply disabling subscription from ETL fixture for now.

pdr_backend/lake/test/test_etl.py
idiom-bytes: I have fixed this one and several others...

pdr_backend/lake/test/test_etl.py
```diff
@@ -393,7 +393,6 @@ def test_calc_bronze_start_end_ts(tmpdir):
     )
 
     etl = ETL(ppss, gql_data_factory)
-    etl.raw_table_names = ["raw_table_1", "raw_table_2", "raw_table_3"]
```
idiom-bytes, line 391: These have moved to gql_data_factory and is not the right way to fix the test.

kdetry (approved):
lgtm

idiom-bytes:
```lgtm, but I think we shouldn't remove the slots process from the GQL```

We're removing everything non-critical from the core-flows and adding the complexity back in, bit-by-bit.

PR #1080, opened 2024-05-21
Fix Issue 1079: Enforce UnixTimeMs when dropping tables

### LINKED ISSUE #1079: [Lake][CLI] Fix drop logic
**Describe the bug**
Right now, drop is dropping everything because it doesn't enforce UnixTimeMs.

image

To fix this, enforce type
image

and update CLI calls
image

idiom-bytes:
Fixes #1079 

Currently, all records are being dropped because timestamp is not being cast to Ms... so everything is Ms > S.

Changes proposed in this PR:
- [x] Enforce UnixTimeMs in drop function, such that we drop the right timestamp
- [x] Update CLI commands to cast st_ts to UnixTimeMs
- [x] Drop CLI command is now fixed

### DIFF
--- a/pdr_backend/cli/cli_module_lake.py
+++ b/pdr_backend/cli/cli_module_lake.py
@@ -10,7 +10,6 @@ from pdr_backend.lake.lake_validate import LakeValidate
 from pdr_backend.lake.persistent_data_store import PersistentDataStore
 from pdr_backend.lake.table import drop_tables_from_st
 from pdr_backend.ppss.ppss import PPSS
-from pdr_backend.util.time_types import UnixTimeMs
 
 logger = logging.getLogger("cli")
 
@@ -64,7 +63,7 @@ def do_lake_query(args, ppss):
 @enforce_types
 def do_lake_raw_drop(args, ppss):
     pds = PersistentDataStore(ppss.lake_ss.lake_dir, read_only=False)
-    drop_tables_from_st(pds, "raw", UnixTimeMs(args.ST))
+    drop_tables_from_st(pds, "raw", args.ST)
 
 
 @enforce_types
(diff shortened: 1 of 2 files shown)


pdr_backend/cli/cli_module_lake.py
```diff
@@ -63,7 +64,7 @@ def do_lake_query(args, ppss):
 @enforce_types
 def do_lake_raw_drop(args, ppss):
     pds = PersistentDataStore(ppss.lake_ss.lake_dir, read_only=False)
-    drop_tables_from_st(pds, "raw", args.ST)
+    drop_tables_from_st(pds, "raw", UnixTimeMs(args.ST))
```
calina-c, line 67: this is not needed. The arg parser does the conversion and these are parsed args.

https://github.com/oceanprotocol/pdr-backend/blob/35b3c948e2fda776ab8ae21a6e02c35f6071707f/pdr_backend/cli/cli_arguments_lake.py#L23

pdr_backend/lake/table.py
```diff
@@ -35,7 +36,7 @@ def is_etl_table(table_name: str) -> bool:
 
 
 @enforce_types
-def drop_tables_from_st(pds: PersistentDataStore, type_filter: str, st):
+def drop_tables_from_st(pds: PersistentDataStore, type_filter: str, st: UnixTimeMs):
```
calina-c, line 39: this is the only required change. Please revert the unnecessary conversions in cli_module_lake.

calina-c (changes requested):

pdr_backend/cli/cli_module_lake.py
calina-c: n.b. if args.ST were not a UnixTimeMs already, you would have to do UnixTimeMs.from_timestr(). Since args.ST is already an instance of UnixTimeMs and the constructor of UnixTimeMs takes in an int, what this change does is to reconvert the UnixTimeMs to an int and then back into a UnixTimeMs.

pdr_backend/lake/table.py
idiom-bytes: not sure what mypy will say, but sure

idiom-bytes: done, ty

calina-c (approved):

PR #1083, opened 2024-05-23
Fix 769: Rename from <persistent-data-store> to <duckdb-data-store>

### LINKED ISSUE #769: [Lake][OHCLV] Update OHCLV data_factory to use csv_data_writer
### Motivation
OHLCV is still writing/reading from parquet.

We may instead want to use the CSVDataStore + PersistentDataStore to save/read from. But doing this right now is hard, OHLCV code is tightly-bound w/ saving/loading/etc.

### DoD
- [x] OHCLV Data Factory CSV/Saving/Querying code has been abstracted away and is using CSVDataStore + PersistentDataStore
- [x] OHCLV Data Factory and DataStore logic have been separated
- [x] OHCLV Data Factory is querying from PersistentDataStore 
- [x] OHCLV Data Factory tests are up-to-date and every test is passing

------------

### Update According to Trent's comment
- [x] Change the IO methods to a function, in a new module cvsutil.py
- [x] use them into the CSVDataStore class and make it cleaner and thinner

idiom-bytes:
Fixes #769 

Changes proposed in this PR:
- [x] Renamed everything from persistent_data_store -> duckdb_data_store

### DIFF
--- a/pdr_backend/lake/etl.py
+++ b/pdr_backend/lake/etl.py
@@ -6,7 +6,7 @@ from enforce_typing import enforce_types
 
 from pdr_backend.lake.gql_data_factory import GQLDataFactory
 from pdr_backend.lake.payout import Payout
-from pdr_backend.lake.persistent_data_store import PersistentDataStore
+from pdr_backend.lake.duckdb_data_store import DuckDBDataStore
 from pdr_backend.lake.prediction import Prediction
 from pdr_backend.lake.slot import Slot
 from pdr_backend.lake.subscription import Subscription
@@ -73,9 +73,9 @@ class ETL:
             If exists, drop them and rebuild
         """
         # drop the tables if it exists
-        pds = PersistentDataStore(self.ppss.lake_ss.lake_dir)
+        duckDB = DuckDBDataStore(self.ppss.lake_ss.lake_dir)
         for table_name in self.temp_table_names:
-            pds.drop_table(TempTable(table_name).fullname)
+            duckDB.drop_table(TempTable(table_name).fullname)
 
     def _move_from_temp_tables_to_live(self):
         """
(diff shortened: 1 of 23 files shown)


pdr_backend/lake/etl.py
```diff
@@ -73,25 +73,25 @@ def _drop_temp_sql_tables(self):
             If exists, drop them and rebuild
         """
         # drop the tables if it exists
-        pds = PersistentDataStore(self.ppss.lake_ss.lake_dir)
+        duckDB = DuckDBDataStore(self.ppss.lake_ss.lake_dir)
```
calina-c, line 76: I don't quite agree to naming this variable duckDB. It's very confusing and error-prone. The library import for DuckDB is duckdb. This variable name is duckDB. To keep the code readable, we need to have a clear distinction on what we are using: the duckdb library or a custom datastore object?

Also, variable names in Python should always be `snake_case`.

I suggest `duckdb_ds` or `duck_ds` or `duck_datastore` for previous instances of `pds`.

pdr_backend/lake/test/test_duckdb_data_store.py
```diff
@@ -150,95 +142,93 @@ def test_fill_table_from_csv(tmpdir):
 
 
 def test_multiton_instances(tmpdir):
-    persistent_data_store_1 = PersistentDataStore(str(tmpdir))
-    persistent_data_store_2 = PersistentDataStore(str(tmpdir))
+    duckdb_1 = DuckDBDataStore(str(tmpdir))
```
calina-c, line 145: same comment as above, and in other instances

calina-c (changes requested):

pdr_backend/lake/test/test_duckdb_data_store.py
idiom-bytes: I named things as db in many instances. I'll do that again.

pdr_backend/lake/etl.py
idiom-bytes: just like most other instances, i renamed it simply to "db"

PR #1095, opened 2024-05-28
Fix dynamic ppss start date causes issues

### LINKED ISSUE #1094: [Lake][ETL] Use raw tables timestamp for bronze tables start_timestamp calculation
Issue:
Because `lake_ss` start timestamp value could be dynamic by `now` which changes whenever it's used there is a gap between table start dates, resulting different number of rows between raw and etl tables

Proposed fix:
Make start timestamp constant between all tables by:
- at GQL lvl - get the start time from ppss when fetch process starts and pass the sam value to all the tables
- at ETL lvl - get start timestamp from raw tables as min of all raw tables timestamp

### Accepted Fx
image

### DoD:
- [x] The pipeline expects a fixed st_ts and end_ts throughout the run, the system should not be changed
- [x] To fix, override ppss with a fixed datetime such that the dates do not chang 
- [x] Figure out st_ts and end_ts before calling the pipeline. Then, call the pipeline with a fixed st_ts and end_ts, such that these parameters DO NOT CHANGE during the course of the ETL pipeline.

KatunaNorbert:
Fixes #1094.

### DoD:
- [ ] Figure out st_ts and end_ts before calling the pipeline. Then, call the pipeline with a fixed st_ts and end_ts, such that these parameters DO NOT CHANGE during the course of the ETL pipeline.
- [ ] Add checks/validations/asserts that (end-ts - st_ts) > 1 day. Anything below this, is a horrible configuration. If subgraph goes down for a day, this configuration fails.

### DIFF
--- a/pdr_backend/lake/etl.py
+++ b/pdr_backend/lake/etl.py
@@ -185,10 +185,74 @@ class ETL:
             values[table_name] = UnixTimeMs(max_timestamp) if max_timestamp else None
 
         return values
+    
+    @enforce_types
+    def _get_min_timestamp_values_from(
+        self, tables: List[NamedTable]
+    ) -> Dict[str, Optional[UnixTimeMs]]:
+        """
+        @description
+            Get the min timestamp values from the tables
+
+        @arguments
+            table_names - The list of table names to get the min timestamp values from
+            table_type - The type of table to get the min timestamp values from
+        @returns
+            values - The min timestamp values from the tables
+        """
+        min_timestamp_query = (
+            "SELECT '{}' as table_name, MIN(timestamp) as min_timestamp FROM {}"
+        )
+        print(tables)
+
+        all_db_tables = DuckDBDataStore(self.ppss.lake_ss.lake_dir).get_table_names()
+
+        queries = []
+
+        for table in tables:
+            if table.fullname not in all_db_tables:
+                logger.info(
+                    "_get_min_timestamp_values_from - Table %s does not exist.",
+                    table.fullname,
+                )
+                continue
+
+            queries.append(min_timestamp_query.format(table.fullname, table.fullname))
+
+        logger.info("_get_min_timestamp_values_from - queries: %s", queries)
+
+        table_names = [table.fullname for table in tables]
+        none_values: Dict[str, Optional[UnixTimeMs]] = {
+            table_name: None for table_name in table_names
+        }
+
+        if len(queries) == 0:
+            return none_values
+
+        final_query = " UNION ALL ".join(queries)
+        result = DuckDBDataStore(self.ppss.lake_ss.lake_dir).query_data(final_query)
+
+        logger.info("_get_min_timestamp_values_from - result: %s", result)
+
+        if result is None:
+            return none_values
+
+        values: Dict[str, Optional[UnixTimeMs]] = {}
+
+        timesamps = []
+
+        for row in result.rows(named=True):
+            table_name = row["table_name"]
+            min_timestamp = row["min_timestamp"]
+            if min_timestamp:
+                timesamps.append(UnixTimeMs(min_timestamp - 1))
+            values[table_name] = UnixTimeMs(min_timestamp) if min_timestamp else None
+
+        return min(timesamps)
 
     @enforce_types
     def get_timestamp_values(
-        self, table_names: List[str], default_timestr: str
+        self, table_names: List[str], default_timestr: UnixTimeMs
     ) -> UnixTimeMs:
         max_timestamp_values = self._get_max_timestamp_values_from(
             [NamedTable(tb) for tb in table_names]
@@ -200,7 +264,7 @@ class ETL:
         # check if all values are None in max_timestamp_values
         # and return the default_timestr if so
         if all(value is None for value in max_timestamp_values.values()):
-            return UnixTimeMs.from_timestr(default_timestr)
+            return default_timestr
 
         values = []
         if len(max_timestamp_values) > 0:
@@ -210,7 +274,7 @@ class ETL:
 
         logger.info("get_timestamp_values - values: %s", values)
         timestamp = (
-            max(values) if len(values) > 0 else UnixTimeMs.from_timestr(default_timestr)
+            max(values) if len(values) > 0 else default_timestr
         )
         return timestamp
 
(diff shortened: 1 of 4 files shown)


idiom-bytes:
Hi @KatunaNorbert, 

Rather than "fixing" the problem w/ a big system-level change and no-test...

**[Requirements]**
I want you to please create a test that shows me why/how the problem is happening.
Please use with/patch and the fixture from test_etl.py to try and isolate the issue.

**[Overview]**
In my mental model, **the functions we use in GQL and ETL to calculate st_ts and end_ts should work.**
@ gql_data_factory.py - LoC 291, 296
@ etl.py - LoC 213

Whether you use:
st_ts: "05-01-2024"
st_ts: "28d ago"

Should both resolve to the checkpoint/logic/tables processed. No rows should be dropped, no duplicate/extra rows should be created. "predictions" and "bronze_predictions" should have the same number of rows.

pdr_backend/lake/gql_data_factory.py
```diff
@@ -285,14 +281,16 @@ def _update(self):
...
         """
         fin_ut = self.ppss.lake_ss.fin_timestamp
+        start_ut_ppss = self.ppss.lake_ss.st_timestamp
 
         for table in (
             TableRegistry().get_tables(self.record_config["gql_tables"]).values()
         ):
             # calculate start and end timestamps
-            st_ut = self._calc_start_ut(table)
+            st_ut = self._calc_start_ut(table, start_ut_ppss)
```
idiom-bytes, line 293: i think this makes sense, but i don't understand how it fixes the problem you described in the ticket

further, you didn't provide any test or other setup showing me how this fixes the problem

pdr_backend/lake/etl.py
```diff
@@ -224,12 +286,14 @@ def _calc_bronze_start_end_ts(self) -> Tuple[UnixTimeMs, UnixTimeMs]:
             ETL updates should use to_timestamp by calculating
             min(max(source_tables_max_timestamp)).
         """
-        from_timestamp = self.get_timestamp_values(
-            self.bronze_table_names, self.ppss.lake_ss.st_timestr
+        st_timestr = self._get_min_timestamp_values_from(
+            [NamedTable(tb) for tb in self.gql_data_factory.raw_table_names]
```
idiom-bytes, line 290: if you get min_timestamp from raw_tables... you're going to get the very first event... which means you'll be re-processing all events historically. 

This is NOT what you want to do.

KatunaNorbert: First run:
Using the `ppss.lake_ss.st_ts` is the exact problem because it can either be fix or dynamic
- if the values is a fixed date than not problem, all tables are getting the same start date
- if value is dynamic, like 18d ago than the value is different whenever you use it and you will end up with different start dates.
Ex: gql start at 2024.05.19_11:00 -> st_date 2024.05.1_11:00
      etl start at  2024.05.19_11:10 -> st_date 2024.05.1_11:10
     -> 10min gap between the start dates
     
Second/third:
You got it wrong, the start date is read from ppss just for the first run when there are no rows in the bronze tables, in all the other cases it gets the start time from the bronze tables.
This was not changed, the only thing changes is that on the first run when there is no data in the bronze tables it gets the start date from raw tables instead of ppss, and the reason is explain above ( the ppss can be dynamic )

pdr_backend/lake/gql_data_factory.py
KatunaNorbert: It fixes the problem by having a constat start timestamp across all the tables while the fetch process is running instead of initiating it at each table level which causes different values.

I have not provided a test because It takes a lot of time and we should validate the approach before implementing the solution e-e and then just ditching it

KatunaNorbert:
Added test to show how ppss.lake_ss.st_timestr could be dynamic and have different values

idiom-bytes:
```Added test to show how ppss.lake_ss.st_timestr could be dynamic and have different values in different places we use it```
thank you, i see the issue now
"we need st_ts and end_ts to be consistent across the whole GQL + ETL execution"

[Example]
basically, because we do not maintain a checkpoint, time shifts as the pipeline runs 

when gql_starts
now = 1

when etl_starts
now = 2

when etl_ends
now = 3

[Zoom Out]
It's the same problem that was happening in ETL, but now across the whole pipeline.

**[Simple Solution]**
Just resolve `st_ts` and `end_ts` into a fixed timestamp than then gets passed in throughout the whole flow...

image

**Comment by KatunaNorbert** [comment]
> **[Simple Solution]** Just resolve `st_ts` and `end_ts` into a fixed timestamp than then gets passed in throughout the whole flow... example...

What if a run `lake raw update` and then `lake etl update`?

**Comment by idiom-bytes** [comment]
```
What if a run lake raw update and then lake etl update, or drop etl tables and run etl update again?
```
@KatunaNorbert @calina-c 

The pipeline interface expects you to give it a fixed `st_ts` + `end_ts` and the system was modified in such a way where these parameters can change... 

The problem isn't that the data pipeline doesn't support a moving checkpoint (it will to a certain extent and this will break too...) it's that things changed in such a way where ppss configuration can break the pipeline.

So, please, figure out st_ts and end_ts before calling the pipeline. Then, call the pipeline with a fixed st_ts and end_ts, such that these parameters DO NOT CHANGE during the course of the ETL pipeline.

When we implement a checkpoint system that isn't dependant on these things, we will no longer have any of these problems.

Overall: you're testing functionality that is edge case/bad configuration... **so we're going to limit what we're going to do here and the rest should be simply asserted/error thrown.**

**Comment by idiom-bytes** [comment]
WRT the core issue, I'm going to create a separate PR so we can try to close this off.

WRT error on CLI and the discussion below, I'll create a separate ticket.

```
For me it doesn't sound like an edge case
```
Can you please be more specific as to why you don't think this is an edge case/bad configuration?

I'm arguing that we should not support anything less than a day. It doesn't even make sense for what our system is trying to do  (predict the future). 

**Although the system will still work (because it will start from the last_record available > st_ts)**, trying to support anything below 1d lookback on ppss.yaml doesn't sound like a productive idea.

```
ppss.lake_ss.st_ts = "1d ago"
ppss.lake_ss.end_ts = "now"
```

@trentmc can you please provide some guidance?

idiom-bytes:
I have now fixed it here...
https://github.com/oceanprotocol/pdr-backend/pull/1106
image

And have crated this ticket for the future, when I implement a looping update.
https://github.com/oceanprotocol/pdr-backend/issues/1107

Closing this PR

PR #1097, opened 2024-05-29
Remove more redundant code from ETL old versions

calina-c:


pdr_backend/lake/etl.py
```diff
@@ -86,7 +89,7 @@ def do_etl(self):
             logger.info("do_etl - Drop build tables.")
 
             # Sync data
-            self.gql_data_factory.get_gql_tables()
+            self.gql_data_factory._update()
```
idiom-bytes, line 92: so good to finally see this fixed...

idiom-bytes (approved):
much cleaner, great!

PR #1098, opened 2024-05-29
More ETL removals, including Table

calina-c:
Contains #1097, but more radical. This PR also removes Table, which has become redundant. Without a table registry, we have no use with Table objects.

Table = NamedTable that contains type + PPSS configuration

PPSS configuration comes always from the data factory, so we can inject it more freely without creating connective objects such as Table.

### DIFF
--- a/pdr_backend/analytics/test/test_get_predictions_info.py
+++ b/pdr_backend/analytics/test/test_get_predictions_info.py
@@ -6,7 +6,7 @@ from enforce_typing import enforce_types
 
 from pdr_backend.analytics.get_predictions_info import get_predictions_info_main
 from pdr_backend.lake.prediction import Prediction
-from pdr_backend.lake.table import Table
+from pdr_backend.lake.table import NamedTable
 from pdr_backend.ppss.ppss import mock_ppss
 
 
@@ -31,8 +31,8 @@ def test_get_predictions_info_main_mainnet(
         fin_timestr=fin_timestr,
     )
     predictions_df = _gql_datafactory_first_predictions_df
-    predictions_table = Table(Prediction, ppss)
-    predictions_table.append_to_storage(predictions_df)
+    predictions_table = NamedTable.from_dataclass(Prediction)
+    predictions_table.append_to_storage(predictions_df, ppss)
 
     feed_addr = "0x2d8e2267779d27c2b3ed5408408ff15d9f3a3152"
 
@@ -96,8 +96,8 @@ def test_get_predictions_info_bad_date_range(
     )
 
     predictions_df = _gql_datafactory_first_predictions_df
-    predictions_table = Table(Prediction, ppss)
-    predictions_table.append_to_storage(predictions_df)
+    predictions_table = NamedTable.from_dataclass(Prediction)
+    predictions_table.append_to_storage(predictions_df, ppss)
 
     feed_addr = "0x2d8e2267779d27c2b3ed5408408ff15d9f3a3152"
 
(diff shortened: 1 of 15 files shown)


pdr_backend/analytics/test/test_get_predictions_info.py
```diff
@@ -31,8 +31,8 @@ def test_get_predictions_info_main_mainnet(
         fin_timestr=fin_timestr,
     )
     predictions_df = _gql_datafactory_first_predictions_df
-    predictions_table = Table(Prediction, ppss)
-    predictions_table.append_to_storage(predictions_df)
+    predictions_table = NamedTable.from_dataclass(Prediction)
```
idiom-bytes, line 34: yeah, this is already so much better...

pdr_backend/lake/test/conftest.py
```diff
@@ -601,16 +601,16 @@ def setup_data(
...
     }
 
-    gql_tables["pdr_predictions"].append_to_storage(preds)
-    gql_tables["pdr_truevals"].append_to_storage(truevals)
-    gql_tables["pdr_payouts"].append_to_storage(payouts)
-    gql_tables["pdr_slots"].append_to_storage(slots)
+    gql_tables["pdr_predictions"].append_to_storage(preds, ppss)
+    gql_tables["pdr_truevals"].append_to_storage(truevals, ppss)
+    gql_tables["pdr_payouts"].append_to_storage(payouts, ppss)
+    gql_tables["pdr_slots"].append_to_storage(slots, ppss)
```
idiom-bytes, line 613: I'm seeing ppss being passed around so much.
It's growing too.
And it's a configuration file.

Isn't there a better way?

pdr_backend/lake/csv_data_store.py
```diff
@@ -123,8 +123,8 @@ def __init__(self, base_path: str, table_name: str):
         self.table_name = table_name
 
     @staticmethod
-    def from_table(table):
-        return CSVDataStore(table.base_path, table.table_name)
+    def from_table(table, ppss):
+        return CSVDataStore(ppss.lake_ss.lake_dir, table.table_name)
```
idiom-bytes, line 127: I'm seeing ppss being passed around so much.
It's growing too.
And it's a configuration file.

Isn't there a better way?

pdr_backend/lake/gql_data_factory.py
```diff
@@ -113,11 +77,14 @@ def _prepare_temp_table(self, dataclass: Type[LakeMapper], st_ut, fin_ut):
             # 3. in preparation to append, check missing data to move FROM CSV -> TO TEMP TABLES
             # 4. resume appending to CSVs + Temp tables until complete
         """
-        table = Table(dataclass, self.ppss)
+        table = NamedTable.from_dataclass(dataclass)
+        temp_table = TempTable.from_dataclass(dataclass)
```
idiom-bytes, line 81: :+1:

pdr_backend/lake/gql_data_factory.py
```diff
@@ -256,7 +225,7 @@ def _do_subgraph_fetch(
             pagination_offset += pagination_limit
 
         if len(buffer_df) > 0:
-            table.append_to_storage(buffer_df, TableType.TEMP)
+            temp_table.append_to_storage(buffer_df, self.ppss)
```
idiom-bytes, line 228: so much ppss being passed around...

idiom-bytes (approved):
I think this is overall a great improvement, however I'm seeing ppss being passed around so much and I wonder if there is a better pattern.... like a decorator or something else, such that the ppss is passed around by default, and can just be accessed wherever needed...

I have created a ticket to address this later.
https://github.com/oceanprotocol/pdr-backend/issues/1108

calina-c:
> I think this is overall a great improvement, however I'm seeing ppss being passed around so much and I wonder if there is a better pattern.... like a decorator or something else, such that the ppss is passed around by default, and can just be accessed wherever needed...
> 
> I have created a ticket to address this later. #1108

I agree. I will address this later.

PR #1100, opened 2024-05-29
[ETL] Incremental ETL Pipeline

### LINKED ISSUE #1001: [ETL] Incremental Pipeline
### Motivation

The ETL needs to be able to fetch 5 mins of data, every 5 minutes, in such a way that each fetch updates historical records allowing us to understand when previous predictions became correct, calculating revenue, accuracy, and many other metrics.

### Update Step - Incrementally updating the Lake
When you run the "lake update"
1. New records will arrive that need to be processed
1. These new records (such as pdr_payout) if applicable should be: (a) cleaned up into their raw/bronze table, (b) update historical records to reflect this event arriving such as a prediction that is now paid out.
1. After all records have been yielded to temp tables and the pipeline ends, records are then swapped to prod tables.

Screenshot from 2024-05-06 13-27-25

Data Workflows
- All data workflows will take place as a series of SQL queries, and different tables used to execute the incremental/swap strategy.
- All data workflows will have tests and coverage demonstrating it working.

### DoD:
- [x] Prediction and payouts are being processed incrementally
- [x] Bronze prediction is being updated incrementally
- [x] Other tables and bronze tables are currently not processed

Task:
- [x] Process new pdr-predictions into duckdb
- [x] Process new pdr-payouts into duckdb
- [x] Verify the incremental update step works [#982](https://github.com/oceanprotocol/pdr-backend/issues/982)
- [x] Create SQL that process new pdr-predictions into new_events_bronze_predictions_table
- [x] Create SQL that process new pdr-payouts into update_events_bronze_predictions_table
- [x] Create SQL & merge strategy that processes new_events and update_events into incremental bronze_tables
- [x] Verify that null records inside bronze-predictions are being updated correctly
- [x] Verify everything is working e2e

idiom-bytes:
Fixes #1001 

@kdetry ... after our conversation I took the time to organize this draft such that you have a good foundation to take it in the right direction. I'm concerned that although "your solution seems simpler" (to copy the table and then do onEventHandlers), it will lead to a less efficient and burdensome pipeline...

Like I have mentioned, my goal is to lean further into DuckDB/SQL such that all logic is done in SQL, through more atomic operations, such where everything is only processed once.

I have provided 2 examples in here.

**[Example 1 - Required ETL/SQL]**
In here, I have simplified things such that we only need predictions and payouts.

Trueval events do not contain the UserID, so it will not do anything useful until the slot table arrives.

This is the solution I have provided the following DoD, although it's a pseudo implementation.
The overall architecture/approach however remains the same.

```
pdr_backend/lake/required_etl.py
pdr_backend/lake/required_sql_bronze_predictions.py
pdr_backend/lake/required_sql_payouts.py
pdr_backend/lake/required_sql_predictions.py
```

Changes proposed in this pseudo-code:
- [x] 1 event processed can now generate multiple rows on multiple tables  
- [x] _temp tables hold new events
- [x] _update tables hold update events
- [x] _update_raw tables is designed to hold N update events per ID
- [x] because we're not analyzing update events, reduce multiple update events into 1
1. update logic generates 5 prediction events for slot ID
2. reduce step groups by ID, and turns 5 rows into a single (1) update record, before joining with millions
- [x] merge -> drop -> merge records in such a way that updates production tables

**[Example 2 - Broader ETL/SQL]**

The other SQL queries, were the ones I originally wrote and I have now paused.
```
pdr_backend/lake/sql_etl_bronze_predictions.py
pdr_backend/lake/sql_etl_payouts.py
pdr_backend/lake/sql_etl_predictions.py
pdr_backend/lake/sql_etl_truevals.py
```

These are much more extensive, and show you how the system grows to also include updates to the slots table, and much more, **without causing additional reading/copying/processing.**

I have done this to pseudo code/show/provide direction, as to how this approach continues to scale and work with the same implementation as required_etl.py

**This is not required for now... only required_*.py.**

### After this PR completes - Continue furthering: 
1. Renaming, and reducing table types & operations => Reduce everything to "EventsTable" Temp & Prod
TempTable -> becomes NewTable (INSERT)
UpdateTable -> becomes UpdateTable (UPDATE)

NewTable + UpdateTable -> become EventsTable (INSERT + UPDATE Records - 1 per row)
-> These can then be collapsed/reduced with a group operation.
-> These can then be merged 1:1 w/ the prod table
-> In the future, we may want to store the EventsTable for faster development workflows around the Lake/ETL

Perhaps introduce:
-> eventName (predictoorSlot, predictoorPrediction, predictoorTrueval)
-> eventType (new vs. update)

2. Most of my queries need to use `table.fullname` rather than `table.table_name` and it's not intuitive. I.E. The most obvious interface name should likely be the one I use the most. This also matches up to the base/class oop nature of the code base. 
- [x] Rename NamedTable/Table.table_name -> base_table_name.
- [x] Rename NamedTable/Table.table_name -> table_name

### DoD:
- [x] Re-write the queries such that we're processing each contract/subgraph event only once (subscribe, slot, prediction, trueval, payout) into New + Update events.
- [x] Re-write queries such that every Contract Event, yields the right New + Update records
- [x] Re-write ETL such that New + Update records are going into their respective tables (TempTable + UpdateTable)
- [x] Re-write ETL such that Update records are reduced into a single row (I.E. N SlotUpdate records -> 1 SlotUpdate record) before being joined historically
- [x] Re-write ETL such that only 1:1 Update:Prod record is executed, reducing historical join complexity.
- [x] Re-write ETL such that as new events show up, existing records are updated.
- [x] Clean up logs and output so we can follow what's happening from inside the CLI. Right now there are a lot of poor logging (like every query being sent to subgraph) that is just spam and adds no value since things have evolved quite a bit.
- [x] Disabled prediction subgraph data from being used in the ETL pipeline, such that enforce that the ETL is correctly enriching historical data as we would expect
- [X] Started to abstract conftest and  test functionality to more easily build a multi-stage test
- [x] Export large set of pdr_predictions and pdr_payouts
- [x] Setup mock/fixture so we can easily retrieve larger data samples for processing
- [x] Create test_first_step_workflow() to verify that the `first work` updates with itself, doesn't look for historical records, and all incoming null payouts/values w/ matching payout events, are updated successfully
- [x] Implement a ppss clamp in ETL so checkpoint calculations are clamped to ppss.st_timestamp, ppss.fin_timestamp
- [x] Create test_incremental_workflow() to verify the fisrt-batch updates successfully, and that multi-stage scenarios are working
- [x] Debug and fix remaining issues with incremental pipeline
- [x] Fix start/end conditional checks that are causing 1-2 rows to not be populated/processed correctly
- [x] Rewrite swap strategy so that each table is processed atomically
- [x] Swap old ETL for incremental ETL and remove any old functionality that is no longer used
- [x] make sure everything is wired up to the new system
- [x] start running the pipeline via cli and make sure nothing breaks...
- [x] fix duplicate issue where by simply incrementing the pipeline, we would see a duplicate record
- [x] Cleanup table types, table object, and interface
- [x] Drop any tests that no longer apply
- [x] Update ETL tests that are valid to use new fixture
- [x] fix duplicate issue where you are incrementing the pipeline, and then sett a new fin_ts < last_ts of the last record and a bunch of duplicates are created.
- [x] add lake validation for checking if all valid payouts from raw that should be in bronze_predictions, are in bronze_predictions

### Prepare Mege
- [x] merge latest from main
- [ ] review code
- [ ] simplify logic, address DRY, and cleanup code wherever appropriate before merging
- [ ] merge PR


idiom-bytes:
Configured ETL for hard-mode such that there is no data coming from subgraph and we need to rebuild all state through ETL
- Some of the predictions should remain null (because we don't have those payouts yet)
- Some of the payouts cannot find a prediction (because it happened in the past, so it can't be joined)
image

Cleaned CLI output to not spam (i.e. we don't need to log every query, and every result)... It's now compact, easier to read, and I'm able to understand full e2e ETL.
image

Generated CSV files w/ extended sample data so we can more easily configure a thorough ETL test, that includes incremental validations. This test data is configured to be as empty as possible, such that we can verify that all state is being built correctly on-the-fly.
image

Starting to built out an incremental test w/ a much larger dataset, such that i can more easily replicate and resolve issues. As an example my current configuration has 2058 predictions, but I only see 2057 in my bronze_predictions table.
image

idiom-bytes:
Updates regarding ETL sample data and coverage:
- New ETL fixture `_sample_etl`
- Sample data and fixture has roughly 2000 predictions and payouts ~= 1mb
- datetime ranges (from,to) `("2024-05-05_00:00", "2024-05-05_02:00")`
- e2e tests are now setup and running

**Test 1 => 1 ETL step**
Step 1 - ("2024-05-05_00:00", "2024-05-05_02:00")

**Test 2 => 3 ETL steps**
Step 1 - ("2024-05-05_00:00", "2024-05-05_00:40")
Step 2 - ("2024-05-05_00:40", "2024-05-05_01:20")
Step 3 - ("2024-05-05_01:20", "2024-05-05_02:00")

Next:
- I'm now debugging Step 1 -> Step 2 and fixing issues in the incremental step

pdr_backend/lake/test/test_required_etl.py
```diff
@@ -0,0 +1,289 @@
...
+    assert prod_valid_payouts == 1678
+    assert prod_null_payouts + prod_valid_payouts == 2055
+
+
+# pylint: disable=too-many-statements
+@enforce_types
+@pytest.mark.parametrize(
+    "_sample_etl", [("2024-05-05_00:00", "2024-05-05_00:40")], indirect=True
+)
+def test_etl_do_incremental_bronze_step(_sample_etl):
```
idiom-bytes, line 104: test_etl_do_incremental_bronze_step is now showing all work, and completing correctly e2e
1Step-All Records run == same rows as == 3Step-incremental run

idiom-bytes:
I have now completed the 2 final features that were required and the latest is pushed to the PR
- [x] Fix start/end conditional checks that are causing 1-2 rows to not be populated/processed correctly
- [x] Rewrite swap strategy so that each table is processed atomically

I'm now going to:
- [ ] start removing old functionality
- [ ] make sure everything is wired up to the new system
- [ ] start running the pipeline via cli and make sure nothing breaks

KatunaNorbert:
There are a lot of changes in this PR and a lot of them are outside the scope and makes reviewing very hard to follow.
Could you only pick the changes related to this PR?

pdr_backend/cli/cli_arguments.py
```diff
@@ -39,14 +43,21 @@
...
 """
 
 HELP_OTHER_TOOLS = """
 Power tools:
   pdr multisim PPSS_FILE
+  pdr arima_plots PPSS_FILE [--debug_mode False]
   pdr deployer (for >1 predictoor bots)
-  pdr lake PPSS_FILE NETWORK
+  pdr lake raw|etl update PPSS_FILE NETWORK
+  pdr lake raw|etl drop PPSS_FILE NETWORK ST
```
KatunaNorbert, line 58: lake commands are duplicate, they are both mentioned in `HELP_OTHER_TOOLS` and `Inspect and manage lake`

KatunaNorbert:
ETL completes successfully and results in the describe command are looking good, but testing or inspecting the updates on live tables it's currently difficult, maybe something could be added to the describe command to show the number or rows containing null values or something else that could make inspecting easier.
Another improvement to validate updates could be checking if there is a predictions with payout but still containing null values. In other words all predictions that have a payouts associated should not have null values

KatunaNorbert:
I'm getting some issues here. This is how I'm testing:
- running etl update to get the latest events
- get latest payout event id and check if predictions with the same id has null values
- i expect predictions that have payout event to not have null values, but they do both in db and CSVs

idiom-bytes:
```
but testing or inspecting the updates on live tables it's currently difficult, maybe something could be added to the describe command to show the number or rows containing null values or something else that could make inspecting easier.
```

```
Another improvement to validate updates could be checking if there is a predictions with payout but still containing null values. In other words all predictions that have a payouts associated should not have null values
```

these two comments, and

```
i expect predictions that have payout event to not have null values, but they do both in db and CSVs
Looks to mainly appear on 1h feeds
```
led me and norbert to talk, i have added a validation that makes sure that from all predictions that are there, all valid matches have been updated and have payout != null.

**although these vaidations might be redundant with tests, they are very useful operationally** this is something that I have to keep telling myself.

pdr_backend/cli/cli_arguments.py
idiom-bytes: I don't see a problem with this. I don't see why it should be exclusive... but if it is, then i'll move more stuff from "power tools" to "lake"

KatunaNorbert:
>i have added a validation that makes sure that from all predictions that are there, all valid matches have been updated and have payout != null.

Was this change pushed? I only see validation on each payout having a matching prediction, bot not on predictions being updated

idiom-bytes:
"I only see validation on each payout having a matching prediction, bot not on predictions being updated"

@KatunaNorbert 
- The tests provide deep coverage over thousands of records showing that predictions are updated
- The validations show that all payouts that came in, have been matched up with bronze predictions and there are no NULL values. This means that they have been updated correctly.

READMEs/lake-and-etl.md
```diff
@@ -0,0 +1,224 @@
...
+- [**Data Warehouse - "Database"**](#data-warehouse---database)
+- [**ETL**](#etl)
+- [**Checkpoints**](#checkpoints)
+- [**Dumb Components**](#dumb-components)
+- [**Examples**](#examples)
+- [**Dos and Don'ts**](#dos-and-donts)
+
+
+## ETL Tables & Architecture
+To provide summaries on Predidctoor users and systems, we want to fetch all new data and make it available for analysis.
```
KatunaNorbert, line 28: a small typo here **Predidctoor** -> **Predictoor**

READMEs/lake-and-etl.md
```diff
@@ -0,0 +1,224 @@
...
+It contains many folders and CSVs, that maintain a complete set of records and reliable information. It should be easy to understand and to work with.
+
+Currently, the lake (CSVs) can be accessed via the CSVDataStore (CSV wrapper) and operated with via the CLI module `lake raw` commands `pdr lake raw update ppss.yaml sapphire-mainnet` (cli_module_lake.py)
+
+Some features include:
+1. The lake has a few commands: `describe`, `validate`, `update`, `drop`, and `query`
+1. The lake is made up of many timeseries objects that should exist from `lake_ss.st_ts` and `lake_ss.end_ts`.
+1. All the data inside lake folders should exist between `st_ts` and `end_ts`.
+1. The `st_ts` and `end_ts` checkpoints come from `ppss.yaml` and the `lake_ss` component.
+1. You can set `st_ts` and `end_ts` to relative times ("3 hours ago" -> "now"), so new new records are always fetched and processed.
```
KatunaNorbert, line 51: **new new records** -> **new records**

pdr_backend/lake/table_pdr_payouts.py
```diff
@@ -0,0 +1,22 @@
...
+    "ID": Utf8,
+    "token": Utf8,
+    "user": Utf8,
+    "slot": Int64,
+    "timestamp": Int64,
+    "payout": Float64,
+    "predictedValue": Boolean,
+    "revenue": Float64,
+    "roundSumStakesUp": Float64,
+    "roundSumStakes": Float64,
```
KatunaNorbert, line 20: The roundSumStakes values are not shown inside the diagram, should they be removed? Also saw them in the subgraph payout query

idiom-bytes:
```
The roundSumStakes values are not shown inside the diagram, should they be removed? Also saw them in the subgraph payout query
```
I have not changed the fetch or the schema because I did not want to break anything that's currently dependent on it

"The Diagram" is an up-to-date reflection of what the ETL generates.

KatunaNorbert: Isn't the payouts table part of the Lake/ETL?

calina-c:
very difficult to review for me as well. Can we cherry-pick? It currently shows 440 files changed, which is because of licenseheader updates in main.

idiom-bytes:
@calina-c @KatunaNorbert you can change the selected commit from the dropdown when reviewing the PR, such that you're only looking at the changes I made

image

if cherry-picking is the way you want to move forward, then so be it

PR #1106, opened 2024-05-30
[Duckdb] Fix Issue1094: Resolve "natural language" st_ts and end_ts in ppss to fixed values, as expected by lake/etl systems

### LINKED ISSUE #1094: [Lake][ETL] Use raw tables timestamp for bronze tables start_timestamp calculation

idiom-bytes:
Fixes #1094 

Changes proposed in this PR:
- [x] Resolve "natural language" date to a "fixed value" date such that the lake commands can run as expected
- [x] Lake commands expect to run st_ts and end_ts all the way through.
- [x] We'll have to eventually update the interface when everything is running in a single thread/loop, but we're not there yet, it's all one-shot through the CLI.

### DIFF
--- a/pdr_backend/cli/cli_module_lake.py
+++ b/pdr_backend/cli/cli_module_lake.py
@@ -10,6 +10,7 @@ from pdr_backend.lake.lake_validate import LakeValidate
 from pdr_backend.lake.duckdb_data_store import DuckDBDataStore
 from pdr_backend.lake.table import drop_tables_from_st
 from pdr_backend.ppss.ppss import PPSS
+from pdr_backend.util.time_types import UnixTimeMs
 
 logger = logging.getLogger("cli")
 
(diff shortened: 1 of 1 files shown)


KatunaNorbert (approved):
Strictly only testing `lake etl update` it's working

idiom-bytes:
I have continued testing/playing with this, testing different commands, etc... i'm closing this as part of sprint planning.

PR #1111, opened 2024-05-31
Issue #1108: Cleanup PPSS handling in ETL

### LINKED ISSUE #1108: [Lake][DuckDB] Clean ppss handling, such that it's not so verbose
### Background / motivation

In lake, ppss is being passed around like crazy although it's a configuration file...
https://github.com/oceanprotocol/pdr-backend/pull/1098

isn't there a better way to do all of this?

Approach A: Singleton
Approach B: Decorator
Approach C: Some other pattern...

### TODOs / DoD

Tasks:
- [ ] Clean up ppss as a pattern, such that it's accessed when needed, and doesn't have to be passed around so much

### Related github issues
https://github.com/oceanprotocol/pdr-backend/pull/1098

calina-c:
Closes #1108.

- automatic retrival of the caller ppss. In theory, tables should not be instantiated outside a class that has no ppss. So the solution I proposed is to add frame inspection and loop until we get the caller that has a ppss object. This will usually be GQLDataFactory, but we support any object that has a ppss attribute as a table intantiator.
- the only problem was with tests: in tests, tables are "orphan", so I'm retrieving the ppss object defined in the test `locals`.

I think a similar retrieval can be done on other ppss-dependant objects as well (CSVDataStore, DuckDBDataStore all originate from the same place so they are likely candidates), but here it makes most sense. Let me know if my solution is good to go :)

### DIFF
--- a/pdr_backend/lake/table.py
+++ b/pdr_backend/lake/table.py
@@ -1,3 +1,4 @@
+import inspect
 import logging
 from enum import Enum
 from typing import Optional, Type, Union
@@ -6,8 +7,9 @@ import polars as pl
 from enforce_typing import enforce_types
 
 from pdr_backend.lake.csv_data_store import CSVDataStore
-from pdr_backend.lake.lake_mapper import LakeMapper
 from pdr_backend.lake.duckdb_data_store import DuckDBDataStore
+from pdr_backend.lake.lake_mapper import LakeMapper
+from pdr_backend.ppss.ppss import PPSS
 from pdr_backend.util.time_types import UnixTimeMs
 
 logger = logging.getLogger("table")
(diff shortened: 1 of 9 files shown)


pdr_backend/lake/table.py
```diff
@@ -73,12 +75,46 @@ def drop_tables_from_st(db: DuckDBDataStore, type_filter: str, st: UnixTimeMs):
     logger.info("truncated %s rows from %s tables", trunc_count, table_count)
 
 
+def get_caller_ppss():
+    caller_ppss = None
+    prev_frame = inspect.currentframe()
```
idiom-bytes, line 80: Hi calina, 

I do not think it's best practice to be going through the call stack in this way to retrieve data.... I have never seen this implemented anywhere for functional/engineering purposes.

**[Separating Concerns]**
The lake cli commands expect "fixed date values" rather than "fuzzy, interpretable dates"... the following implementation to execute cli commands addresses this exactly as required, without adding any additional requirements or complexity.
https://github.com/oceanprotocol/pdr-backend/pull/1106/files

**[No additional systems needed]**
I do not believe this warrants any additional overhead/complexity by:
- using a build process for ppss
- using the callstack to retrieve previous ppss data
- using some sort of push/pop system to manage ppss
- using any sort of more advanced ppss implementation

thank you

idiom-bytes (changes requested):
Please separate ppss/stack feature into it's own ticket/discussion/PR so we can merge the rest. Thank you!

calina-c:
> Please separate ppss/stack feature into it's own ticket/discussion/PR so we can merge the rest. Thank you!

There is no... "rest". This PR removes passing around the ppss file in favour of the Table object automatically retrieving it from its parent.

calina-c:
Superseded by #1318, where it is rebased with main

PR #1114, opened 2024-05-31
Adds PPSS object retrieval in cli module.

calina-c:
Small enhancement to the cli module, where PPSS file is automatically added to the parser list if PPSS_FILE and NETWORK are present.


calina-c:
@trizin I asked for your review, specifically, since you've worked on the arg parsers

PR #1130, opened 2024-06-05
Fix #1128: [Sim] Improve trading: don't do sell-then-rebuy. How: track position

### LINKED ISSUE #1128: [Sim] Improve trading: don't do sell-then-rebuy. How: track position
## Basic Idea

Improve trading in sim engine.
- Same basic algorithm as before: if predict up then buy, and sell 5 min later; if predict down, then sell, and buy 5 min later. 
- But remove the unneeded buys & sells between the candles: 
  - If the previous candle had "buy" and the next candle has a "buy"
  - Then _don't do the sell-then-re-buy in between_. 
  - Similar for if "sell".
  - How: track "position", and whether it is short or long, then exit the position as needed

## Thorough Description

Signals: `pred_up`, `pred_down`, `nothing (not confident)`

### On "pred_up" Signal:
- **Action:** Buy and hold tokens
- **On signal change:** If the signal changes, sell the same amount of tokens that were initially bought
- **Profit Calculation:** Compare the amount received from the sale to the amount spent on buying the tokens

### On "pred_down" Signal:
- **Action:** Sell tokens
- **On signal change:** If the signal changes, buy back the same amount of tokens initially sold
- **Profit Calculation:** Compare the amount received from the initial sale to the amount spent on buying back the tokens

trizin:
Fixes #1128


pdr_backend/sim/sim_engine.py
```diff
@@ -57,7 +57,9 @@ def __init__(
         self.exchange = exchange_mgr.exchange(
             "mock" if mock else ppss.predictoor_ss.exchange_str,
         )
-
+        self.position_open = ""  # long, short, ""
+        self.position_size = 0  # amount of tokens in position
+        self.position_worth = 0  # amount of USD in position
```
trentmc, line 62: Instead of "position_open" and "position_size", why not just have `position_size` and allow it to be negative? Maybe better, up to you.

pdr_backend/sim/sim_engine.py
```diff
@@ -57,7 +57,9 @@ def __init__(
         self.exchange = exchange_mgr.exchange(
             "mock" if mock else ppss.predictoor_ss.exchange_str,
         )
-
+        self.position_open = ""  # long, short, ""
```
trentmc, line 60: (Putting here arbitrarily: please git merge main, to avoid discrepancy with main branch. Just a few changes but it threw me off.)

pdr_backend/sim/sim_engine.py
```diff
@@ -171,16 +173,38 @@ def run_one_iter(self, test_i: int, mergedohlcv_df: pl.DataFrame):
...
+            st.trader_profits_USD.append(0)
+        elif pred_down and self.position_open == "":
+            # Open short position if pred down and no position open
+            tokcoin_amt_send = trade_amt * conf_down / curprice
+            usd_received = self._sell(curprice, tokcoin_amt_send)
+            self.position_open = "short"
+            self.position_worth = usd_received
+            self.position_size = tokcoin_amt_send
+            st.trader_profits_USD.append(0)
+        elif self.position_open == "long" and not pred_up:
```
trentmc, line 192: The large # if statements is hard to track. A couple suggestions:
- group by if position_open = "long" vs "short". Could be nested within, though better is one case vs the other.
- better than "not pred_down" and "not pred_up": perhaps you have three states: "pred_down", "pred_nomove", "pred_up"

trentmc (changes requested):

trentmc:
One more feedback: it should exit the position if time > threshold. (Triple barrier method)

pdr_backend/sim/sim_engine.py
trentmc: Good Q. Try both, see what comes out cleanest. The overall if/then/else code block should look clean.

---
A related idea: maybe the if/then/else can go into its own method? (If not too many args)

trentmc (changes requested):
Could you merge with main please 

And ensure linters are happy

ppss.yaml
```diff
@@ -50,7 +50,7 @@ trader_ss:
     confidence_threshold: 0.02 # skip trade if confidence < this
     init_holdings:
       - 100000 USDT
-      - 2 BTC
+      - 100000 BTC
```
trentmc, line 53: Why 10000 BTC? that's a crazy huge amount

trentmc:
The trading logic before was logical and super simple. While the logic now is logical, it's slightly more complex. 

If we're going to have any increase in complexity, the new approach should make $ better.

Does it? 

### First cut benchmark

I ran a quick benchmark on BTC/USDT 5m, max_n_train=1000, autoregressive_n=2, ClassifLinearRidge. Full settings: [my_ppss.yaml](https://github.com/user-attachments/files/15809416/my_ppss.yaml.txt)

Results:
- Trader profit in main branch: $44.05
- Trader profit in this PR: $26.94
- Convergence plots below

### Main branch: trader plot vs time
main

### This PR: trader profit vs time
pr1130

Of course, this is just one benchmark problem, on one particular configuration. It's possible that this PR is better on average than "main branch".

### Proposed next steps

My suggestion: let's thoroughly benchmark this approach, and compare to main. If it does better (to justify increased complexity), we merge.

trentmc (approved):
Can merge once bug fix in main is done

PR #1149, opened 2024-06-07
Fix sim plots tab switches to initial state when new data is loaded

### LINKED ISSUE #1146: [Bug, sim_plots] The display annoyingly goes back to first tab
### Background / motivation

Let's say that I've done a simulation run. And I have the plot open for it in the browser at http://127.0.0.1:8050/

If I'm on the first tab ("Profit"), it displays the profit plots, good.

But if I'm on any other tab (eg "Model Performance") then within a few seconds, the display will go back to the "Profit tab". Even if I don't do anything. This is super super annoying. (It says "updating..." then makes the annoying switch.)

Or, let's say that I'm running a simulation live, and have clicked on "Model Performance" tab. Then new plot data comes in, it says "updating", and makes the annoying switch.

### TODOs

- [x] Make it so that it does not switch tabs unless the _user_ has switched tabs. "Updating" should definitely not switch tabs

KatunaNorbert:
Fixes #1146.

### DIFF
--- a/pdr_backend/sim/dash_plots/callbacks.py
+++ b/pdr_backend/sim/dash_plots/callbacks.py
@@ -60,9 +60,10 @@ def get_callbacks(app):
         Input("interval-component", "n_intervals"),
         Input("selected_vars", "value"),
         State("selected_vars", "value"),
+        State("selected-tab", "data"),
     )
     # pylint: disable=unused-argument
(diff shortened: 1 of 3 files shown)


trentmc (approved):
Approved.

I also updated from main.

PR #1189, opened 2024-06-11
Fix #1188 - sim engine record total profits

### LINKED ISSUE #1188: [Multisim] Multisim csv records profits from last epoch, rather than total across all epochs

trizin:
Fixes #1188

### DIFF
--- a/pdr_backend/predictoor/predictoor_agent.py
+++ b/pdr_backend/predictoor/predictoor_agent.py
@@ -421,7 +421,7 @@ class PredictoorAgent:
             self.iter_number % self.ppss.predictoor_ss.aimodel_ss.train_every_n_epochs
         ) == 0:
             model_f = AimodelFactory(self.ppss.predictoor_ss.aimodel_ss)
-            model = model_f.build(X, ybool)
+            model = model_f.build(X, ybool, ycont, y_thr)
         else:
             assert self.crt_trained_model is not None
             model = self.crt_trained_model


pdr_backend/sim/sim_state.py
```diff
@@ -81,6 +81,23 @@ def recent_metrics_names() -> List[str]:
...
+                "pdr_profit_OCEAN": sum(self.pdr_profits_OCEAN),
+                "trader_profit_USD": sum(self.trader_profits_USD),
+            }
+        )
+
+        if extras and "prob_up" in extras:
+            rm["prob_up"] = sum(self.probs_up) / len(self.probs_up)
+
+        return rm
+
```
trentmc, line 100: Don't add this (`SimState.sim_metrics()`), because it confuses the pattern of what "recent_metrics()" means; everything is a "sim_metric".

Instead: in multisim_engine.py, where it calculates run_metrics, it needs to compute a "final" aggregate or summary value for each of the metrics. This includes cumulative values for predictoor profit & trader profit. But also everything else. In some cases, it's already aggregate ("acc_est", "acc_l", "acc_u"). In the other cases you need to compute an average. Here's a rough cut:
```python
recent_metrics = sim_engine.st.recent_metrics()
run_metrics = {
    "acc_est": recent_metrics["acc_est"],
    "acc_l": recent_metrics["acc_l"],
    "acc_u": recent_metrics["acc_u"],
    "f1": np.mean(st.aim.f1s),
    "precision": np.mean(st.aim.precisions[1:]), # avoid 1st sample, it may be off
    "recall": np.mean(st.aim.recalls[1:]), # ""
    "loss": np.mean(st.aim.losses[1:]), # ""
   "yerr": np.mean(st.aim.yerrs[1:]),# ""
    "pdr_profit_OCEAN" : np.sum(st.pdr_profits_OCEAN),
    "trader_profit_USD": np.sum(st.trader_profits_USD),
}
```

calina-c: AFAIK, we already have a slicing function, that returns the recent metrics.
If appropriate, we can have an aggregation function to calculate significant values from historical data. I do think these functions should be separated, though. I.e. `recent_metrics` and `aggregated_metrics` should not be common.

trentmc: @calina-c SimState.recent_metrics() is what you refer to. Some of its metrics are an aggregate that we can use for multisim, some aren't. That's why I propose NOT to create any new function, but rather to compute as needed in multisim.

PR #1190, opened 2024-06-11
Fix #1180: [ARIMA feed plots] Display ADF graph as table

### LINKED ISSUE #1180: [Bug, ARIMA feed plots] Left-side ADF bar plot looks wrong (8 bars for 4 values). To fix, convert to table
### TODO

Recommended solution:

Remove the bar plots. Rather, simply have a table of ADF values, like below. Then make each row selectable.

| Transform | ADF |
| ---- | ---- |
| BC=F, diff=0 | 0.951 |
| BC=T, diff=0 | 0.893 |
| BC=T, diff=1 | 0.021 |
| BC = T, diff=2 | 0.018 |

### Background / motivation

See the image below. 

There are supposed to be 4 "official" bars in the bar plot (eg blue, or perhaps grey too?). 

But we had wanted them to be clickable. The problem was that if ADF value was super-low (eg 0.01) it was hard to click. So the solution was: 4 extra bars. That's what the white bars are for. 
- The white bars were supposed to be underneath (in the z-axis) the other ones, at full width. 
- But the white bars are actually below the blue/etc bars in the y-axis. So now it looks like there are 8 bars. This is confusing, and looks weird too.
- Finally, the white bars were supposed to be the color of the background of the plot (grey). They're white not grey. Weird.

What we want:
- 4 "official" bars vertically distributed
- some means of clicking on something, eg white bars underneath the blue ones in the z-axis

Or even better: Remove the bar plots. Rather, simply have a table of ADF values. (See TODO above)

Screenshot 2024-06-10 at 13 02 46

KatunaNorbert:
Fixes #1180.


pdr_backend/statutil/dash_plots/view_elements.py
```diff
@@ -15,14 +15,31 @@ def display_plots_view(columns):
 
 
 # pylint: disable=line-too-long
-def get_column_graphs(figures: list[dict], title: str, tooltip: str):
+def get_column_graphs(id_parent: str, figures: list[dict], title: str, tooltip: str):
```
calina-c, line 18: I suggest parent_id instead of id_parent

pdr_backend/statutil/dash_plots/view_elements.py
```diff
@@ -32,31 +49,36 @@ def get_column_graphs(figures: list[dict], title: str, tooltip: str):
...
                 "alignItems": "center",
                 "justifyContent": "center",
+                "width": "100%",
             },
         ),
     )
 
 
+def get_table(columns, data):
+    return dash_table.DataTable(
```
calina-c, line 68: @KatunaNorbert I suggest you also take a look at the LakeInfo dashboard. I used the dbc table component and I liked it more than the DataTable from Dash. We can of course have different elements, but maybe you'll like that one better too :)

pdr_backend/statutil/assets/styles.css
```diff
@@ -5,3 +5,7 @@
     background-color: #2b2a26;
     text-align: start;
 }
+
+[class*="table-container"]{
```
calina-c, line 9: Here too, I suggest using the container and column/row structures from dbc. Easier to maintain and pretty much standard since it's Bootstrap underneath. Also gives an uniform feel to our dashboards.

calina-c (commented):
I left some tiny suggestions, if you don't mind :)

pdr_backend/statutil/dash_plots/view_elements.py
KatunaNorbert: I looked at that one but turns out that it doesn't support selection out of the box

pdr_backend/statutil/assets/styles.css
KatunaNorbert: I don't really like using Bootstrap components because they don't provide the flexibility and control that I want. They are faster at higher lvl styling but when you want to do something more custom you end up overriding a lot.

calina-c (approved):

trentmc (approved):

PR #1217, opened 2024-06-13
Fix #1213: [Sim, Aimodel] Nudge regr to have err->0 around current_price

### LINKED ISSUE #1213: [Sim, Aimodel] Nudge regressor to have error -> 0 around current_price
### Background / motivation

We recently shipped regressors, wrapped by classifiers.

The regressors so far are global models (vs local ones that interpolate between points).

Therefore error on any given training point is typically >0. Including error around the current price.

But if there's non-negligible error at current price, then predictions near it will be pretty bad.

To address this, let's nudge the regressor to have error -> 0 around current_price.

Candidate ideas:
1. have 1000x weight on most recent sample, which forces it to have error --> 0.
2. add an offset to the model such that error on current price = 0. I could do this within each linear model's offset, or in a macro-way. Better to do it macro, because then it works for any regressor as we go beyond linear.

### TODOs / DoD

- [ ] Try idea (1), see how it goes
- [ ] Try idea (2), see how it goes
- [ ] Decide whether to keep none, one or both

trentmc:
Fixes #1213

### DIFF
--- a/pdr_backend/aimodel/aimodel.py
+++ b/pdr_backend/aimodel/aimodel.py
@@ -20,6 +20,7 @@ class Aimodel:
         self._y_thr = y_thr  # threshold value for True vs False
         self._sk_classif = sk_classif  # sklearn classifier model
         self._imps_tup = None  # tuple of (imps_avg, imps_stddev)
+        self._ycont_offset = 0.0  # offset to the output of regression
 
     @property
     def do_regr(self) -> bool:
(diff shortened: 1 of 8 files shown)


trentmc:
### Test results
Let's see curprice_xval in `[-8, -5, -2, 2, 5, 8]`

Do it on a regularized linear learning (pick ridge). Note that it's similar results for the other linear learning approaches.

Do it on a non-linear function to show how it's forced. Pick quadratic.

```console
export SHOW_PLOT=true
pytest pdr_backend/aimodel/test/test_aimodel_factory.py::test_aimodel__regr_0error_on_cur_price[RegrLinearRidge]
```

### Result at -8
neg8

### Result at -5
neg5

### Result at -2
neg2

### Result at +2
pos2

### Result at +5
pos5

### Result at +8
pos8

Issue #1209, opened 2024-06-13
[Plots] Write basic end-to-end tests for sim plots & arima plots


trentmc:
### Background / motivation

Sim engine plots have now hit a relatively stable macro-state: using plotly, etc.

But we don't have tests for them at the macro level yet.

We do have _some_ macro-level tests for arima plots: in statutil/test/test_arima_dash_plots.py and test_dash_plots_utils.py. However they're not enough either. For example, [#1206](https://github.com/oceanprotocol/pdr-backend/issues/1206) was a small bug, yet it prevented arima plots from launching. The existing tests didn't catch it.

### TODOs

- [x] Write basic end-to-end tests for sim engine plots (`pdr sim_plots`), such that after a short sim run, can they plot everything without something crashing. To keep runtime fast, this could be done as part of test_sim_engine.py -> done in #1228
- [ ] Write basic end-to-end tests for arima plots (`pdr arima_plots`). Test it on the data that`test_seasonal.py::_get_data()` uses

### Related issues

A good follow-up issue to this one is: [#1208](https://github.com/oceanprotocol/pdr-backend/issues/1208) "In sim and pdr bot tests, need to test both classifier _and_ regressor ai models"

PR #1228, opened 2024-06-17
Add sim_plots testing with Dash testing (Dash duo)

calina-c:
Work on #1209.

### What was done:
- add dependencies for dash duo end-to-end-testing, removed an unused dependency from the codebase
- integration testing for sim_plots dashboard:
     - test all tabs show up
     - test clicking on tabs display each figure/figure group as needed
     - [ ] PENDING: test clicking on bars in var_imps chart toggles contour/line plot -> see below for details
     - improve coverage for various modules -> see below for details

### Varimps issue
I have to work with test parameters to ensure we get a contour plot and varimps can have both variables displayed. However, there is also an UX issue. Due to resizing and adding tabs, the tooltip of the varimps bar covers the bar itself and prevents clicking. We need to fix that first and I'll talk to @KatunaNorbert about it + create a follow-up issue. In the meantime, I propose we merge this branch and continue integration tests for other dashboards.

### Coverage changes
`dash_plots/callbacks.py`: 31% -> 75% (probably full coverage after fixing varimps and adding test for click changes)
`dash_plots/view_elements`: 97% -> 100%
`sim_dash.py`: 80% -> 75% (removed some rows and left as a simple wrapper)
`sim_engine.py`: 99%->98%
`sim_plotter.py`: 36% -> 80% (probably full after fixing varimps and adding test for click changes)
`sim_state.py`: 96% remained the same


calina-c:
Screenshot 2024-06-18 at 09 41 23

UX issue that prevents me from clicking the bar is attached.

calina-c:
Update ref. the var_imps issue:
- the visual click issue is minor and we'll see what we can do, but that will surely be a separate task
- I don't think the effort into adding and testing the varimps bar click is warranted, at least not right now. It's very difficult due to the following factors:
    1. there is no way to manipulate the test data in sim_engine.py to balance variable importances so that multiple bars appear. And if just a single bar appears, all bars are considered "selected" so there is no way to distinguish through the "selected vars" if the click was done
    2. no way to detect the redraw of the line/contour plot itself. We could try some "fireworks" like giving it a unique id to fix this, but still, see (3)
    3. It is incredibly difficult to perform the click. Plotly figures are svgs, so we can not detect individual elements with selectors. We could try offsets of x/y from the corner of the figure but since the selenium backend uses who-knows-what resolution it will take a very long time to figure out and it's possible it will behave badly on other setups.
    
TL;DR: I think we can close the sim_plots part of this task. @trentmc, what do you think?

trentmc:
> TL;DR: I think we can close the sim_plots part of this task. @trentmc, what do you think?

OK.

trentmc (approved):

PR #1265, opened 2024-06-20
Fix 1262: Get prediction data for sim

### LINKED ISSUE #1262: [Sim] Get chain prediction data for trading
**Target**
When running sim with predictions from chain, make sure the required data is present into lake

**To dos**
- [x] add flag inside sim_ss for specifying if chain or model data should be used for predictions signal
- [x] if chain data is selected, check if ppss is configured in a right way for fetching chain data for traid requirements
- [x] if chain data is selected, check if the data is present into lake
- [x] if data is not present fetch data
- [x] write test

KatunaNorbert:
Fixes #1262.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -18,6 +18,8 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
       - name: Setup repository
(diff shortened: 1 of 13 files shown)


KatunaNorbert:
I commented out one of the test because I couldn't find a solution for it, working on local but fails inside github actions.
Will come back to it inside the base branch, let's merge this PR so we can move things forward

kdetry:
LGTM, we can look at the issue at the epic branch

kdetry (approved):

PR #1294, opened 2024-06-25
Fix #1288: Standardise mypy ignores and add dependencies README.

### LINKED ISSUE #1288: [Dependencies] Create a dependency checklist of ignored warnings, caveats and their reasons
### Background / motivation

We have a few ignored warnings from 3rd party libraries. E.g. here: https://github.com/oceanprotocol/pdr-backend/issues/1242.

Also, there are dependencies we marked as unstable, i.e. asyncio in the code here: https://github.com/oceanprotocol/pdr-backend/blob/80604eb8ee0d3340f55d8b25b2befcf6736a01e0/setup.py#L38 and related to this issue that has been opened and reopened accidentally a few times: https://github.com/oceanprotocol/pdr-backend/issues/790

Due to the volume of dependencies to upgrade (each Monday dependabot throws 3-8 dependency upgrades my way) this is becoming difficult to keep just in mind. Plus, it is not right that this information is just in my mind, it should be available and manageable by the entire dev team, when issues arise.

### TODOs / DoD
What this document should contain:
- dependencies that are frozen and why they are frozen. What are we looking for in the changelog of a new version, in order to do a safe upgrade? -> as is the case for asyncio right now
- warnings that have been ignored from 3rd parties and why. What can we remove from the ignore list after an upgrade?
- instructions on adding ignores when adding dependencies -> as it was the case with `types-requests` recently, which was added as a dependency although we generally address missing stubs from 3rd parties by ignoring the issue in mypy.

I would like to document these in a README or something similar.

After discussing with Trent:
- [ ] we should not allow pytest tests to pass in CI if they contain warnings (pre-requisite: a tutorial for S3)

calina-c:
Closes #1288.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Test with pytest
         id: pytest
         run: |
-          coverage run --source=pdr_backend --omit=*/test/*,*/test_ganache/*,*/test_noganache/* -m pytest -W error
+          coverage run --source=pdr_backend --omit=*/test/*,*/test_ganache/*,*/test_noganache/* -m pytest
           coverage report
           coverage xml
       - name: Publish code coverage
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Test with pytest
         id: pytest
         run: |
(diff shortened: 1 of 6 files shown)


calina-c:
Run here: https://github.com/oceanprotocol/pdr-backend/actions/runs/9660765724/job/26647028851?pr=1294
Shows that warnings become errors, as intended for new CI standards.

trentmc (approved):
Please link to this from the the dev readme. Then, approved. Thank you!

Issue #1202, opened 2024-06-12
[Sim, ppss] Define average stake amount and accuracy, *per feed*


trizin:
### Background / motivation

Status quo: ppss enables the user to set a single average accuracy (of other traders) overall. This is the setting `predictoor_ss.sim_only.others_accuracy`. Same for stake; setting is `others_stake`.

Datapoint: average accuracy makes a huge difference to profitability of predictoor. Eg value of 50% makes it easier to be profitable, value of 53% makes it much harder, 56% harder yet

Datapoint: in practice, average accuracy is quite different per feed. On some feeds it might be 51.8%, on others 53%, on others 56%. These accuracies tend to be fairly stable over the weeks, and slowly increasing due to an arms race among predictoors.

The datapoints above lead to a problem: when we simulate, if we change the feed, we have to update the single average accuracy to be reflective of the given feed we're predicting. It's a pain. 

### Towards a solution

We need to enable users to define average stake amount and accuracy, _per feed_. This will help sim give more realistic profitabilities for prediction bots.

### TODO

- [ ] ppss.yaml enables user to define average accuracy and stake amount, _per feed_. Add unit tests accordingly
- [ ] reflect this upwards in the code to the level of the simulator

PR #1295, opened 2024-06-25
Feat #1202 - Fetch stake and accuracy per feed from live feeds

trizin:
#1298

- ppss parameter to toggle "fetch stake and accuracy per feed from live feeds"
- Function to fetch stake and accuracy and tests


KatunaNorbert:
Good quick solution for calculating this 2 values, should offer much accurate environment for predictoor simulations.
My only comment is that the periods for calculating this values inside the accuracy endpoint are fixed, so this could become less relevant depending on the period you are running your simulations on.
Ex: for 5m fields the accuracy and stakes are calculated over the last week, but the simulation could be run over the last month

pdr_backend/sim/sim_engine.py
```diff
@@ -58,6 +61,24 @@ def __init__(
         else:
             self.multi_id = str(uuid.uuid4())
 
+        use_live_data = self.ppss.predictoor_ss.use_live_stake_and_accuracy
+
+        if use_live_data:
+            data = get_avg_stake_and_accuracy_for_feed(self.predict_feed)
+            if data:
+                self.others_accuracy, self.others_stake = data
+                self.others_stake = self.others_stake.amt_eth
```
KatunaNorbert, line 70: I think the `self.others_stake` is referring to total stake for an epoch, but the stake amount from the api( `self.others_stake` ) is per day, so for an accurate stake you have to divide that by number of epochs

pdr_backend/sim/sim_engine.py
```diff
@@ -203,8 +224,8 @@ def run_one_iter(self, test_i: int, mergedohlcv_df: pl.DataFrame):
         st.aim.update(acc_est, acc_l, acc_u, f1, precision, recall, loss, yerr)
 
         # track predictoor profit
-        tot_stake = others_stake + stake_amt
-        others_stake_correct = others_stake * pdr_ss.others_accuracy
+        tot_stake = self.others_stake + stake_amt
+        others_stake_correct = self.others_stake * self.others_accuracy
```
KatunaNorbert, line 228: Just found another issue here, the `self.others_accuracy` value when accuracy endpoint data is used is in this format `52.3` which multiplied by total stake gives a higher value than the total stake.
I think what is expected is that the `self.others_accuracy` values is in this format `0.523`, this was the `others_stake_correct` is a lower number than the `self.others_stake`

KatunaNorbert (approved):
LGTM

PR #1297, opened 2024-06-25
Fix #1263: Transform lake data into prediction signals

### LINKED ISSUE #1263: [Sim] Transform lake data into prediction signals for trader
### Target

Create prediction signal that is suitable to be used for the trader step

Parent epic: [#1257](https://github.com/oceanprotocol/pdr-backend/issues/1257) Integrate subgraph prediction into sim trading

### TODO

- [x] Transform the lake data into prediction signals
Read data from pdr_payouts table, filter payouts by slot and calculate the prediction from `roundSumStakes` and `roundSumStakesUp`
Final data should be stored `slot - prediction` pairs list
- [x] Use prediction signal inside trader sim
- [x] Test and Verify

kdetry:
Fixes #1263 

Changes proposed in this PR:
- The lake data is used in the sim engine iteration


pdr_backend/sim/sim_engine.py
```diff
@@ -271,6 +291,51 @@ def save_state(self, i: int, N: int):
...
+            self.ppss.web3_pp.owner_addrs,
+        )
+
+        sPE = 300 if self.predict_feed.timeframe == "5m" else 3600
+        # Filter contracts with the correct token pair and timeframe
+        contract_to_use = [
+            addr
+            for addr, feed in contracts.items()
+            if feed.symbol
+            == f"{self.predict_feed.pair.base_str}/{self.predict_feed.pair.quote_str}"
```
KatunaNorbert, line 308: you can use `self.predict_feed.pair.pair_str` instead of `f"{self.predict_feed.pair.base_str}/{self.predict_feed.pair.quote_str}"`

pdr_backend/sim/sim_engine.py
```diff
@@ -271,6 +291,51 @@ def save_state(self, i: int, N: int):
...
+            and feed.seconds_per_epoch == sPE
+        ]
+
+        query = f"""
+            SELECT
+                slot,
+                CASE
+                    WHEN roundSumStakes = 0.0 THEN NULL
+                    WHEN roundSumStakesUp = 0.0 THEN NULL
```
KatunaNorbert, line 317: for this case when `roundSumStakesUp = 0.0` the value shouldn't be `NULL` because maybe all the predictoors have staked down. It should be `NULL` only if `roundSumStakes = 0.0`

pdr_backend/sim/sim_engine.py
```diff
@@ -271,6 +291,51 @@ def save_state(self, i: int, N: int):
         return True, False
 
     @enforce_types
+    def _get_prediction_dataset(
```
KatunaNorbert, line 294: could you rename this and the test to `_get_predictions_signals_data`? This way is more specific

KatunaNorbert (approved):

PR #1301, opened 2024-06-26
Fix #1300: Add/update licenseheaders.

### LINKED ISSUE #1300: [License headers] License headers are outdated, need to be updated to 2024
### Background / motivation

Currently the license headers are still set for 2023. We need to run the command to update them.

calina-c:
Update licenseheaders for README files to 2024.
Add licenseheaders to new files.

@trentmc do you want to setup some configuration to do this automatically when new files are added/the year changes?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,3 +1,7 @@
+<!--
+Copyright 2024 Ocean Protocol Foundation
+SPDX-License-Identifier: Apache-2.0
+-->
 ---
 name: Bug report
 about: Create a report to help us improve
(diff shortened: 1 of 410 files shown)


trentmc:
> @trentmc do you want to setup some configuration to do this automatically when new files are added/the year changes?

This would be amazing (assuming it's sufficiently lightweight)

calina-c:
> > @trentmc do you want to setup some configuration to do this automatically when new files are added/the year changes?
> 
> This would be amazing (assuming it's sufficiently lightweight)

I've investigated and it seems we can only do it via pre-commit hooks. Using the CI, there is no specific job and running the raw command does not show modifiable files, even if the `dry` option is present.

Regarding pre-commits, I've always been partial to pre-commits being part of our flow, but they are currently not, so I don't want to rattle any more cages regarding the process.

TL;DR: I updated the files with licenseheaders, added a template and instructions in dev.md on how devs can do this once a year. And that's it :)

trentmc (approved):

PR #1314, opened 2024-06-27
Fix #1108: PPSS "passing-around" cleanup (alternate version) 

### LINKED ISSUE #1108: [Lake][DuckDB] Clean ppss handling, such that it's not so verbose

calina-c:
Closes #1108


calina-c:
Closed due to feedback from Roberto.

PR #1315, opened 2024-06-27
Fix #1313: mock GQLDataFactory for the sim engine

### LINKED ISSUE #1313: [Sim] Mock subgraph calls inside tests
For testing the chain data insertion inside the trading simulations, the test makes request to production subgraph which can overload the system.

To do:
- [x] update the tests to use mocked subgraph data

kdetry:
Fixes #1313

the description will be added


pdr_backend/sim/sim_engine.py
```diff
@@ -374,6 +374,7 @@ def _get_past_predictions_from_chain(self, ppss: PPSS):
         """
         data = db.query_data(query)
 
+        print("data-->", data)
```
KatunaNorbert, line 377: remove prints or use logger

lake_data/plots/daily_unique_predictoors.png
KatunaNorbert, line 1: wondering if this will cause conflicts with main, there was a PR merged that removed this cached values

pdr_backend/sim/sim_engine.py
kdetry: done

KatunaNorbert (approved):

PR #1129, opened 2024-06-05
[Frozen][To Be Merged After DuckDB Inetgration] - Improve lakeinfo structure and reporting

calina-c:
Work on #1104. Cleanup and filtering. Moving validation report to a single tab.

PR #1319, opened 2024-06-28
Fix #1104: Improve lakeinfo structure and reporting

### LINKED ISSUE #1104: [Dash][App][Lake] Improve describe to be more user/debug friendly
### Motivation
Improve lake app to provide more visibility/intuition about the lake. The user should be able to use it as a workbook/dashboard w/ a simple way of obtaining intuition about the lake.

### FILTER
- single userid only
- no additional filters for now

### TABLES
all sql tables, shown as a dataframe, split by system-tables
-> raw tables
-> bronze tables

DO NOT SOLVE FOR ETL TABLES
These are temporary/swap/system tables and in an ideal world, would be hidden behind roles.
- Temp Table
- Update Table
- ETL Views

### PLOTS
You can select/update what's grouped by/sorted by
- userid
- timeframe
- asset

###  DoD
- [ ] all lake tables visualized in dataframe format
- [ ] table rows plotted as a scatter plot by timestamp/day

calina-c:
Closes #1104. Cherry picked from #1129.

TODO: In the meantime, we added tests for our dashboards. I think we should add Dash testing to this dashboard as well, but I would prefer doing it in a different PR.


trentmc (approved):

calina-c:
Decided with Trent to add some basic unit tests to LakeInfo.

idiom-bytes:
@trentmc @calina-c 
I'm not sure why all of this was separated from the lake module?

**we now have 3 clis that are poorly named... structured**
cli/cli_module.py (main cli)
cli/cli_module_lake.py (lake cli)
lake_info/cli.py (lakeinfo cli)

**the core problem is that these tools exist to serve the lake module and help operate the lake along with drop, update, and many other functionalities...**

the validations/info/debuggin/operational commands should remain inside of lake module... not moved into something that is to serve FE

I believe all frontend/webapp tools should exist in a separate module
`/pdr_backend/` -> serves all the core engine systems, apis, etc..
`/pdr_frontend/` -> provides all implementations for all webapp requirements

all core functionality for the web apps should then be referenced from pdr_backend

pytest should first run all coverage inside `pdr_backend/` and then `pdr_frontend/`

idiom-bytes:
I have created a ticket requesting that we address this, including separating the webapp logic from engine logic, so we don't run into these problems in the future

https://github.com/oceanprotocol/pdr-backend/issues/1419

trentmc:
The "pdr_backend/" directory is meant to house *all* the code, frontend and backend and everything in between. It's meant to mirror the name of the repo itself. The way that most well structured python repos work. And, needed for pypi. For historical reasons it has "backend" in the name. Let's not change that name right now, it would be a huge pain.

All cli stuff should be in pdr_backend/cli/. It can be cli to run lake, sim, sim plots, predictoor bots, trader bots, or dashboard.

And we have many support directories. Building blocks, if you will. Ppss, util, lake, aimodel, etc etc. They interlink though each has fairly well-defined responsibility.

We should 100% definitely NOT add a new directory parallel to pdr_backend. It would totally mess with the overall structure of the repo, including compromising how it needs to behave for pypi.

And there should definitely not be a directory named frontend/, anywhere. We have sim plots, we have the feed analysis app, and now we have the dashboard app. Each gets their own subdirectory of pdr_backend/, with an appropriate label.

PR #1340, opened 2024-07-05
Fix #1258: '[Warning] DeprecationWarning: datetime.datetime.utcnow()'

### LINKED ISSUE #1258: [Warning] DeprecationWarning: datetime.datetime.utcnow() ... scheduled for removal
### Summary

In console, run:
```console
pytest pdr_backend/predictoor/test
``` 

Tests will pass, but it will give the warning:
```text
pdr_backend/predictoor/test/test_predictoor_agent.py: 30 warnings
  /Users/trentmc/code/pdr-backend/pdr_backend/util/time_types.py:133: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    dt = datetime.utcnow()
```

### Full traceback

```text
====================================================================================================================================== 1 passed in 1.06s ======================================================================================================================================
(venv) trentmc@trents-m3: ~/code/pdr-backend $ pytest pdr_backend/predictoor/test
===================================================================================================================================== test session starts =====================================================================================================================================
platform darwin -- Python 3.12.3, pytest-8.2.2, pluggy-1.5.0
rootdir: /Users/trentmc/code/pdr-backend
configfile: pytest.ini
plugins: web3-6.19.0, typeguard-4.2.1, env-1.1.3, dash-2.17.1, asyncio-0.21.1, requests-mock-1.12.1
asyncio: mode=Mode.STRICT
collected 20 items                                                                                                                                                                                                                                                                            

pdr_backend/predictoor/test/test_find_shared_slots.py ......                                                                                                                                                                                                                            [ 30%]
pdr_backend/predictoor/test/test_predictoor_agent.py ...........                                                                                                                                                                                                                        [ 85%]
pdr_backend/predictoor/test/test_stakes_per_slot.py ...                                                                                                                                                                                                                                 [100%]

====================================================================================================================================== warnings summary =======================================================================================================================================
pdr_backend/predictoor/test/test_predictoor_agent.py: 30 warnings
  /Users/trentmc/code/pdr-backend/pdr_backend/util/time_types.py:133: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    dt = datetime.utcnow()

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================================================================================================================== 20 passed, 30 warnings in 12.75s ===============================================================================================================================
```

trentmc:
Fix #1258 

Places where `datetime.utcnow()` was used, and what was done:
1. `dt_now_UTC()` used it. So, reimplemented as simply `return datetime.now(UTC)`.
1. `test_time_now_implicit()` used it. But that had a couple hitches, which took hours to fix. Details below.
2. One or two other places in tests.

Details on `test_time_now_implicit()` work:
- It used `freezegun.freeze_time()` decorator. However, that method can't properly handle timezones
  - When `datetime.utcnow()` was used, the issue was masked
  - However when switching to `datetime.now(UTC)`, the issue became apparent
   - I wrote a new a test `test_how_freezegun_is_wrong()` to illustrate freezegun's issue. We will keep that, to emphasize that we're not using it.
- To make `test_time_now_implicit()` pass, I used `time_machine.travel()` decorator instead. `time_machine` _can_ handle timezones. Bonus: unlike freezegun, it's still maintained.
- There was another hitch though. `time_machine.travel()` needs a timezone object, which was 5 hours offset from UTC. I couldn't find anything off-the-shelf to get this timezone. However `zoneinfo.available_timezones()` is a list of all timezones. So, I created a new method `tz_offset_from_utc()` that iterates through that list to find a timezone that had the target offset. I tested that too.

Other refactoring:
- `pdr_backend/ppss/test/test_timestamp.py` should not exist. It was only testing `pdr_backend/util/time_types.py` functionality. So I moved its tests there
- To help debugging, I created a simple `timestr()` function. Much easier than working with unix times (in s or ms -- too many digits for my brain); or in ISO times came (the string sometimes had timezone and sometimes not, and sometimes had ms and sometimes not).


setup.py
```diff
@@ -42,6 +42,7 @@
     "requests-mock==1.12.1",
     "scikit-learn==1.5.0",
     "statsmodels==0.14.2",
+    "time_machine==2.14.2",
```
calina-c, line 45: I suggest we remove freezegun, in this case. There is no point in having two dependencies that do the same(-ish) thing.

pdr_backend/util/time_types.py
```diff
@@ -132,7 +133,42 @@ def pretty_timestr(self) -> str:
...
-    dt = datetime.utcnow()
-    dt = dt.replace(tzinfo=timezone.utc)  # tack on timezone
-    return dt
+def timestr(dt: datetime) -> str:
+    """Simple time string, useful for testing"""
+    return dt.strftime("%Y-%m-%d %H:%M:%S")
+
+
+@enforce_types
+def tz_offset_from_utc(delta_hours: int):
```
calina-c, line 142: as far as I can tell, this function only serves the purpose of defining a constant for the tests (`TZ_UTC_MINUS_5`). I think it would be better if we removed this function. In the tests, we can simply define the tested timezone as ""America/New_York" or something similar.

Alternately, we could at least move this function in some test resources/utils part for it to be clearly delimited from main code.

calina-c (changes requested):

setup.py
trentmc: https://github.com/oceanprotocol/pdr-backend/pull/1344

pdr_backend/util/time_types.py
trentmc: That makes sense. This PR does that: https://github.com/oceanprotocol/pdr-backend/pull/1344

Issue #1348, opened 2024-07-05
[DeprecationWarning] for pkg_resources, from test_dash_plots.py


trentmc:
### How to reproduce

In console:
```console
pytest pdr_backend/sim/test/test_dash_plots.py
```

Returns error: 
```text
DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
```

### Notes

My system setup: MacOS running on MacBook Pro M3

Dependencies guidelines [here](https://github.com/oceanprotocol/pdr-backend/blob/main/READMEs/dependencies.md).

### Full traceback

```text
(venv) trentmc@trents-m3: ~/code/pdr-backend $ pytest pdr_backend/sim/test/test_dash_plots.py 
====================================================================================================== test session starts ======================================================================================================
platform darwin -- Python 3.12.3, pytest-8.2.2, pluggy-1.5.0
rootdir: /Users/trentmc/code/pdr-backend
configfile: pytest.ini
plugins: time-machine-2.14.2, env-1.1.3, dash-2.17.1, asyncio-0.21.1, typeguard-4.3.0, web3-6.20.0, requests-mock-1.12.1
asyncio: mode=Mode.STRICT
collected 0 items / 1 error                                                                                                                                                                                                     

============================================================================================================ ERRORS =============================================================================================================
___________________________________________________________________________________ ERROR collecting pdr_backend/sim/test/test_dash_plots.py ____________________________________________________________________________________
pdr_backend/sim/test/test_dash_plots.py:9: in <module>
    from pdr_backend.sim.dash_plots.util import get_figures_by_state
pdr_backend/sim/dash_plots/util.py:6: in <module>
    import stopit
venv/lib/python3.12/site-packages/stopit/__init__.py:10: in <module>
    import pkg_resources
venv/lib/python3.12/site-packages/pkg_resources/__init__.py:105: in <module>
    warnings.warn(
E   DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
==================================================================================================== short test summary info ====================================================================================================
ERROR pdr_backend/sim/test/test_dash_plots.py - DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
[...]

PR #1355, opened 2024-07-06
Follow-up on #1348: Add dependency explanation ref. to stopit decorator.

calina-c:
This issue was already closed by adding the ignore in pytest.ini, but I also added some explanatory test in dependencies.md to cover our bases.

For the record, I also tried to replace `stopit` with something better or create a custom decorator but I only found solutions compatible with Linux and Mac. The sim_plots dashboard would still freeze for windows users. For now, we don't need to replace the `stopit` dependency since it only causes this small warning in the tests, so I didn't insist with my research on execution stoppers.

@trentmc do you think we should delve into this more? I think we can live with this warning for now.


trentmc (approved):
Current solution is fine. 

Let's not spend time on this further.

PR #1356, opened 2024-07-06
Fix #932 - Paralellise ohlcv feed fetching

### LINKED ISSUE #932: [Lake] Parallelize fetching data across >>1 coins
### Background / motivation

We currently fetch data from Binance one token at at time.

So if there are 10 coins (BTC, ETH, DOT, ..) then it takes 10x longer.

### TODOs / DoD

- [ ] Parallelize fetching data, in `lake/ohclv_data_factory.py`. It's likely a straightforward conversion of a for loop

### Related github issues

- [#804](https://github.com/oceanprotocol/pdr-backend/issues/804) "[Sim, make $] Make benchmarking parallel, via threading"

Note: we could also parallelize grabbing data within a feed. However this isn't as easy, because the algorithm needs to check what's there already. So consider this for later. But also maybe not needed because at some point we'll have historical data repo / bundle.

calina-c:
Closes #932.

### DIFF
--- a/pdr_backend/lake/test/test_ohlcv_data_factory.py
+++ b/pdr_backend/lake/test/test_ohlcv_data_factory.py
@@ -2,10 +2,11 @@
 # Copyright 2024 Ocean Protocol Foundation
 # SPDX-License-Identifier: Apache-2.0
 #
+import asyncio
 import os
 import time
 from typing import List
-from unittest.mock import Mock, patch
+from unittest.mock import AsyncMock, Mock, patch
 
 from enforce_typing import enforce_types
 import numpy as np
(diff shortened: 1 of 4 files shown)


trentmc (approved):

pdr_backend/lake/test/test_ohlcv_data_factory.py
```diff
@@ -314,7 +315,7 @@ def test_get_mergedohlcv_df_calls(
     mock_merge_rawohlcv_dfs.return_value = Mock(spec=pl.DataFrame)
     _, factory = _lake_ss_1feed(tmpdir, "binanceus ETH/USDT h 5m")
 
-    factory._update_rawohlcv_files = Mock(return_value=None)
+    factory._update_rawohlcv_files = AsyncMock(return_value=None)
```
idiom-bytes, line 318: thanks for providing different examples, great!

pdr_backend/sim/multisim_engine.py
```diff
@@ -54,6 +57,15 @@ def run(self):
         ss = self.ss
         logger.info("Multisim engine: start. # runs = %s", ss.n_runs)
         self.initialize_csv_with_header()
+
+        # must be done in advance, as it's async
+        for run_i in range(ss.n_runs):
```
idiom-bytes, line 62: makes sense too as this is where multisim kicks off, cheers!

PR #1383, opened 2024-07-11
Separates some concerns in payouts management.

calina-c:


calina-c:
Some of these changes happened naturally in main, in other ways. Others I recycled into a different PR.

PR #1395, opened 2024-07-15
Fix #1385: Add switch to display selected predictoors feeds only

### LINKED ISSUE #1385: [PdrDashboard, v0.2] Filter feeds table based ion predictoors
**Target:**
Make it easy to find for a selected predictoor, the feeds it predicted for

**To do:**
- [ ] add a toggle to automatically filter feeds by predictoors that have made predictions
- [ ] if toggle is activated filter the feeds accordingly
- [ ] make all feeds selected by default

KatunaNorbert:
Fix #1385

### DIFF
--- a/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
+++ b/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
@@ -192,7 +192,8 @@ def get_callbacks(app):
             feed_ids = get_feed_ids_based_on_predictoors_from_db(
                 lake_dir,
                 predictoor_addrs=[
-                    predictoors_table[i]["user"] for i in predictoors_table_selected_rows
+                    predictoors_table[i]["user"]
+                    for i in predictoors_table_selected_rows
                 ],
             )
             filtered_data = [obj for obj in feeds_data if obj["contract"] in feed_ids]
--- a/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
+++ b/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
@@ -5,8 +5,7 @@ from pdr_backend.analytics.predictoor_dashboard.dash_components.util import (
     get_predictoors_data_from_db,
     get_payouts_from_db,
     filter_objects_by_field,
-    get_unique_payout_ids_based_on_predictoors_from_db,
-    filter_feeds_data_by_payout_IDs
+    get_feed_ids_based_on_predictoors_from_db,
 )
 from pdr_backend.analytics.predictoor_dashboard.dash_components.view_elements import (
     get_graph,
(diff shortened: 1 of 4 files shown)


pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
```diff
@@ -138,14 +139,42 @@ def update_predictoors_table_on_search(search_value, predictoors_data):
...
+    def update_feeds_table_on_search(
+        search_value,
+        feeds_data,
+        predictoor_feeds_only,
+        predictoors_table_selected_rows,
+        predictoors_data,
+        lake_dir,
+    ):
+        # Extract selected predictoor addresses
+        predictoors_addrs = [
```
calina-c, line 157: I would suggest adding a function outside of the callbacks, with the purpose of extracting **selected** rows from the predictoors/feeds table. Even if it is a one-liner, it seems specific enough to add, since we repeat it in multple places.

pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
```diff
@@ -138,14 +139,42 @@ def update_predictoors_table_on_search(search_value, predictoors_data):
...
+            predictoors_data[i]["user"] for i in predictoors_table_selected_rows
+        ]
 
         # filter feeds by pair address
-        filtered_data = filter_objects_by_field(feeds_data, "pair", search_value)
+        filtered_data = (
```
calina-c, line 162: so why don't we integrate this in the query as well?
I know we discussed it. When searching for predictoors/feeds based on user input, we are still filtering from a list of objects instead of hitting the DB (with a LIKE query, no big deal). The current implementation is slow and should be addressed soon.

I asked you on Slack wrt to the functionality described here. Especially so, if we are hitting the database for feed ids based on predictoors, let us have not an id query, but a full feed query which also selects based on search value. This would take care of the feeds query.

We can adapt the predictoors part from a list to a db hit later, if outside the scope here.

calina-c: basically what we do right now is slow because of the following reasons:
- we retrieve the full feed list at app startup, and load it to the state
- in this function, we filter our list based on a keyword -> call it list A
- we hit the database **for another list!**: a list of feed ids -> call it list B
- we filter list A based on list B

So we have the startup database hit, the ids database hit and then the filter costs us O(len(A)) * O(len(B)), if you look at the list comprehension, especially since B is not ordered/hashed, so the lookup for "obj["contract"] in feed_ids" is not optimised: https://stackoverflow.com/questions/13884177/complexity-of-in-operator-in-python

Instead, what we can do:
- one database hit here, to retrieve all feed information based on two inputs: the selected predictoors and the pair search value input
- at startup, we can use the same query, without the LIKE filter

calina-c (changes requested):

pdr_backend/analytics/predictoor_dashboard/dash_components/util.py
```diff
@@ -53,6 +53,35 @@ def get_predictoors_data_from_db(lake_dir: str):
...
+            FROM {Payout.get_lake_table_name()}
+            WHERE (
+                {" OR ".join([f"ID LIKE '%{item}%'" for item in predictoor_addrs])}
+            )
+            GROUP BY LEFT(ID, POSITION('-' IN ID) - 1)
+        );
+    """
+
+    # Execute the query
+    resp = _query_db(
```
calina-c, line 75: I think you can have this in a single line

pdr_backend/analytics/predictoor_dashboard/dash_components/util.py
```diff
@@ -53,6 +53,35 @@ def get_predictoors_data_from_db(lake_dir: str):
...
+        lake_dir,
+        query,
+    )
+
+    # Extract feed addresses from the response
+    feed_ids = [obj["feed_addr"] for obj in resp]
```
calina-c, line 81: I'm not sure about this. In theory, there should be a way to retrieve the data as a list instead of a DataFrame. Or to select the column as a list directly, instead of iterating through the objects. I'll investigate to help.

calina-c (changes requested):

pdr_backend/analytics/predictoor_dashboard/dash_components/util.py
calina-c: maybe with `unnest`? https://duckdb.org/docs/sql/query_syntax/select.html#unnest-function

calina-c: or `list()`

Returns a list of all regions in order of the amount column:

SELECT list(region ORDER BY amount DESC)
FROM sales;

https://duckdb.org/docs/sql/functions/aggregates.html

pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
calina-c: db query to be addressed later

calina-c (approved):

PR #1398, opened 2024-07-15
Fix #1384: Improve Predictoors Table

### LINKED ISSUE #1384: [PdrDashboard, v0.2] Improve Predictoors Table
Parent: https://github.com/oceanprotocol/pdr-backend/issues/1357

Enhance the Predictoors table by shortening the user column values, adding columns for profit, accuracy, and average stake, and sorting the table by profit by default.

## DoD

***Shorten User Column Values:***
- [ ] Display only the first and last few characters of the user column values to make them shorter and more readable.

***Add Columns:***
- [ ] Calculate profit, ACC and avg stake for every predictoor
- [ ] Add columns for profit, accuracy, and average stake to the table.

***Default Sorting:***
- [ ] Sort the table by the profit column by default.

kdetry:
Fixes #1384 

Changes on this PR:
- the total_profit, total_acc, avg_stake columns have been added.
- the sorting logic has been added
- the user column is shortened

### DIFF
--- a/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
+++ b/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
@@ -39,7 +39,7 @@ def get_callbacks(app):
             user_payout_stats = get_user_payouts_stats_from_db(files_dir)
             return feeds_data, predictoors_data, user_payout_stats, None
         except Exception as e:
-            return None, None, dash.html.H3(str(e))
+            return None, None, None, dash.html.H3(str(e))
 
     @app.callback(
         Output("accuracy_chart", "children"),
--- a/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
+++ b/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
@@ -7,6 +7,7 @@ from pdr_backend.analytics.predictoor_dashboard.dash_components.util import (
     get_user_payouts_stats_from_db,
     filter_objects_by_field,
     get_feed_ids_based_on_predictoors_from_db,
+    merge_payout_stats_with_predictoors,
 )
 from pdr_backend.analytics.predictoor_dashboard.dash_components.view_elements import (
     get_graph,
@@ -36,8 +37,12 @@ def get_callbacks(app):
     def get_input_data_from_db(files_dir):
         try:
             feeds_data = get_feeds_data_from_db(files_dir)
-            predictoors_data = get_predictoors_data_from_db(files_dir)
             user_payout_stats = get_user_payouts_stats_from_db(files_dir)
+
+            predictoors_data = merge_payout_stats_with_predictoors(
+                user_payout_stats, get_predictoors_data_from_db(files_dir)
+            )
+
             return feeds_data, predictoors_data, user_payout_stats, None
         except Exception as e:
             return None, None, None, dash.html.H3(str(e))
@@ -126,7 +131,6 @@ def get_callbacks(app):
             Input("predictoors_table", "selected_rows"),
             Input("predictoors_table", "data"),
             Input("predictoors-data", "data"),
-            Input("user-payout-stats", "data"),
         ],
     )
     def update_predictoors_table_on_search(
@@ -134,7 +138,6 @@ def get_callbacks(app):
         selected_rows,
         predictoors_table,
         predictoors_data,
-        user_payout_stats,
     ):
         selected_predictoors = [predictoors_table[i] for i in selected_rows]
         if not search_value:
@@ -144,37 +147,6 @@ def get_callbacks(app):
             filtered_data = filter_objects_by_field(
                 predictoors_data, "user", search_value, selected_predictoors
             )
-
-        if user_payout_stats:
-            # Her bir data elemanı için, eşleşen user_payout_stat bulunur ve güncellenir.
-            # Eşleşme yoksa, orijinal data elemanı korunur.
-            filtered_data = [
-                {
-                    **data,
-                    **next(
-                        (
-                            user_payout_stat
-                            for user_payout_stat in user_payout_stats
-                            if user_payout_stat["user"] == data["user"]
-                        ),
-                        {},
-                    ),
-                }
-                for data in filtered_data
-            ]
-
-            # shorten the user address
-            for data in filtered_data:
-                new_data = {
-                    "user_address": data["user"][:5] + "..." + data["user"][-5:],
-                    "total_profit": round(data["total_profit"], 2),
-                    "avg_accuracy": round(data["avg_accuracy"], 2),
-                    "avg_stake": round(data["avg_stake"], 2),
-                    "user": data["user"],
-                }
-
-                data.clear()
-                data.update(new_data)
 
         selected_predictoor_indices = [
             i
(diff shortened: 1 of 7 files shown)


calina-c:
@kdetry @KatunaNorbert @idiom-bytes @trentmc 
Looking at this PR and working with the predictoor dashboard (and honestly using my personal banking app) recently have given me an idea. At most times, people will only be interested in some specific accounts and it's difficult for them to manage the addresses of those accounts. I propose we add an aliasing table, where users can locally "name" their accounts in a user-friendly way or maybe mark them as favourites (and show them on top of the list). We can possibly do some sort of "favourite" mark for feeds, as well.

Improvements I see:
- easier to filter and manage. The user can set the name for "0xabc1234..." as "Alice", "0xefgh5678..." as "Bob" and filter by those values. We can even toggle smth like:
    - only showing the favourites
    - showing the favourites on top
    - showing everything
obviously the first case would ease a lot of mental load for the user and database load for the app itself
- maybe autoselection based on favourites?

For us it is an extra JOIN, but for the user it might be a huge time-saver since they don't have to paste in filters all the time.

Let me know what you guys think. Maybe it's a bad idea, maybe it's not doable *now*, but I'd like to discuss it and have an issue if it looks useful to you.

trentmc:
> people will only be interested in some specific accounts and it's difficult for them to manage the addresses of those accounts. I propose we add an aliasing table, where users can locally "name" their accounts in a user-friendly way

Gm @calina-c , you're absolutely right, people mostly want to see their accounts, or compare their accounts to global. 

The [Predictoor Dashboard prototype slides](https://docs.google.com/presentation/d/18y_nfpc3e-dop5NtfIisgUkc-oPt8dft9WyAnqhSqZg/edit#slide=id.g20f5357f625_0_1639) had this feature too:
Screenshot 2024-07-16 at 09 28 19

And in fact, @KatunaNorbert and I were talking about it yesterday:)

And, we discussed a simple way to do it: specify via ppss.yaml. My specific suggestion: **have a new section in predictoor_ss called "my_addresses" where you can optionally list your addresses**. That's it. Then this can get picked up by Predictoor Dashboard or anywhere else. 
- We know how to deal with address format already, see eg web3_pp section.
- I'm recommending to put it into predictoor_ss, versus more general, because we may want to specify different addresses for other contexts (eg traders). One step at a time.

KatunaNorbert:
> And in fact, @KatunaNorbert and I were talking about it yesterday:)

I like it when multiple people have the same idea, this means it's something of a high values.

We can prioritise it, add it to v0.2 and start working on it right away. Looks like we have a clear description on how to implement it: 
- specify addresses inside ppss.yaml
- make addresses and corresponding feeds selected inside the table when the app starts
- nice to have: add a button so app resets to your predictoors when clicking on a button

trentmc:
> Looks like we have a clear description on how to implement it:
> specify addresses inside ppss.yaml

Yes, and specifically, put it inside the predictoor_ss section. Because it's really for people running predictoors. This also allows us to re-use all the predictoor_ss infra.

KatunaNorbert:
Created the issue:
https://github.com/oceanprotocol/pdr-backend/issues/1402

kdetry:
I found it inconvenient to fetch all payouts and calculate the table stats with Python because of reloading a huge data on memory. Instead, I implemented an SQL logic to improve performance and reduce performance costs

pdr_backend/analytics/predictoor_dashboard/test/test_util.py
```diff
@@ -127,3 +128,30 @@ def test_unrelated_trigger(sample_table_rows):
...
+    tmpdir,
+    _sample_payouts,
+):
+    ppss, _ = _prepare_test_db(
+        tmpdir, _sample_payouts, table_name=Payout.get_lake_table_name()
+    )
+
+    result = get_user_payouts_stats_from_db(ppss.lake_ss.lake_dir)
+
+    print("result--->", result)
```
KatunaNorbert, line 143: remove print or use logger if needed

pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
```diff
@@ -142,13 +145,49 @@ def update_predictoors_table_on_search(
                 predictoors_data, "user", search_value, selected_predictoors
             )
 
+        if user_payout_stats:
+            # Her bir data elemanı için, eşleşen user_payout_stat bulunur ve güncellenir.
+            # Eşleşme yoksa, orijinal data elemanı korunur.
```
KatunaNorbert, line 150: I think we should translate this to english 😄

pdr_backend/analytics/predictoor_dashboard/dash_components/util.py
```diff
@@ -54,6 +54,23 @@ def get_predictoors_data_from_db(lake_dir: str):
...
+                "user",
+                SUM(payout - stake) AS total_profit,
+                SUM(CASE WHEN payout > 0 THEN 1 ELSE 0 END) * 100.0 / COUNT(*) AS avg_accuracy,
+                AVG(stake) AS avg_stake
+            FROM 
+                {Payout.get_lake_table_name()}
+            GROUP BY 
+                "user" 
+        """,
+    )
```
KatunaNorbert, line 71: good job with the query, the data comes in very fast!
I'm wondering if we can unite this query with the get predictoors one and have a get predictoors data query containing all the data we need, should be straight forward.

KatunaNorbert:
Could you also make make the column containing the tables wider? It should be wide enough to have all the columns visible

pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
kdetry: Sorry my fault :D

pdr_backend/analytics/predictoor_dashboard/dash_components/util.py
```diff
@@ -143,3 +160,30 @@ def select_or_clear_all_by_table(
         selected_rows = list(range(len(rows)))
 
     return selected_rows
+
+
+@enforce_types
+def process_user_payout_stats(
```
KatunaNorbert, line 166: Looks much cleaner now, but we should add a better name to this function: "get_predictoors_data_from_payouts" or something similar

KatunaNorbert:
The accuracy and avg stake on the predictoor row doesn't mach the values from the metrics. I have all the feeds selected for the predictoor so they should mach:
Screenshot 2024-07-17 at 14 43 05

kdetry:
> The accuracy and avg stake on the predictoor row doesn't mach the values from the metrics. I have all the feeds selected for the predictoor so they should mach: Screenshot 2024-07-17 at 14 43 05

```python
for predictor, feed in product(predictoors, feeds):
        # only filter for this particular predictoor and feed pair
        # in order to properly group the data
        filtered_payouts = [
            p for p in payouts if predictor in p["ID"] and feed.contract in p["ID"]
        ]

        slots, accuracies, profits, stakes = process_payouts(filtered_payouts)

        if not slots:
            continue

        avg_stake = ((stakes[-1] + avg_stake) / 2) if avg_stake else stakes[-1]
        total_profit = (profits[-1] + total_profit) if total_profit else profits[-1]
        avg_accuracy = (
            ((accuracies[-1] + avg_accuracy) / 2) if avg_accuracy else accuracies[-1]
        )

``` 

In the code block above, you averaged the accuracy value calculated for each pair and predictor. In my example, I calculate the average of all payouts of the predictor.

KatunaNorbert:
I see, so it should be fixed in the metrics calculation then

KatunaNorbert (approved):
LGTM! Thanks for also fixing the metrics issue

PR #1401, opened 2024-07-16
[Fix #1400] Fix selections in feed and predictoor list

### LINKED ISSUE #1400: [PdrDashboard, v0.2] Fix table selected row indexing
**Issue:**
When select a table row on any of the tables and then you do a search, some issues appear:
- If a select a row which is the 10th element for example then do a search and filter the table down to 2 element I get an error. Looks like app tries to get the index 10 but only 2 elements are now in the list.
- If I do the search first, select the element and remove the text from search box the wrong row is selected.

Looks like it's an indexing problem.

**Solutions:**
- Easiest one is to remove the selected rows each time a search occurs, and have the user select again
- Best solution is to make the changes necessary so whatever filtering and searching happens the selected rows remain selected and there are no issues

calina-c:
Closes #1400.
Modifies filtering s.t. it retains already checked boxes.

### DIFF
--- a/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
+++ b/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
@@ -120,18 +120,34 @@ def get_callbacks(app):
 
     @app.callback(
         Output("predictoors_table", "data"),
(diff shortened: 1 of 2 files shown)


KatunaNorbert (approved):
Looks good!

PR #1405, opened 2024-07-16
Fix #1404: Allow date period selection

### LINKED ISSUE #1404: [PdrDashboard, v0.2] Add buttons for date period selection
**Target**
Allow sectioning the data by different defined periods: 1D, 1Week, 1Month etc.

**To do**
- [ ] Add buttons for selecting different time periods
- [ ] When one button is clicked make it selected and update plots and metric with specific period data
- [ ] Display max available date range bellow the buttons

KatunaNorbert:
Fixes #1404


kdetry (approved):
Later, we have to add its tests to the test_callback by using dash_duo

PR #1411, opened 2024-07-17
[Fix #1409] Add predictoor and feed startup data instead of loading on a dash component

### LINKED ISSUE #1409: [PdrDashboard, v0.2] Move feeds and predictoors data from store to app
**Target**
Move static feeds and predictoors data to be fetched and stored before the app starts instead of storing it to state to make callbacks cleaner

To do:
- [x] fetch feeds and predictoors data before the app is started and store inside the app
- [x] update all the callbacks inside the app to not use these states as arguments

calina-c:
Closes #1409

### DIFF
--- a/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
+++ b/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
@@ -28,16 +28,15 @@ from pdr_backend.cli.arg_feeds import ArgFeeds
 def get_callbacks(app):
     @app.callback(
         [
-            Output("error-message", "children"),
             Output("show-favourite-addresses", "value"),
             Output("is-loading", "value"),
         ],
         Input("is-loading", "value"),
     )
     # pylint: disable=unused-argument
-    def get_input_data_from_db(is_loading):
+    def startup(is_loading):
         show_favourite_addresses = True if app.favourite_addresses else []
-        return None, show_favourite_addresses, 0
+        return show_favourite_addresses, 0
 
     @app.callback(
         Output("accuracy_chart", "children"),
--- a/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
+++ b/pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
@@ -1,20 +1,19 @@
-from dash import Input, Output, State
+import copy
+
 import dash
-from pdr_backend.analytics.predictoor_dashboard.dash_components.util import (
-    get_feeds_data_from_db,
-    get_predictoors_data_from_db,
-    get_payouts_from_db,
-    filter_objects_by_field,
-    get_feed_ids_based_on_predictoors_from_db,
-)
-from pdr_backend.analytics.predictoor_dashboard.dash_components.view_elements import (
-    get_graph,
-)
+from dash import Input, Output, State
+
 from pdr_backend.analytics.predictoor_dashboard.dash_components.plots import (
     get_figures_and_metrics,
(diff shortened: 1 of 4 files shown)


pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
```diff
@@ -1,58 +1,43 @@
...
 def get_callbacks(app):
     @app.callback(
-        Output("feeds-data", "data"),
-        Output("predictoors-data", "data"),
-        Output("error-message", "children"),
-        Output("show-favourite-addresses", "value"),
-        Output("is-loading", "value"),
-        Input("data-folder", "data"),
+        [
+            Output("error-message", "children"),
```
KatunaNorbert, line 31: Looks like this error message output is alway going to be `None` now, so let's remove it.
Following up on this, what happens if there is no data in the lake, or something wrong with fetching the data, is the error displayed somewhere?

pdr_backend/analytics/predictoor_dashboard/dash_components/view_elements.py
```diff
@@ -129,7 +129,6 @@ def get_metric(label, value, value_id):
 def get_layout():
     return html.Div(
         [
-            dcc.Store(id="data-folder"),
             dcc.Store(id="feeds-data"),
             dcc.Store(id="predictoors-data"),
```
KatunaNorbert, line 133: I think this 2 should also be removed now that the data is stored inside the app

pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
```diff
@@ -1,58 +1,43 @@
...
+            Output("show-favourite-addresses", "value"),
+            Output("is-loading", "value"),
+        ],
+        Input("is-loading", "value"),
     )
-    def get_input_data_from_db(files_dir):
+    # pylint: disable=unused-argument
+    def get_input_data_from_db(is_loading):
```
KatunaNorbert, line 38: Let's rename this method, there is no longer data loaded from db inside it

pdr_backend/analytics/predictoor_dashboard/dash_components/callbacks.py
calina-c: I can display the error on command startup. I think that's in line with what we have already in CLI

KatunaNorbert: Maybe you can set the children of the error-message component before the app starts with the error if there is any

pdr_backend/analytics/predictoor_dashboard/predictoor_dash.py
```diff
@@ -24,6 +29,18 @@ def predictoor_dash(ppss: PPSS, debug_mode: bool):
...
     folder = ppss.lake_ss.lake_dir
-    app.layout.children[0].data = folder
+    app.lake_dir = folder
+    try:
+        app.feeds_data = get_feeds_data_from_db(folder)
+        app.predictoors_data = get_predictoors_data_from_payouts(
+            get_user_payouts_stats_from_db(folder)
+        )
+    except Exception as e:
+        print(
```
KatunaNorbert, line 39: Shouldn't we use the `logger` here?

calina-c: no, it's a CLI print when running the command, not a debugging print.

KatunaNorbert (approved):

PR #1417, opened 2024-07-18
Fix #1413: Update and split dash_duo tests

### LINKED ISSUE #1413: [PdrDashboard, v0.2] Update and split dash_duo tests
***Epic:***
https://github.com/oceanprotocol/pdr-backend/issues/1357

**Target:**
Enhance the predictoor dashboard tests to make it more tested

**Objective:**
- [ ] To update and split the existing dash_duo tests to accommodate the new features and improvements in the predictoor dashboard.

kdetry:
Fixes #1413

### DIFF
--- a/pdr_backend/analytics/predictoor_dashboard/test/conftest.py
+++ b/pdr_backend/analytics/predictoor_dashboard/test/conftest.py
@@ -1,8 +1,30 @@
 import pytest
+import dash_bootstrap_components as dbc
+
+from dash import Dash
 from selenium.webdriver.chrome.options import Options
 
 from pdr_backend.lake.payout import mock_payouts, mock_payouts_related_with_predictions
 from pdr_backend.lake.prediction import mock_daily_predictions, mock_first_predictions
+
+from pdr_backend.analytics.predictoor_dashboard.dash_components.callbacks import (
+    get_callbacks,
+)
+from pdr_backend.analytics.predictoor_dashboard.dash_components.view_elements import (
+    get_layout,
+)
+from pdr_backend.analytics.predictoor_dashboard.test.resources import (
+    _prepare_test_db,
+    _clear_test_db,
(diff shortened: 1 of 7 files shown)


pdr_backend/analytics/predictoor_dashboard/test/conftest.py
```diff
@@ -42,3 +64,65 @@ def pytest_setup_options():
...
+                {{%scripts%}}
+                {{%renderer%}}
+            </footer>
+        </body>
+    </html>
+    """
+
+    app.layout = get_layout()
+    app.lake_dir = ppss.lake_ss.lake_dir
+    app.feeds_data = get_feeds_data_from_db(ppss.lake_ss.lake_dir)
```
calina-c, line 120: to avoid possible inconsistencies with this part, let's create a function that takes app and ppss as parameters, called e.g. `setup_predictoor_dashboard` and add predictoors_data, feeds_data, lake_dir, favourite_addresses. Call this function here and where this code is duplicated. I wanted to do this myself at the last round of fixes but I missed it

calina-c: other dashboards have at most one line of setup, so that's why we just copy that setup in the tests, but here I think it's safer to do a function

calina-c (commented):
I think there are some more cases to cover.
- starting up with favourite addresses that have payouts -> should check all feeds by default and load payouts at startup.
- toggling the switches for "my predictoors" and "only feeds with payouts"
- date toggles should filter the data

I understand we are trying to split the tests, but we also need to figure out how to combine some of these to avoid long time waits. I.e. maybe a single test for toggles/switches that tries both inputs, as opposed to starting with a "virgin" setup every time.

KatunaNorbert (approved):

PR #1421, opened 2024-07-18
Fix #1001: [Lake][ETL] Implement incremental ETL pipeline

idiom-bytes:
## Pull Request Description
This PR incorporates all updates required to implement the incremental ETL pipeline found in [PR #1000](https://github.com/oceanprotocol/pdr-backend/pull/1100)

**What happened?**

branch 685 was deleted... I did not realize this PR was a branch from there, so the original PR was also closed....this PR rescues all the work and prepares all changes for merging.

PR #1423, opened 2024-07-19
Fix #1001: [Lake][ETL] Implement incremental ETL pipeline

### LINKED ISSUE #1001: [ETL] Incremental Pipeline

idiom-bytes:
## Pull Request Description
By the way of cherry-picking all changes found in [PR #1000](https://github.com/oceanprotocol/pdr-backend/pull/1100), this PR incorporates all updates required to implement the incremental ETL pipeline

**What happened?**
branch 685 was deleted... I did not realize this PR was a branch from there, so the original PR was also closed....this PR rescues all the work and prepares all changes for merging.

- [x] Reviewed [new DuckDB/CRUD functions](https://github.com/oceanprotocol/pdr-backend/pull/1421#discussion_r1683816037)


pdr_backend/lake/duckdb_data_store.py
```diff
@@ -159,6 +163,47 @@ def insert_from_csv(self, table_name: str, csv_folder_path: str):
...
+        @example:
+            move_table_data("temp_people", "people")
+        """
+
+        # Check if the table exists
+        move_table_query = self.get_query_move_table_data(from_table, to_table)
+        self.execute_sql(move_table_query)
+
+    @enforce_types
+    def update_data(self, df: pl.DataFrame, table_name: str, column_name: str):
```
calina-c, line 182: are these intentional additions? I remember removing something very similar, specifically update_data not being used anywhere. Let's discuss this after we clear if they are intentional additions or conflict artifacts.

calina-c: same for "move_table_data"

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -243,34 +288,79 @@ def query_data(self, query: str) -> Optional[pl.DataFrame]:
...
+        return f"""
+        DELETE FROM {drop_table_name}
+        WHERE ID IN (
+            SELECT DISTINCT ID 
+            FROM {ref_table_name}
+        );
+        """
+
+    @enforce_types
+    def drop_records_from_table_by_id(self, drop_table_name: str, ref_table_name: str):
```
calina-c, line 313: I think this function name is a bit non-descriptive. Maybe something like `drop_from_table_using_reference_table_ids`? Because otherwise I expect one of the parameters to be a list of ids.

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -243,34 +288,79 @@ def query_data(self, query: str) -> Optional[pl.DataFrame]:
...
+        @example:
+            drop_records_from_table_by_id("bronze_pdr_slots", "update_pdr_slots")
+        """
+        query = self.get_query_drop_records_from_table_by_id(
+            drop_table_name, ref_table_name
+        )
+        self.execute_sql(query)
+
+    @enforce_types
+    def get_query_move_table_data(self, from_table, to_table):
```
calina-c, line 328: let's use "source" and "destination"/"dest" instead of to/from.

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -243,34 +288,79 @@ def query_data(self, query: str) -> Optional[pl.DataFrame]:
...
 
         # Move the data from the temporary table to the permanent table
-        self.execute_sql(
-            f"INSERT INTO {permanent_table_name} SELECT * FROM {temp_table.fullname}"
+        return (
+            f"INSERT INTO {to_table.table_name} SELECT * FROM {from_table.table_name};"
         )
+
+    @enforce_types
+    def row_count(self, table_name: str) -> int:
```
calina-c, line 354: I think row_count is also one of the things we removed in main. See above for update_data.

pdr_backend/lake/etl.py
```diff
@@ -56,27 +77,64 @@ def _drop_temp_sql_tables(self):
...
-            table_name = dataclass.get_lake_table_name()
-            logger.info("move table %s to live", table_name)
-            temp_table = TempTable(table_name)
-            permanent_table = NamedTable(table_name)
 
-            if db.table_exists(temp_table.fullname):
-                db.move_table_data(temp_table, permanent_table)
+        bronze_tables = [BronzePrediction]
+        for table in bronze_tables:
+            prod_table = Table.from_dataclass(table)
```
calina-c, line 96: I think some of the contents of this for can be on the Bronze Table class itself. The object should know how to move itself into the next step.

calina-c: You can add a mixin on all Bronze tables that abstracts this functionality.

pdr_backend/lake/etl.py
```diff
@@ -193,12 +251,9 @@ def get_timestamp_values(
         self, table_names: List[str], default_timestr: str
     ) -> UnixTimeMs:
         max_timestamp_values = self._get_max_timestamp_values_from(
-            [NamedTable(tb) for tb in table_names]
+            [Table(tb) for tb in table_names]
         )
 
-        logger.info(
```
calina-c, line 199: why was this logging line removed?

pdr_backend/lake/etl.py
```diff
@@ -241,78 +302,31 @@ def _calc_bronze_start_end_ts(self) -> Tuple[UnixTimeMs, UnixTimeMs]:
...
+        db = DuckDBDataStore(self.ppss.lake_ss.lake_dir)
 
-        for dataclass in _ETL_REGISTERED_LAKE_TABLES:
-            etl_func = dataclass.get_fetch_function()
-            etl_func(
-                path=self.ppss.lake_ss.lake_dir,
-                st_ms=st_timestamp,
-                fin_ms=fin_timestamp,
+        # if st_ts or fin_ts != ppss, then we must have records somewhere
+        is_first_run: bool = self._clamp_checkpoints_to_ppss is True or (
```
calina-c, line 317: I think you can have a separate function adnotated with `@property` on the class named is_first_run, i.e. self.is_first_run. Much easier to read logically.

calina-c (changes requested):
reviewed up until (including) lake/etl.py. I'll resume later.

pdr_backend/lake/gql_data_factory.py
```diff
@@ -69,10 +69,34 @@ def __init__(self, ppss: PPSS):
             }
         }
 
-    def _prepare_temp_table(self, dataclass: Type[LakeMapper], st_ut, fin_ut):
+    @enforce_types
+    def _calc_start_ut(self, table: Table) -> UnixTimeMs:
```
calina-c, line 73: I think this can also be a property of the table, e.g. `etl_start_ut`

calina-c: same for what involves end_ut

pdr_backend/lake/table.py
```diff
@@ -78,20 +80,24 @@ def drop_tables_from_st(db: DuckDBDataStore, type_filter: str, st: UnixTimeMs):
...
+class Table:
+    def __init__(self, table_name: str, table_type: TableType = TableType.PROD):
         self.table_type = table_type
+        self._base_table_name = table_name
         self._dataclass: Union[None, Type["LakeMapper"]] = None
 
     @property
-    def fullname(self) -> str:
+    def table_name(self) -> str:
+        if self.table_type == TableType.NEW_EVENTS:
```
calina-c, line 91: let's make this a loop :)

pdr_backend/lake_info/overview.py
```diff
@@ -288,3 +290,52 @@ def validate_lake_tables_no_duplicates(self) -> List[str]:
...
+                WHERE bronze_predictions.ID IS NULL
+                AND payouts.timestamp BETWEEN bronze_predictions.min_timestamp AND bronze_predictions.max_timestamp
+            )
+            select * from unmatched_payouts;
+        """
+
+        rows_df: pl.DataFrame = self.db.query_data(query_unmatched_payouts)
+
+        if rows_df is None or rows_df.shape[0] == 0:
+            logger.info("No unmatched payouts found in the lake.")
```
calina-c, line 337: I think you also need to add the violation into the list

ppss.yaml
```diff
@@ -10,7 +10,7 @@ lake_ss:
     - binance BTC/USDT ETH/USDT 5m
   #    - binance BTC/USDT ETH/USDT BNB/USDT XRP/USDT ADA/USDT DOGE/USDT SOL/USDT LTC/USDT TRX/USDT DOT/USDT 5m
   #    - kraken BTC/USDT 5m
-  st_timestr: 30 days ago # starting date for data
```
calina-c, line 13: is this intentional or was it supposed to happen in my_ppss.yaml instead?

pdr_backend/lake_info/overview.py
calina-c: I'm also a bit confused since the lake validate command was removed from the help string, so do we only do validations in describe now?

calina-c (dismissed):

pdr_backend/lake/duckdb_data_store.py
idiom-bytes: yeah, they are intentional implementations
- update_data is used for incremental updates as we want to modify existing records
- same with move_table_data, it's used in the swap strategy

I'll review the CRUD wrapper to make sure the functionality isn't duplicated

idiom-bytes: yes, these are intentional additions

"remember removing something very similar, specifically update_data not being used anywhere "
can you show me? perhaps this was added ahead of time of time, and even I myself removed it as it was not being used by any update routine... now it is.

"same for "move_table_data"
same statement as above. it's now being used to do an update/swap routine.

pdr_backend/lake/duckdb_data_store.py
idiom-bytes: I have renamed it from `drop_from_table_using_reference_table_ids` -> `drop_common_records_by_id`

However `drop_common_records_by_id` isn't actually being used, as I deprecated it's usage and instead build the query so they whole swap strategy can be executed atomically.

image

So, now we only have: `get_query_drop_common_records_by_id" and "get_query_move_table_data"

pdr_backend/lake/duckdb_data_store.py
idiom-bytes: why is that?

"from_" and "to_" is common nomenclature throughout the whole engine... "source_" and "destination_" isn;t.

not only are they longer, but not common practice in the rest of the code base

pdr_backend/lake/etl.py
idiom-bytes: because I am a heavy user of the lake CLI and the lake CLI is incomprehensible

it generates a ton of repeating/redundant/granular logs that bloats the output

so i removed a bunch of it and made sure that you can still follow all core logic.

pdr_backend/lake/duckdb_data_store.py
idiom-bytes: I have specifically added these things for tooling and validations around the CLI... these operations are much faster done through the DB. 

Please operate the lake, become a heavy user, debug incremental pipelines, and then make decisions as to what functionality to drop.

idiom-bytes:
**To everyone. On DX and "cleaning up":** please operate the lake, become a heavy user, debug incremental pipelines, engineer/operarte new features, and then lets discuss what functionality/requirements need to be updated/changed/drop.

pdr_backend/lake/etl.py
idiom-bytes: tagging @trentmc here as I'm concerned that this could lead to more complex problems...

No. **Please separate the concerns of the ETL pipeline from the Table.**

Read the readme...
Tables know nothing about ETL.
This is done very intentionally...

**The ETL may eventually be executed in a SQL cluster, or a distributed MPP pipeline. None of this should exist inside of "BronzeTable". Bronze table is just a wrapper for table schema/validation/etc...**

Please be more considerate of **LoB** - Locality of Behavior... The ETL is responsible for bronze/silver/gold/ ETL processes/swap/etc... Tables should know nothing about this...

image

pdr_backend/lake/etl.py
idiom-bytes: 1. but it's not a "property" or a class variable... it's a check made before the ETL execution begins.
2. the other member variable _clamp_checkpoints_to_ppss, 
in addition to other checks (that are expensive) must be called in order to calculate this condition

perhaps i can move this somewhere else that makes more sense...

idiom-bytes: thought about it a bit more...
1. `is_first_run` is specifically being calculated for the bronze_tables, based on `bronze_table` ETL requirements...
2. `silver_tables` and `gold_tables` may have other `is_first_run` logic/checks...

this doesn't make sense to generalize

pdr_backend/lake/gql_data_factory.py
idiom-bytes: please read my comments/feedback

we should not add this logic to table...

pdr_backend/lake/table.py
idiom-bytes: done!

pdr_backend/lake_info/overview.py
idiom-bytes: "I think you also need to add the violation into the list"
Nice catch!

idiom-bytes: "I'm also a bit confused since the lake validate command was removed from the help string, so do we only do validations in describe now?"

This was removed with your DOT removal PR... I'll re-introduce it.

image

idiom-bytes: "I'm also a bit confused since the lake validate command was removed from the help string..."

Part was removed with your DOT removal PR... I don't know when the rest was removed... I'll reintroduce them.

image

idiom-bytes: reintroduced helper text and reorganized helper functions so everything is in logical/onboarding order

ppss.yaml
idiom-bytes: Mistake, nice catch!

Resolved.

pdr_backend/lake/etl.py
calina-c: I would recommend separating the condition itself into a function. Call it is_first_bronze run if necessary, but it looks like a bunch of conditions that should have their own place and allow the reader of the code to quickly glimpse within a line "it is the first run", not get tangled in the conditionals. If needed, they can look at the function for details.

idiom-bytes:
```
but it looks like a bunch of conditions that should have their own place and allow the reader of the code to quickly glimpse within a line "it is the first run", not get tangled in the conditionals
```

(1) this check/logic is very specific to this routine
(2) it's an extremely simple condition
(3) every other process that can be abstracted and benefits from it, has been
(4) i'd like to expand on silver, gold, and other ETL workflows before abstracting things further

```
If needed, they can look at the function for details
```
abstracting everything also forces users to follow 5 threads to stitch the whole thing together

can we please expand on the system (silver, gold, other ETL) before trying to abstract everything?

idiom-bytes: done

calina-c:
I can't tell which comments are new and which old. I trust that you resolved only comments that you addressed. I would also recommend a second look from another dev on the team, just to be sure. Otherwise LGTM.

idiom-bytes:
"I can't tell which comments are new and which old. I trust that you resolved only comments that you addressed. I would also recommend a second look from another dev on the team, just to be sure. Otherwise LGTM."

I have unresolved the conversation on all items where I didn't implement any change.

Either the comments do not apply, or seemed subjective enough where I strongly disagree due to architectural or engineering related considerations.

KatunaNorbert:
Tested is and I might found an issue. The `trueval` field is NULL for all the rows inside the `bronze_pdr_predictions` table

idiom-bytes:
```
Tested is and I might found an issue. The trueval field is NULL for all the rows inside the bronze_pdr_predictions table
```

You can't expect trueval to work correctly until it's implemented in the Incremental ETL.

Even now, trueval will be NULL for new predictions that are arriving.

Basically, lake only works with historical data... As soon as you start fetching new data, predictions have null truevals, and they won't be updated correctly.

**The error is expecting truevalue field to be set for predictions.**

idiom-bytes:
You did find an issue. Nice work.

truevalue had never been implemented from the subgraph fetch, so it was missing from payout.
image

it's now being fetched from subgraph -> saved to csv -> processed in ETL -> updated in bronze table

I have updated the ETL tests to validate that it's working

KatunaNorbert (approved):

idiom-bytes:
Answering trizin's commenst in here

``slot_id is literally {contract}-{slot}, it's redundant```
I have a PR to work update slot-related queries, i can implement it there -> https://github.com/oceanprotocol/pdr-backend/pull/1466

`slot_id` will be required for certain joins. Having this pre-processed (string split) and then available for joins (string equal vs. string like operation) will save computation time when it's needed most.

```pair, timeframe, source: duplicated data, should be derived from contract address```
100% agree, I've wanted to do this for a while. i'm working towards this and added this to the ETL & Analytics backlog.

```
remove PDR_PREDICTIONS and PDR_PAYOUTS, just a single table BRONZE_PDR_PREDICTIONS should be enough
```
This is the goal (to only have bronze), this would also reduce DB size. The challenge is that we need to build all the bronze tables before removing all raw_data from DB.

Now that we have Incremental ETL, this should be possible.

**I'm proposing we deprecate raw tables once we have more bronze tables in place, and start working towards silver (aggregate) tables**

**Challenge:**
If raw tables aren't in the DB, then any ETL work is going to be slow (scanning + reading CSVs into memory for processing)
If raw tables are in DB, then DuckDB is very fast and we can do queries/aggregates/summaries efficiently.

idiom-bytes:
We had multiple reviewers, discussed feedback, and agreed on how to move forward across the board. Merging.

PR #1440, opened 2024-07-23
Fix #1434: Show the confidence interval on the Accuracy plot

kdetry:
Fixes #1434

PR #1444, opened 2024-07-23
Fix #1438: Calculate and display tx fee costs

### LINKED ISSUE #1438: [PdrDashboard, v0.3] Display prediction tx costs
**Target**
Let predictoors know approximately how much they spend on transaction fees.

**To do**
- [ ] calculate the gas fees costs by looking at the cost of 1 prediction tx cost and multiply it by an avg gas price value
- [ ] create a new costs plot and 'rename' the current plot to ‘stakes'
- [ ] display total costs on top along side the other

KatunaNorbert:
Fixes #1438


pdr_backend/pdr_dashboard/dash_components/plots.py
```diff
@@ -99,16 +103,25 @@ def _empty_profit_scatter() -> List[go.Scatter]:
...
+    return [go.Scatter(x=[], y=[], name="costs", width=5)]
+
+
 def _empty_stakes_bar() -> List[go.Bar]:
     return [go.Bar(x=[], y=[], name="stakes", width=5)]
 
 
-def _empty_trio() -> Tuple[go.Figure, go.Figure, go.Figure]:
-    return _empty_accuracy_scatter(), _empty_profit_scatter(), _empty_stakes_bar()
+def _empty_trio() -> Tuple[go.Figure, go.Figure, go.Figure, go.Figure]:
```
calina-c, line 114: it's not a trio anymore, though :)

pdr_backend/pdr_dashboard/dash_components/util.py
```diff
@@ -207,3 +209,32 @@ def get_date_period_text(payouts: List):
         - {datetime.fromtimestamp(end_date).strftime('%d-%m-%Y')}
     """
     return date_period_text
+
+
+def calculate_gass_fee_costs(web3_pp, feed_contract_addr, prices):
```
calina-c, line 214: I think it should be "gas" instead of "gass"

calina-c:
btw @KatunaNorbert, @kdetry is working on a solution to simplify the tuples into named tuples here: https://github.com/oceanprotocol/pdr-backend/pull/1440. I suggest we integrate that first to make it easier for you here as well.

pdr_backend/pdr_dashboard/dash_components/plots.py
idiom-bytes: i have no idea what a trio is (i do) but the naming has to change if the parameters ever do

kiss... just name it tuple

pdr_backend/pdr_dashboard/dash_components/view_elements.py
```diff
@@ -185,9 +185,10 @@ def get_graphs_column():
 def get_graphs_column_metrics_row():
     return html.Div(
         [
-            get_metric(label="Avg Accuracy", value="50%", value_id="accuracy_metric"),
-            get_metric(label="Total Profit", value="50%", value_id="profit_metric"),
-            get_metric(label="Avg Stake", value="50%", value_id="stake_metric"),
+            get_metric(label="Avg Accuracy", value="0%", value_id="accuracy_metric"),
+            get_metric(label="Total Profit", value="0 OCEAN", value_id="profit_metric"),
+            get_metric(label="Tx Costs", value="0 OCEAN%", value_id="costs_metric"),
```
kdetry, line 190: I guess we do not need the "%" char here

pdr_backend/pdr_dashboard/dash_components/view_elements.py
```diff
@@ -223,7 +224,17 @@ def get_graphs_column_plots_row():
...
             html.Div(id="accuracy_chart"),
             html.Div(id="profit_chart"),
-            html.Div(id="stake_chart"),
+            html.Div(
+                [
+                    html.Div(id="cost_chart", style={"width": "48%"}),
+                    html.Div(id="stake_chart", style={"width": "48%"}),
+                ],
+                style={
+                    "width": "100%",
```
kdetry, line 233: Just a note, we can move the styles to the style.css for HTML tags later. Unfortunately, the className property does not exist in datatables.

kdetry (changes requested):

pdr_backend/pdr_dashboard/dash_components/view_elements.py
KatunaNorbert: For that, we would need to define a class or an id

calina-c (approved):

kdetry (approved):

PR #1475, opened 2024-07-30
Fix #1473:  Add navigation and create feeds page

### LINKED ISSUE #1473: [PdrDashboard] v0.4: Feeds page
### Target

Add a new page dedicated to feeds, where users can leverage a big table containing all the existing feeds with multiple metrics and properties.

Super epic issue can be found [here](https://github.com/oceanprotocol/pdr-backend/issues/1328)

### TODO
- [x] Add navigation and transform dapp into multipage application [#1474](https://github.com/oceanprotocol/pdr-backend/issues/1474) - Norbert
    - Add Home and Feeds navigations
    - Replace current dashboard title with navigation
    - When clicking on home redirect to current predictoor dashboard, on Feeds show the new Feeds page
- [x] #1485
    - Under the menu display a row containing main metrics over all feeds, such as: Volume, Feeds, Sales, Revenue, Accuracy(avg)
- [x] Create a table that contains all the columns from the mockup provided bellow  [#1478](https://github.com/oceanprotocol/pdr-backend/issues/1478) - Mustafa
- [x] Create a filters section that contains multiple components for different columns and allow table filtering [#1484](https://github.com/oceanprotocol/pdr-backend/issues/1484) - Norbert
    - The components should be compressed and the dropdown list should open when component is clicked
    - Add a Reset All button then clears all the filter values
- [ ] Add search bar that allow search based on: Add, Base Token, Quote token, Exchange [#1520](https://github.com/oceanprotocol/pdr-backend/issues/1520) - Norbert
- [x] Expandable rows [#1506](https://github.com/oceanprotocol/pdr-backend/issues/1506) - Calina / Norbert
    - when clicking on a specific row, the address should be saved and a component should open
    - when clicking again the component should close
- [x] Create and display data inside expanded row [#1506](https://github.com/oceanprotocol/pdr-backend/issues/1506)  - Norbert
    - display following charts: Volume, Sales, Accuracy, Predictions, Stakes over time
- [ ] Dash_duo / selenium tests for the new page
- [x] #1491 split up utils.py
- [x] create a yaml file for tooltip texts
- [x] Create and implement NumberFormatter [#1501](https://github.com/oceanprotocol/pdr-backend/issues/1501) - Mustafa
- [x] Find and add the websocket buyer address to the DF buyers, only for this module [#1508](https://github.com/oceanprotocol/pdr-backend/issues/1508) - Mustafa
- [x] Fix the table title and do nor remove the "Avg" phrase  [#1501](https://github.com/oceanprotocol/pdr-backend/issues/1501) - Mustafa
- [x] Fix the legend on the home page and remove the exchange name from the legend items [#1515](https://github.com/oceanprotocol/pdr-backend/issues/1515) - Mustafa

### UI sketch for v0.4:
- inspiration from [DefiLlama](https://defillama.com/yields)
Screenshot 2024-07-29 at 16 20 55

Screenshot 2024-07-29 at 16 22 40

KatunaNorbert:
Fixes #1473

### DIFF
--- a/pdr_backend/pdr_dashboard/pages/home.py
+++ b/pdr_backend/pdr_dashboard/pages/home.py
@@ -1,0 +1,18 @@
+from dash import html, dcc
+import dash_bootstrap_components as dbc
+from pdr_backend.pdr_dashboard.dash_components.view_elements import get_main_container
+
+
+def layout(app):
+    return html.Div(
+        [
+            dcc.Store(id="user-payout-stats"),
+            dcc.Loading(
+                id="loading",
+                type="default",
+                children=get_main_container(app),
+                custom_spinner=html.H2(dbc.Spinner(), style={"height": "100%"}),
+            ),
+        ],
+        style={"height": "100%"},
+    )
--- a/pdr_backend/pdr_dashboard/predictoor_dash.py
+++ b/pdr_backend/pdr_dashboard/predictoor_dash.py
@@ -38,7 +38,6 @@ def predictoor_dash(ppss: PPSS, debug_mode: bool):
             Full error: {e}"""
         )
         return
-
     get_callbacks(app)
     app.run(debug=debug_mode, port=port)
 
@@ -52,7 +51,7 @@ def setup_app(app, ppss: PPSS):
         get_user_payouts_stats_from_db(ppss.lake_ss.lake_dir)
     )
     app.favourite_addresses = ppss.predictoor_ss.my_addresses
-    app.layout = get_layout(app)
+    app.layout = get_layout()
 
     # fetch token prices
     current_date_ms = UnixTimeMs(int(datetime.now().timestamp()) * 1000 - 300000)
--- a/pdr_backend/pdr_dashboard/assets/styles.css
+++ b/pdr_backend/pdr_dashboard/assets/styles.css
@@ -115,3 +115,7 @@ button{
     border: 0px none;
     margin-left: 5px;
 }
+
+.ms-auto{
+    margin: 0 auto !important;
+}
--- a/pdr_backend/pdr_dashboard/dash_components/callbacks.py
+++ b/pdr_backend/pdr_dashboard/dash_components/callbacks.py
@@ -17,13 +17,31 @@ from pdr_backend.pdr_dashboard.dash_components.util import (
 )
 from pdr_backend.pdr_dashboard.dash_components.view_elements import (
     get_graph,
+    get_navbar,
+    NAV_ITEMS,
 )
+
+from pdr_backend.pdr_dashboard.pages import home, feeds
 
 from pdr_backend.cli.arg_feeds import ArgFeeds
 
 
 # pylint: disable=too-many-statements
 def get_callbacks(app):
+    @app.callback(
+        Output("page-content", "children"),
+        Output("navbar-container", "children"),
+        [Input("url", "pathname")],
+    )
+    def display_page(pathname):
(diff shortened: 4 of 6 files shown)


pdr_backend/pdr_dashboard/pages/home.py
```diff
@@ -0,0 +1,158 @@
...
+    get_predictoors_data,
+    get_metric,
+    get_date_period_selection_component,
+)
+from pdr_backend.pdr_dashboard.dash_components.util import (
+    get_feed_ids_based_on_predictoors_from_db,
+)
+
+
+def layout(app):
```
calina-c, line 15: I would suggest converting all of these to a class, now. Since we evolved the knowledge required by each component, I would suggest creating Page/AppPage objects which take in `app` in the constructor. Functions that use knowledge from app or any other common structures can therefore be independent and parameter passing reduced. We can define a light interface using abc that requires all pages to have a layout() function.

In time, the procedural approach will get too complicated to manage so it's better to do this now.

kdetry (approved):
We can deal with its style later

pdr_backend/pdr_dashboard/pages/home.py
```diff
@@ -0,0 +1,158 @@
...
+    )
+
+
+def get_input_column(app):
+    feed_cols, feed_data = get_feeds_data(app)
+    predictoor_cols, predictoor_data = get_predictoors_data(app)
+
+    selected_predictoors = list(range(len(app.favourite_addresses)))
+
+    if app.favourite_addresses:
```
calina-c, line 58: I know this is my work, but seeing it moved helped me realise we can simplify the ifs a bit. If we move app as a class attribute and it's easier to separate, let's retrieve the selected feeds in a separate function and keep the functional part away from the layout part. i.e. have a function "get_feeds_for_predictoors" instead of this block here (58-71)

pdr_backend/pdr_dashboard/pages/home.py
```diff
@@ -0,0 +1,158 @@
...
+            ]
+
+        selected_feeds = list(range(len(feed_ids)))
+    else:
+        selected_feeds = []
+
+    return html.Div(
+        [
+            html.Div(
+                get_table(
```
calina-c, line 76: I also think by this point we can have get_predictoors_table and get_feeds_table. The "get_table" function used to be simpler buy now it's reached 7 parameters and is almost a pass-through to the table itself

pdr_backend/pdr_dashboard/predictoor_dash.py
```diff
@@ -38,8 +41,8 @@ def predictoor_dash(ppss: PPSS, debug_mode: bool):
             Full error: {e}"""
         )
         return
-
-    get_callbacks(app)
+    get_callbacks_home(app)
```
calina-c, line 44: having a Page-like OOP structure would also help in these cases here. Maybe we can have callbacks autoregistered by each page instead of calling these functions. This comment you can ignore, if you think it's too "science-fiction" :)

pdr_backend/pdr_dashboard/pages/home.py
KatunaNorbert: In my opinion, the `get_table` method has too many components and it's worth generalizing them rather than repeating them, it also contains other components beside the table, like inputs

KatunaNorbert:
Do you think a Class is needed?
It will probably have 1 or 2 methods and there will most likely be just 3 pages

PR #1479, opened 2024-07-30
[aimodelss] Define Seed value for models

### LINKED ISSUE #1477: [AimodelSS] Define seed value for models
Feature to define seed value for models within config yaml

trizin:
Fixes #1477

- aimodels now supports a `seed` parameter for reproducibility.


idiom-bytes:
i reviewed this and it looks ok (i.e. in ai_model_ss it's either the seed or None, which gets passed to the rest of the system)

but, tests are failing...

idiom-bytes (approved):
i reviewed this and it looks ok...

ai_model_ss defines the seed, or it's None
value is then passed to the rest of the system

but, tests are failing...

PR #1482, opened 2024-07-30
[Fix #1481] PdrDashboard v0.4 Improve multipage DRY

### LINKED ISSUE #1481: [PdrDashboard, v0.4] Fix DRY and OOP with new multipage application structure
A number of DRY and OOP concerns have appeared due to changes in v0.4 from the one-page to multi-page Dash structure. These need to be addressed before they become a problem.

calina-c:
Closes #1481

### DIFF
--- a/pdr_backend/pdr_dashboard/predictoor_dash.py
+++ b/pdr_backend/pdr_dashboard/predictoor_dash.py
@@ -41,8 +41,7 @@ def predictoor_dash(ppss: PPSS, debug_mode: bool):
             Full error: {e}"""
         )
         return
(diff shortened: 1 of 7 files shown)


pdr_backend/pdr_dashboard/predictoor_dash.py
kdetry, line 1: I know it's not related to the context, but I'm suggesting this because it's a refactoring PR. We could move the logic to fetch token prices to a utility function and we can use it in the setup_app function.

The code part which starts with the "# fetch token prices" comment line

KatunaNorbert (approved):

PR #1487, opened 2024-07-31
Fix #1478: The Feeds Table 

### LINKED ISSUE #1478: [PdrDashboard, v0.4] Feeds Table
Create a table that contains all the columns from the mockup provided below

image

kdetry:
Fixes #1478


KatunaNorbert:
I think there are a couple of improvements missing:
- add measurement units where there are numeric values: (OCEAN) for (Stake, Volume) and (%) for Accuracy
- align the first column to the left (maybe the next 3 as well)
- make table scrollable. After we add more components, the table won't fit in the view
- make a little more space inside the cells, maybe by adding some padding. The values look too crowded

kdetry:
> I think there are a couple of improvements missing:
> 
> * add measurement units where there are numeric values: (OCEAN) for (Stake, Volume) and (%) for Accuracy
> * align the first column to the left (maybe the next 3 as well)
> * make table scrollable. After we add more components, the table won't fit in the view
> * make a little more space inside the cells, maybe by adding some padding. The values look too crowded

Thank you for the feedback, I will work on them and update the PR

pdr_backend/pdr_dashboard/pages/feeds.py
```diff
@@ -1,22 +1,128 @@
...
+
+            feed_item["avg_accuracy"] = str(result[0]) + "%" if result else ""
+            feed_item["avg_stake_(OCEAN)"] = result[2] if result else ""
+            feed_item["volume_(OCEAN)"] = result[1] if result else ""
+
+            subscription_result = find_with_key_value(
+                feed_subcription_stats, "contract", feed["contract"]
+            )
+
+            feed_item["price"] = (
```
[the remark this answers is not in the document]
kdetry: done 👍

KatunaNorbert (approved):

pdr_backend/pdr_dashboard/pages/feeds.py
```diff
@@ -1,22 +1,128 @@
...
+        for feed in temp_data:
+            split_pair = feed["pair"].split("/")
+            feed_item = {}
+            feed_item["addr"] = feed["contract"][:5] + "..." + feed["contract"][-5:]
+            feed_item["base_token"] = split_pair[0]
+            feed_item["quote_token"] = split_pair[1]
+            feed_item["exchange"] = feed["source"].capitalize()
+            feed_item["time"] = feed["timeframe"]
+
+            result = self.get_feeds_stat_with_contract(feed["contract"], feed_stats)
```
calina-c, line 98: I would suggest returning a dictionary here instead of a tuple, and simply updating the initial dictionary with the feed stat dictionary. Why rely on tuple orders if you can use the same data structure as the caller?

pdr_backend/pdr_dashboard/pages/feeds.py
```diff
@@ -1,22 +1,128 @@
...
+                    sort_action="native",  # Enables sorting feature
+                )
+            ],
+            style={"width": "100%"},
+        )
+
+    def get_feeds_stat_with_contract(
+        self, contract: str, feed_stats: List[Dict[str, Any]]
+    ) -> Union[Tuple[float, float, float], None]:
+        result = find_with_key_value(feed_stats, "contract", contract)
```
calina-c, line 68: I this particular case, why are we not using a query? Instead of fetching the entire feeds data and filtering it for a contract, wouldn't it make sense to select that feed directly? Maybe I understood the rest wrong, but the "find_with_key_value" function has a weird vibe to me. If it's really needed, I think we already have this logic somewhere with filters in pdr dashboard.

calina-c (changes requested):

pdr_backend/pdr_dashboard/pages/feeds.py
kdetry: Let me summarize what we do here:
1. Retrieve feed data from the prediction table by grouping contracts. * * *
2. Gather payout stats by grouping them with contracts from the payouts table.

Then, we merge them by selecting contracts on line 98 of feeds.py, so all result rows are used in the data:
`result = self.get_feeds_stat_with_contract(feed["contract"], feed_stats)`

I contemplated merging these queries, but there are some drawbacks:
- Firstly, as we already query the database for the feeds on the home screen (the first step that I marked above) , so we do not query again for the feeds page.
- Since there is no contract column on other tables, we split the ID to get the contract address. If we use the "JOIN" statement, it could lead to performance issues.

pdr_backend/pdr_dashboard/pages/feeds.py
kdetry: I fixed it, and also created another method for the subscription stats (`get_feeds_subscription_stat_with_contract`) and used in this method, for more clear code.

calina-c:
Discussed with @kdetry: I will help split up utils because too many mixed functionalities are caught up in there, leading to my thinking that certain dict utils were already in place, although they were just similar.

Concerns currently tied up in this same file:
- db functions
- dict manipulation stuff, col renaming etc...
- prices and costs

should be separated. While it is true that util.py is a single file for other dashboards and we would break the standard in that sense, Pdr Dashboard is insofar the most complex dashboard and needs to be split.

kdetry:
During our discussion, we identified a need to reorganize the utils.py file for better maintainability and clarity

so our list is below:
- Separate utils.py into specific files (db_utils, dict_utils)
- Create new file(s) for DB Methods and queries, such as db_methods.py, queries.py(optional)

calina-c (approved):

PR #1488, opened 2024-07-31
Fix #1484: Create filters section

### LINKED ISSUE #1484: [PdrDashboard, v0.4] Create inputs for filtering table
### Background / motivation

Add multiple dropdown inputs which offer an intuitive way of filtering table data

### TODOs / DoD

- [ ] add input components that should be compressed and the dropdown list should open when component is clicked
- [ ] add a a Reset All button then clears all the filter values

KatunaNorbert:
Fixes #1484


pdr_backend/pdr_dashboard/callbacks/callbacks_feeds.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def filter_condition(item, field, values):
+    if not values:
+        return True
+    return item[field] in values
+
+
+def range_condition(item, field, min_value, max_value):
+    if min_value is not None and min_value != "" and item[field] < min_value:
```
kdetry, line 21:
```
def range_condition(item, field, min_value, max_value):
    return not (
        (min_value is not None and min_value != "" and item[field] < min_value) or
        (max_value is not None and max_value != "" and item[field] > max_value)
    )

```

pdr_backend/pdr_dashboard/callbacks/callbacks_feeds.py
```diff
@@ -0,0 +1,166 @@
...
+    ctx = callback_context
+    button_id = ctx.triggered[0]["prop_id"].split(".")[0]
+    if (not min_val and not max_val) or button_id == "clear_filters_button":
+        return label_text
+
+    return f"{label_text} {min_val if min_val else ''}-{max_val if max_val else ''}"
+
+
+def filter_condition(item, field, values):
+    if not values:
```
kdetry, line 15:
```
def filter_condition(item, field, values):
    return not values or item[field] in values

```

pdr_backend/pdr_dashboard/callbacks/callbacks_feeds.py
```diff
@@ -0,0 +1,166 @@
...
+            item
+            for item in app.feeds_table_data
+            if filter_condition(item, "base_token", base_token)
+            and filter_condition(item, "quote_token", quote_token)
+            and filter_condition(item, "venue", venue)
+            and filter_condition(item, "time", time)
+            and range_condition(item, "sales", sales_min, sales_max)
+            and range_condition(item, "sales_revenue_(OCEAN)", revenue_min, revenue_max)
+            and range_condition(item, "avg_accuracy", accuracy_min, accuracy_max)
+            and range_condition(item, "volume_(OCEAN)", volume_min, volume_max)
```
kdetry, line 82: I feel like there are lots of "and" in this condition.
Can we do like that ?


```

conditions = [
            ("filter", "base_token", base_token),
            ("filter", "quote_token", quote_token),
            ("filter", "venue", venue),
            ("filter", "time", time),
            ("range", "sales", sales_min, sales_max),
            ("range", "sales_revenue_(OCEAN)", revenue_min, revenue_max),
            ("range", "avg_accuracy", accuracy_min, accuracy_max),
            ("range", "volume_(OCEAN)", volume_min, volume_max),
        ]

        def check_condition(item, condition_type, field, *values):
            if condition_type == "filter":
                return filter_condition(item, field, values[0])
            elif condition_type == "range":
                return range_condition(item, field, values[0], values[1])
            return True

        new_table_data = [
            item
            for item in app.feeds_table_data
            if all(check_condition(item, *condition) for condition in conditions)
        ]

        return new_table_data
```

kdetry (commented):
I suggested some code changes

pdr_backend/pdr_dashboard/callbacks/callbacks_feeds.py
KatunaNorbert: This approach gives errors for some reason

KatunaNorbert: Discovered it was something else

pdr_backend/pdr_dashboard/pages/feeds.py
```diff
@@ -31,6 +68,99 @@ def layout(self):
...
+                self.get_multiselect_dropdown(filter_obj)
+                for filter_obj in filters_objects
+            ]
+            + [
+                self.get_input_filter("Sales"),
+                self.get_input_filter("Revenue"),
+                self.get_input_filter("Accuracy"),
+                self.get_input_filter("Volume"),
+            ],
+            style={
```
calina-c, line 93: lt's add div styles to css as Mustafa proposed last time

pdr_backend/pdr_dashboard/pages/feeds.py
```diff
@@ -31,6 +68,99 @@ def layout(self):
...
+                    },
+                    id=f"{label.lower()}_button",
+                ),
+            ],
+            id=f"{label.lower()}_dropdown",
+            label=label,
+            style={
+                "backgroundColor": "white",
+            },
+            toggle_style={
```
calina-c, line 122: same here

calina-c: and in other places in this file

pdr_backend/pdr_dashboard/test/conftest.py
```diff
@@ -60,6 +60,7 @@ def pytest_setup_options():
     options = Options()
     options.add_argument("--headless")
     options.add_argument("--disable-gpu")
+    options.add_argument("--disable-search-engine-choice-screen")
```
calina-c, line 63: this is moot and confusing in general CIs. It it's headless, there's no search engine choice screen.
How about leaving some options commented, with an explanatory comment saying "these are useful when debugging"?

pdr_backend/pdr_dashboard/test/resources.py
```diff
@@ -53,6 +53,21 @@ def _input_action(dash_duo, input_id, table_id, input_value, expected_rows):
...
 
 
+def _select_dropdown_option(dash_duo, dropdown_id, option_text):
+    dropdown = dash_duo.find_element(dropdown_id)
+    dropdown.click()
+    options = dash_duo.find_elements(".VirtualizedSelectOption")
+    for option in options:
+        if option.text == option_text:
+            option.click()
+            break
```
calina-c, line 63: maybe add an exception if nothing is found, it should definitely break the tests instead of doing nothing.

pdr_backend/pdr_dashboard/test/resources.py
```diff
@@ -53,6 +53,21 @@ def _input_action(dash_duo, input_id, table_id, input_value, expected_rows):
...
+
+
+def _assert_table_row_count(dash_duo, expected_count):
+    table_rows = dash_duo.find_elements("#feeds_page_table tbody tr")
```
calina-c, line 67: I think the table id should pe parameterised

calina-c (changes requested):

pdr_backend/pdr_dashboard/test/conftest.py
KatunaNorbert: I don't see why this could have any bad effect on CIs, Imo it's better to have it in case anyone else encounters the issue

pdr_backend/pdr_dashboard/test/resources.py
KatunaNorbert: it will break the test on the next step if option is not found

pdr_backend/pdr_dashboard/callbacks/callbacks_feeds.py
kdetry: Thanks for implementation

kdetry (approved):
Thanks for implementing my suggestions

PR #1466, opened 2024-07-26
Fix #1446: [Lake] Reintroduce subscriptions, and slots

idiom-bytes:
## Description
Fixes #1446 

## Commit message
- [x] Updated GQLDF to include all tables (minus contracts)
- [x] Updated all logic to work w/ latest subgraph updates
- [x] Fixed tests
- [x] Updated sample data and fixtures for new test coverage
- [x] Updated** all ETL tests to work w/ latest fixture

PR #1489, opened 2024-07-31
Fix #1446: [Lake] Reintroduce subscriptions, and slots

### LINKED ISSUE #1446: [ETL] Reintroduce subscriptions and slots into ETL
**Is your feature request related to a problem? Please describe.**
New predictoor dashboard will require subscriptions... we'll likely be doing things with slots as well (and they already existed in GQL/lake previously)

**Describe the solution you'd like**
Re-introduce subscriptions and slots

**DoD:**
- [ ] raw subscriptions table + columns are available
- [ ] raw slots table + columns are available
- [ ] conftest/fixture has been updated to provide raw subscriptions sample data
- [ ] conftest/fixture has been updated to provide raw slots sample data
- [ ] sample data has been generated
- [ ] fixture tests and basic validations are in place

idiom-bytes:
## Description
Fixes #1446 

This PR "cherry-picks" changes from #1466 in order to clean up the PR/changes and make it easier to review

## Commit message
- [x] Updated GQLDF to include all tables (minus contracts)
- [x] Updated all logic to work w/ latest subgraph updates
- [x] Fixed tests
- [x] Updated sample data and fixtures for new test coverage
- [x] Updated** all ETL tests to work w/ latest fixture


pdr_backend/lake/slot.py
```diff
@@ -29,7 +29,6 @@ def __init__(
         self.truevalue = truevalue
         self.roundSumStakesUp = roundSumStakesUp
         self.roundSumStakes = roundSumStakes
```
KatunaNorbert, line 31: The Slot table is very similar with the Trueval table, we should probably combine this 2 tables just into 1 table called Slot.
We could still read Trueval events from the subgraph in order to update the Slot and other tables with the trueval, but I don't think there is a benefit in storing a Trueval table

PR #1492, opened 2024-08-02
[Fix #1490] PdrDashboard v0.4 Fixes predictoor addresses bug if invalid addresses in ppss.

### LINKED ISSUE #1490: [PdrDashboard, v0.4] PPSS my_addresses not found
### Describe the bug
When adding a predictoor address to the `my_addresses` list inside the ppss config that doesn't exist in the database the app doesn't start and an error is shown in the console saying that the id can't be found

### To Reproduce
Steps to reproduce:
1. Add predictoor addresses to `my_addresses` which doesn't exist in the data base
2. Start the dashboard

### Expected behavior
App should start as normal and shouldn't crash if predictoor address can't be found

calina-c:


kdetry (approved):
lgtm

PR #1493, opened 2024-08-02
[Fix #1491] PdrDashboard v0.4 - Split utils.py into concerns

### LINKED ISSUE #1491: [PdrDashboard] v0.4 Split up utils.py
Discussed with @kdetry: I will help split up utils because too many mixed functionalities are caught up in there, leading to my thinking that certain dict utils were already in place, although they were just similar.

Concerns currently tied up in this same file:
- db functions
- dict manipulation stuff, col renaming etc...
- prices and costs

should be separated. While it is true that util.py is a single file for other dashboards and we would break the standard in that sense, Pdr Dashboard is insofar the most complex dashboard and needs to be split.

_Originally posted by @calina-c in https://github.com/oceanprotocol/pdr-backend/issues/1487#issuecomment-2264981347_

calina-c:


KatunaNorbert (approved):
Looks good

PR #1499, opened 2024-08-05
[Issue #1500 - debug] Try running CIs new Barge branch for docker compose fix

### LINKED ISSUE #1500: [CI] All CIs are failing from docker compose command in Barge
Contract deployment times out in all CI runs. First it was intermitent, would work after some reruns. But starting Friday it fails every time.

The problem seems to be from the Barge repository, wrt `docker-compose` command, as seen here:
https://github.com/oceanprotocol/pdr-backend/actions/runs/10241193240/job/28329113626?pr=1498
and encountered by others: https://github.com/orgs/community/discussions/116610

If this is indeed the issue, we need to change the command in start_ocean.py. The Barge repository should probably upgrade to docker v2, since v1 is deprecated from july 2023: https://docs.docker.com/compose/migrate/

I talked to @alexcos20 and we are currently experimenting with a new barge ref, showing promise.

calina-c:
Closes #1500.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -30,6 +30,7 @@ jobs:
         with:
           repository: "oceanprotocol/barge"
           path: "barge"
+          ref: "bug/fix_docker_compose"
 
       - name: Run Barge
         working-directory: ${{ github.workspace }}/barge


calina-c:
Barge was updated in main, so this experimental branch is moot.

PR #1502, opened 2024-08-06
Fix #1501: Implement NumberFormatter

### LINKED ISSUE #1501: [PdrDashboard, v0.4] Implement Numerize and format helpers
- use the numerize library
- operate with raws everywhere up until final show
- for tables, work with column definitions and table data, for metrics and other single displays, work with the components themselves (e.g. get_metric)
- these components should show "human" data and contain the underlying data in the back

kdetry:
Fixes #1501 

For now, it only includes the NumberFormatter class.
I thought it would be better to move forward after we get a consensus about the class

### DIFF
--- a/pdr_backend/util/number_formatter.py
+++ b/pdr_backend/util/number_formatter.py
@@ -1,0 +1,159 @@
+from typing import Union, List, Optional
+import locale
+
+
+class NumberFormatter:
+    """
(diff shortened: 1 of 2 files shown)


kdetry:
We have decided to continue with a 3rd party library and create some helpers to use it

PR #1505, opened 2024-08-06
Fix #1501: Implement Numerize and format helpers

### LINKED ISSUE #1501: [PdrDashboard, v0.4] Implement Numerize and format helpers

kdetry:
Fix #1502

### DIFF
--- a/pdr_backend/pdr_dashboard/test/test_util_format.py
+++ b/pdr_backend/pdr_dashboard/test/test_util_format.py
@@ -69,7 +69,7 @@ def test_format_value():
     assert format_value(1.12, "accuracy_metric") == "1.12%"
     assert format_value(12, "feeds_page_Volume_metric") == "12 OCEAN"
     assert format_value(12, "feeds_page_Revenue_metric") == "12 OCEAN"
-    assert format_value(9876.12, "profit_metric") == "9876.12 OCEAN"
+    assert format_value(9876.12, "profit_metric") == "9.88K OCEAN"
     assert format_value(9876.12, "stake_metric") == "9876.12 OCEAN"
     assert format_value(9876, "costs_metric") == "~9876"
     assert (
@@ -77,7 +77,7 @@ def test_format_value():
         == "0x123...67890"
     )
     assert format_value(12.0, "avg_accuracy") == "12.0%"
-    assert format_value(9876.12, "sales_revenue_(OCEAN)") == "9876.12"
+    assert format_value(9876.12, "sales_revenue_(OCEAN)") == "9.88K"
     assert format_value(9876543, "volume_(OCEAN)") == "9.88M"
     assert format_value(12.0, "avg_stake_per_epoch_(OCEAN)") == "12.0"
     assert format_value(12, "unknown") == "12"
--- a/pdr_backend/pdr_dashboard/test/test_util_format.py
+++ b/pdr_backend/pdr_dashboard/test/test_util_format.py
@@ -1,0 +1,80 @@
+from pdr_backend.pdr_dashboard.util.format import (
+    format_eth_address,
+    format_currency,
+    format_percentage,
+    format_currency_with_decimal,
+    format_currency_with_decimal_and_suffix,
+    format_avg_currency_with_decimal,
+    format_value,
+)
+
+
+def test_format_eth_address():
+    assert (
+        format_eth_address("0x1234567890123456789012345678901234567890")
+        == "0x123...67890"
+    )
+    assert (
+        format_eth_address("0x123456789012345678901234567890123456789")
+        == "0x123...56789"
+    )
+
+
+def test_format_eth_address_empty():
+    assert format_eth_address("") == "..."
+
+
+def test_format_currency():
+    assert format_currency(1234567890.1234567890) == "1.23B OCEAN"
+    assert (
+        format_currency(1234567890.1234567890, show_decimal=True)
+        == "1234567890.12 OCEAN"
+    )
+    assert format_currency(1234567890.1234567890, suffix=" OCEAN") == "1.23B OCEAN"
+    assert (
+        format_currency(1234567890.1234567890, suffix=" OCEAN", show_decimal=True)
+        == "1234567890.12 OCEAN"
+    )
+
+
+def test_format_percentage():
+    assert format_percentage(1) == "1.0%"
+    assert format_percentage(12) == "12.0%"
+    assert format_percentage(12.345) == "12.35%"
+
+
+def test_format_currency_with_decimal():
+    assert format_currency_with_decimal(1234567890.1234567890) == "1234567890.12"
+    assert format_currency_with_decimal(1234567890) == "1234567890"
+
+
+def test_format_currency_with_decimal_and_suffix():
+    assert (
+        format_currency_with_decimal_and_suffix(1234567890.1234567890)
+        == "1234567890.12 OCEAN"
+    )
+    assert format_currency_with_decimal_and_suffix(1234567890) == "1234567890 OCEAN"
+
+
+def test_format_avg_currency_with_decimal():
+    assert format_avg_currency_with_decimal(1234567890.1234567890) == "~1234567890.12"
+    assert format_avg_currency_with_decimal(1234567890) == "~1234567890"
+
+
+def test_format_value():
+    assert format_value(12.0, "feeds_page_Accuracy_metric") == "12.0%"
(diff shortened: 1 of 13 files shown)


pdr_backend/pdr_dashboard/util/format.py
```diff
@@ -0,0 +1,108 @@
...
+        case "sales_revenue_(OCEAN)" | "volume_(OCEAN)":
+            return format_ocean_amount(int(value), w_suffix=False)
+        case "avg_stake_(OCEAN)":
+            return f"~{format_decimal_ocean_amount(float(value), w_suffix=False)}"
+        case _:
+            return str(value)
+
+
+@enforce_types
+def format_addr(addr: str) -> str:
```
calina-c, line 60: I would name this as something more indicative of the shortening. I had this issue when debugging, where I didn't realise which is the address with elipses and which is the full one, based solely on the keys.

pdr_backend/pdr_dashboard/util/format.py
```diff
@@ -0,0 +1,108 @@
...
+        amount (float): Ocean amount.
+    Returns:
+        str: Formatted Ocean amount.
+    """
+
+    return f"{round(amount, 2)}{" OCEAN" if w_suffix else ""}"
+
+
+@enforce_types
+def format_accuracy(accuracy: float) -> str:
```
calina-c, line 99: I would rename:
-> format_accuracy -> format_percentage
-> format_decimal_ocean_amount -> format_currency (and parameterize with "OCEAN", replacing w_suffix with the needed string. Maybe in the future we want to format other currencies)

I also think we could work on DRY betwn format_decimal_ocean amount and format_ocean_amount.

calina-c: TL;DR: I would rather have the specific formatting functions be generic and implement a mapping from the higher-level formatting function.

pdr_backend/pdr_dashboard/util/format.py
```diff
@@ -0,0 +1,108 @@
+from typing import Union
+
+from enforce_typing import enforce_types
+from numerize import numerize
+
+
+@enforce_types
+def format_metric(metric: Union[int, float], value_id: str) -> str:
+    match value_id:
```
calina-c, line 9: I would create a config dict mapping value id to the function name (without the format_ part).
If the metric key is not in the config dict, thow an exception.
Otherwise call format_{function name}.

It helps with clarity and coroborated with my other comments it would look smth like:
{
"costs_metric": "percentage"
...
}

func_name = "format_" + formatting[value_id]
func = <get the function>
return func(metric)

pdr_backend/pdr_dashboard/util/format.py
```diff
@@ -0,0 +1,108 @@
...
+        case "feeds_page_Accuracy_metric" | "accuracy_metric":
```
calina-c, line 10: I would also add another parameter metric_type to replace value Id. i.e. instead of looping through the ids, we loop through an accepted list of format keys, i.e. value, stake, cost.
This means we can reuse these in table cols.

pdr_backend/pdr_dashboard/util/format.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+@enforce_types
+def format_column(value: Union[int, float, str], column_id: str) -> str:
+    match column_id:
+        case "addr":
+            return format_addr(value)
+        case "avg_accuracy":
+            return format_accuracy(float(value))
+        case "sales_revenue_(OCEAN)" | "volume_(OCEAN)":
```
calina-c, line 51: see here, where we could "join" these functions. We have accuracy metrics and accuracy columns. We can apply the same formatting in both cases, provided we don't check for the column_id itself, but a column_type (metric_type respectively)

calina-c (changes requested):
I know it's not finished but I wanted to give some initial feedback.

kdetry:
Tests are added.
the Table column name issue is handled here @KatunaNorbert

pdr_backend/pdr_dashboard/test/test_util_format.py
```diff
@@ -0,0 +1,80 @@
...
+        == "0x123...67890"
+    )
+    assert (
+        format_eth_address("0x123456789012345678901234567890123456789")
+        == "0x123...56789"
+    )
+
+
+def test_format_eth_address_empty():
+    assert format_eth_address("") == "..."
```
KatunaNorbert, line 24: Why `...` is returned when the address is empty? It should return an empty string or return a message such as: 'No address'

pdr_backend/pdr_dashboard/test/test_util_format.py
```diff
@@ -0,0 +1,80 @@
...
+    assert (
+        format_currency_with_decimal_and_suffix(1234567890.1234567890)
+        == "1234567890.12 OCEAN"
+    )
+    assert format_currency_with_decimal_and_suffix(1234567890) == "1234567890 OCEAN"
+
+
+def test_format_avg_currency_with_decimal():
+    assert format_avg_currency_with_decimal(1234567890.1234567890) == "~1234567890.12"
+    assert format_avg_currency_with_decimal(1234567890) == "~1234567890"
```
KatunaNorbert, line 61: this should be rename, the '~' sign doesn't represent an avg but an approximation

pdr_backend/pdr_dashboard/test/test_util_format.py
```diff
@@ -0,0 +1,80 @@
...
+    assert format_value(9876.12, "stake_metric") == "9876.12 OCEAN"
+    assert format_value(9876, "costs_metric") == "~9876"
+    assert (
+        format_value("0x1234567890123456789012345678901234567890", "addr")
+        == "0x123...67890"
+    )
+    assert format_value(12.0, "avg_accuracy") == "12.0%"
+    assert format_value(9876.12, "sales_revenue_(OCEAN)") == "9876.12"
+    assert format_value(9876543, "volume_(OCEAN)") == "9.88M"
+    assert format_value(12, "avg_stake_(OCEAN)") == "~12"
```
KatunaNorbert, line 79: remove the '~' from the average values

KatunaNorbert:
Rename the stake column from 'Avg Stake' -> 'Avg Stake per epoch'
Could you also add the numerized formatting to the values from the Home page?

kdetry:
I guess done, can you review @KatunaNorbert

KatunaNorbert:
The Stake column in Predictoor table and the metrics look like not being formatted:
Screenshot 2024-08-07 at 16 20 06
Screenshot 2024-08-07 at 16 20 37

calina-c (changes requested):

pdr_backend/pdr_dashboard/util/format.py
```diff
@@ -0,0 +1,148 @@
+from typing import Union, Dict, List, Any
+
+from enforce_typing import enforce_types
+from numerize import numerize
+
+FORMAT_CONFIG = {
+    "feeds_page_Accuracy_metric": "percentage",
```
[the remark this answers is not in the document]
kdetry: the key side of the dictionary represents the column id (or metric id), the right side presents the formatter function. I did not understand what would I fix here ?

kdetry:
> The Stake column in Predictoor table and the metrics look like not being formatted: Screenshot 2024-08-07 at 16 20 06 Screenshot 2024-08-07 at 16 20 37

I think we shouldn't format like "8K Ocean" the stake column, because these are already small amount (usually less then 1000) and showing decimals is more important (like 17.12)

I fixed the metric

KatunaNorbert:
> I think we shouldn't format like "8K Ocean" the stake column, because these are already small amount (usually less then 1000) and showing decimals is more important (like 17.12)

In my opinion we should make the format constant everywhere in the app. For decimals I agree that it is important for low amounts, so maybe we could do something like this:
- if amount <1 or 10 show decimals
- else show formatted without decimals

But, why can't work with decimals, If i look at the other columns decimals are shown:
Screenshot 2024-08-08 at 15 21 56

kdetry:
@KatunaNorbert fixed

kdetry:
I have separated the raw data and the display data in the Feeds.py for proper filtering in the callbacks_feed.py.

KatunaNorbert (approved):

PR #1522, opened 2024-08-14
Fix #1519: Dashduo tests and the "sample_etl" structure

### LINKED ISSUE #1519: [PdrDashboard, v0.4] Complete behavior tests for the feed page
- [x] Add selenium / dash_duo tests for the feed page
- [x] Create some combinations of test cases
- [x] add the sample_etl structure and implement for feed tests for now

kdetry:
Fixes #1519


pdr_backend/pdr_dashboard/test/conftest.py
```diff
@@ -168,3 +174,37 @@ def setup_app_with_favourite_addresses(
     setup_app_main(app, ppss)
 
     return app
+
+
+@pytest.fixture
+def setup_app_with_etl_sample_data(tmpdir):
```
idiom-bytes, line 180: please use `_sample_etl` and keep the naming consistent../.

pdr_backend/pdr_dashboard/test/test_callbacks_feeds.py
```diff
@@ -0,0 +1,181 @@
...
+    dash_duo.wait_for_element("#base_token")
+
+    # Test filtering with base token
+    _set_dropdown_and_verify_row_count(dash_duo, "#base_token", "ETH", 3)
+
+    # Test filtering with time
+    _set_dropdown_and_verify_row_count(dash_duo, "#time", "5m", 2)
+
+    _clear_filters(dash_duo)
+    _assert_table_row_count(dash_duo, "#feeds_page_table", 21)
```
idiom-bytes, line 93: These asserts should be validating that you're getting the right columns and required data for all dashboards/work.

Asserting a simple row_count() is likely not appropriate coverage/validation

idiom-bytes (changes requested):
please review comments/requests

pdr_backend/pdr_dashboard/test/test_callbacks_feeds.py
kdetry: I can agree that for the general table test but doing it for every filtering process is unnecessary

kdetry: ok I created a practical way to implement it and I developed this feature for all filter and search tests

KatunaNorbert: For example for filtering we could check that the table columns value respects the filter value: if filter by base token check that all base token column values are equal to filter value

pdr_backend/pdr_dashboard/test/filtered_volume_max_1400.json
```diff
@@ -0,0 +1 @@
+[]
```
calina-c, line 1: I don't think this should be a constant test file. When we test against it, we can use the the empty list directly.

pdr_backend/pdr_dashboard/test/resources.py
```diff
@@ -78,3 +84,54 @@ def start_server_and_wait(dash_duo, app):
     dash_duo.wait_for_element("#predictoors_table")
     dash_duo.wait_for_element("#feeds_table tbody tr")
     dash_duo.wait_for_element("#predictoors_table tbody tr")
+
+
+def _navigate_to_feeds_page(dash_duo):
+    dash_duo.wait_for_element("#feeds")
+    dash_duo.find_element("#feeds").click()
+    time.sleep(1)
```
calina-c, line 92: instead of time.sleeps(), can we use dash duo's wait for element to be visible and pick a relevant element inside the feeds page? It's cleaner, and could save us some time. The 1s sleep is arbitrary, while a "wait for element to be visible" is more relevant here. Btw those functions do have a max timeout so they should also take care of the "max waiting time" part.

pdr_backend/pdr_dashboard/test/resources.py
```diff
@@ -78,3 +84,54 @@ def start_server_and_wait(dash_duo, app):
...
+
+
+def _remove_tags(text):
```
calina-c, line 95: to clarify this functionality, could you add a comment or a docstring giving an example. What tags is it removing and why?

pdr_backend/pdr_dashboard/test/test_callbacks_feeds.py
```diff
@@ -0,0 +1,231 @@
...
+    start_server_and_wait,
+    _clear_filters,
+    _set_dropdown_and_verify_row_count,
+    _set_input_value_and_submit,
+    _remove_tags,
+    _set_searchbar_value,
+)
+
+
+def _get_table_data_as_json(table):
```
calina-c, line 17: This function doesn't return a json, it returns a dictionaty as far as I can tell. Please adjust its name.

pdr_backend/pdr_dashboard/test/test_callbacks_feeds.py
```diff
@@ -0,0 +1,231 @@
...
+        cells = row.find_elements(By.XPATH, ".//td")
+        row_data = {headers[i]: _remove_tags(cells[i].text) for i in range(len(cells))}
+        table_data.append(row_data)
+
+    return table_data
+
+
+def _save_table_data_as_json(table, filename):
+    table_data = _get_table_data_as_json(table)
+    with open("pdr_backend/pdr_dashboard/test/" + filename, "w") as f:
```
calina-c, line 35: I initially thought the new test resources were fixed. However, I know see what you are doing here. I suggest moving these "volatile" test resources in a specific subfolder and adding them to gitignore.

Keep a clear distinction between the reference and in-process resources (json files).

pdr_backend/pdr_dashboard/test/test_callbacks_feeds.py
```diff
@@ -0,0 +1,231 @@
...
+    assert len(rows) == 21
+
+    header = rows[0].find_elements(By.XPATH, ".//th")
+    columns = header
+    assert len(columns) == 12
+
+    # Validate headers
+    header_texts = [_remove_tags(c.text) for c in columns]
+    expected_headers = [
+        "",
```
calina-c, line 68: this header line is a bit of a pink flag to me. Could you help me understand what is happening here?

pdr_backend/pdr_dashboard/test/test_callbacks_feeds.py
```diff
@@ -0,0 +1,231 @@
...
+            By.CLASS_NAME, "dash-graph"
+        )
+    )
+    assert number_of_plots == 6
+
+    # Close modal by clicking the background
+    dash_duo.find_element(".modal").click()
+
+    # Ensure no row is selected
+    assert all(
```
calina-c, line 200: you can assert "not any" instead to avoid negation at a lower level

calina-c (changes requested):

pdr_backend/pdr_dashboard/test/filtered_volume_max_1400.json
KatunaNorbert: Let's keep it like this, in case the test result will be changed in the future, also keeping it consistent with other result formats

pdr_backend/pdr_dashboard/test/resources.py
KatunaNorbert: I agree, if the action is not a generic one such as clicking a button which could have many different effects and if the action does some changes to the other elements.
In the line from above we can replace it, but in many other places we can't

pdr_backend/pdr_dashboard/test/test_callbacks_feeds.py
KatunaNorbert: The first column contains the radio inputs for selecting rows, that's why the header is empty

pdr_backend/pdr_dashboard/test/test_callbacks_feeds.py
KatunaNorbert: Looks like this function is not used, so we can delete it

calina-c:
I am approving to allow merge if urgent. @KatunaNorbert and I are also discussing one last nitpick regarding gitignoring some artefacts, if needed.

calina-c (approved):

Issue #1244, opened 2024-06-19
[Model/trade] Experiment on volume bars, dollar bars, DIBs, etc


trentmc:
## Background / motivation

5-min candles ("time bars") don't have _that_ much info. And 5min (or any time tick) is quite constraining.

We can construct more informative bars from raw trade data. (Where raw trade data = each atomic trade on its own.)
- Volume bars. Eg each bar is 2.0 BTC worth of trades
- Dollar bars. Eg each bar is $100K worth of trades
- Dollar imbalance bars. Eg if the # ticks up or down exceed a threshold, then the imbalance bar value changes

Let's play with it to see how well we can predict or trade against predictions. This can be fully separate from simulation to start with.

## TODOs

- [x] Review Kaiko API
- [x] Review bars, including info bars. See blog posts below
- [x] Get going with raw trade data from kaiko
- [x] Create a branch. Most subsequent work is in that branch (for now)
- [x] Construct bars from raw trade data
  - [x] Volume bars
  - [x] Dollar bars
  - [x] Dollar imbalance bars
  - [ ] And anything else relevant
- [x] Run simulation benchmarks: how does each bar do wrt "trader $ made"
- [ ] More. TBD

#### Resources: Maks Ivanov

1. [Blog](https://towardsdatascience.com/financial-machine-learning-part-0-bars-745897d4e4ba) [PDF](https://github.com/user-attachments/files/15953209/ivanov.pdf) "Financial Machine Learning Part 0: Bars", Feb 27, 2019
   - 🔥🔥**Has Py code for Time Bars, Tick Bars, Volume Bars, Dollar Bars, Dollar Imbalance Bars**. In a nice way that builds from one to the next

#### Resources: Ved Prakash

1. [Blog](https://ved933409.medium.com/major-reasons-why-machine-learning-fails-in-stock-prediction-part-02-7cc201305ce2) [PDF](https://github.com/user-attachments/files/15952446/ved-part2.pdf) "Major reasons why ML fails in stock prediction : Part 2", Feb 17, 2024. It heavily references the book "Adv. in Financial ML" by Marcos López de Prado, and related video (see above).
   - 🔥 **has Py code for Tick Imbalance Bars (TIBs) and Tick Run Bars (TRBs)**
   - [Related video](https://www.youtube.com/watch?v=BRUlSm4gdQ4&t=586s) Marcos Lopez de Prado, "The 7 Reasons Most ML Funds Fail", from QuantCon 2018

[...]

PR #1524, opened 2024-08-16
Issue 1244 add volume bar

### LINKED ISSUE #1233: Put data science job posting onto top of README
Tentative message:

"Do you love AI, Python, and competitive games? The Predictoor team is hiring. [Learn more](https://github.com/oceanprotocol/jobs/blob/main/datascience.md)."

Related: banner for pdr-web https://github.com/oceanprotocol/pdr-web/issues/309

AmandaZYY:
## Description

add volume bar, tick bar and dollar bar

ex: Fixes #1233

## Commit message

issue #1244: add volume bar

ex: Fix #1233: Add pull requests template (PR #1200)


pdr_backend/cli/arg_feed.py
```diff
@@ -124,12 +167,34 @@ def _unpack_feeds_str(feeds_str: str) -> List[ArgFeed]:
 
     exchange_str = feeds_str_split[0]
 
-    timeframe_str = feeds_str_split[-1]
+    threshold_str = feeds_str_split[-1]
     offset_end = None
 
+    if vb_str_ok(threshold_str):
```
calina-c, line 173: is there a possibility of multiple thresholds to be in place? If so, how is that managed?

pdr_backend/cli/arg_feed.py
```diff
@@ -159,12 +225,63 @@ def _unpack_feeds_str(feeds_str: str) -> List[ArgFeed]:
...
     pairs = ArgPairs.from_str(pairs_list_str)
 
-    feeds = [
-        ArgFeed(exchange_str, signal_str, pair_str, timeframe_str)
-        for signal_str in signal_str_list
-        for pair_str in pairs
-        for timeframe_str in timeframe_str_list
-    ]
+    if isinstance(vb_str_list[0], ArgVB):
+        if len(pairs) != len(vb_str_list):
```
calina-c, line 229: the following block is also repeated 3 times with the same content. I will detail more on how this can be helped in further comments.

pdr_backend/cli/arg_tb.py
```diff
@@ -0,0 +1,94 @@
+import re
+from typing import List, Union
+
+from enforce_typing import enforce_types
+
+
+# Tick Bar
+class ArgTB:
```
calina-c, line 8: so what I'm seeing up until now is these classes are very similar but have a different prefix to the string arg itself. There are better ways to manage this without creating 3 virtually identical classes.

The easiest way is to create a class called ArgThreshold which has its string attribute and supports 3 types of prefixes (db, tb and vb). The regex check, the messages, the verification methods etc. can all use self.prefix to determine validity, instead of each class simply hardcoding it.

I don't think there is a need to define 3 classes, ArgThreshold with a `threshold_type` attribute seems enough, but of course you can also use inheritance if you need stronger type checking. In that case, I would also suggest full names for classes i.e. ArgTickThreshold instead of ArgTB etc.

pdr_backend/lake/alt_bar.py
```diff
@@ -0,0 +1,158 @@
...
+    We did investigate the use of trying to solve this in a vectorised manner but found that a For loop worked well.
+
+    :param data: raw ohlcv data from kaiko Contains 6 columns - timestamp, open, high, low, close and volume.
+    :param metric: cum_ticks, cum_dollar_value, cum_volume
+    :param threshold: A cumulative value above this threshold triggers a sample to be taken.
+    :return: The financial data structure with the updated start timestamp since last bar.
+    """
+
+    list_bars = []
+    cache: list = []
```
calina-c, line 25: please rename this variable something like "bars_cache" or another more relevant word. `cache` is often used when dealing with external cache in Python projects, so it is confusing to the reader that this is in fact a custom cache, a list for the purpose of local processing

pdr_backend/lake/alt_bar.py
```diff
@@ -0,0 +1,158 @@
...
+    start_tm = None
+    cum_ticks, cum_dollar_value, cum_volume, cache, high_price, low_price = (
```
calina-c, line 27: please rename high_price and low_price as highest_price and lowest_price to better reflect their purpose

pdr_backend/lake/alt_bar.py
```diff
@@ -0,0 +1,158 @@
...
+                low_price,
+                close,
+                cum_volume,
+                cum_dollar_value,
+                cum_ticks,
+            ]
+        )
+
+        # If threshold reached then take a sample
+        if eval(metric) >= threshold:  # pylint: disable=eval-used
```
calina-c, line 73: at the beginning of this file, please do some checks on the "metric" string. If we are turning off the security feature of using `eval()` we should cover some bases to compensate

calina-c (changes requested):
This is a really nice job! I added some comments and suggestions for an architectural simplification. It shouldn't take long to implement, but if you need any help I'm happy to provide guidance and to work on the code as well.

Not sure if this is your first PR, because I was on holiday recently. If it is, congrats on your first PR 🥳 . And even if not, congrats are still warranted!

pdr_backend/lake/alt_bar.py
AmandaZYY, line 1: but numbers of tick should be an int right?

pdr_backend/cli/arg_feed.py
AmandaZYY: no, for each feed it can only be one of volume bar or tickbar or dollar bar

pdr_backend/cli/arg_feed.py
```diff
@@ -159,12 +184,24 @@ def _unpack_feeds_str(feeds_str: str) -> List[ArgFeed]:
...
     pairs = ArgPairs.from_str(pairs_list_str)
 
-    feeds = [
-        ArgFeed(exchange_str, signal_str, pair_str, timeframe_str)
-        for signal_str in signal_str_list
-        for pair_str in pairs
-        for timeframe_str in timeframe_str_list
-    ]
+    if isinstance(threshold_str_list[0], ArgThreshold):
+        if len(pairs) != len(threshold_str_list):
```
calina-c, line 188: I would suggest moving this check in a different function, only called if the threshold exists. It would help readability.

calina-c: not only the check, but maybe we could even return argfeeds_with_thresholds() or argfeed_without_thresholds() and clean up even more

pdr_backend/cli/arg_threshold.py
```diff
@@ -0,0 +1,93 @@
...
+        return self.prefix + "_" + self.value_str
+
+    def threshold(self) -> float:
+        return self.value
+
+
+class ArgThresholds(List[ArgThreshold]):
+    def __init__(self, thresholds: Union[List[str], List[ArgThreshold]]):
+        if not isinstance(thresholds, list):
+            raise TypeError("hresholds must be a list")
```
calina-c, line 34: there's a small typo in the error message, missing a t

pdr_backend/cli/arg_threshold.py
```diff
@@ -0,0 +1,93 @@
...
+
+        thres = []
+        for thre in thresholds:
```
calina-c, line 37: you said earlier we only have one type of threshold in a list of thresholds. If so, maybe also add a check that all the arg thresholds in a list should have the same type

pdr_backend/cli/arg_threshold.py
```diff
@@ -0,0 +1,93 @@
...
+
+def _unpack_threshold_str(th_str: str) -> tuple:
+    """
+    Unpacks the vb_str into prefix and value_str.
+
+    Example: Given 'vb_2100.5'
+    Return ('vb', '2100.5')
+    """
+    res = th_str.split("_")
+    if len(res) != 2:
```
calina-c, line 90: I think this should go in the init, when you are checking for prefixes. Make sure the strings start with "vb_" and others instead of "vb" and others. Derive the prefix by removing the underscore. Otherwise, there are two points of error throwing, which in fact check very close pieces.

pdr_backend/lake/ohlcv_data_factory.py
```diff
@@ -85,8 +86,44 @@ def get_mergedohlcv_df(self) -> pl.DataFrame:
...
+                "high",
+                "low",
+                "close",
+                "volume",
+                "Cumulative Dollar Value",
+                "Cumulative Ticks",
+            ]
+            df = rawohlcv_dfs[str(feed.exchange)][str(feed.pair)]
+            for threshold_type in ["volume", "tick", "dollar"]:
+                if getattr(feed, f"{threshold_type}_threshold") is not None:
```
calina-c, line 103: this part can still be improved wrt. DRY principle, maybe add functions on the threshold itself.

calina-c (changes requested):
Looks much better! I added some new comments, let me know if you have any questions :D

pdr_backend/cli/arg_threshold.py
AmandaZYY: open this up for discussions, I think it can be a useful use case: "binance BTC/USDT ETH/USDT 5m tb_405,vb_200" which maps BTC/USDT tb_405 and ETH/USDT vb_200" previous consideration was to forcefull seperate different thresholds

pdr_backend/cli/arg_feed.py
```diff
@@ -45,18 +47,29 @@ def __init__(
         else:
             self.timeframe = timeframe
 
+        if threshold is None:
+            self.threshold = None
+        elif isinstance(threshold, str):
+            self.threshold = ArgThreshold(threshold)
+        else:
```
calina-c, line 54: I think the final else is the same as the first if block

pdr_backend/lake/ohlcv_data_factory.py
```diff
@@ -85,8 +86,38 @@ def get_mergedohlcv_df(self) -> pl.DataFrame:
...
+                "open",
+                "high",
+                "low",
+                "close",
+                "volume",
+                "Cumulative Dollar Value",
+                "Cumulative Ticks",
+            ]
+            df = rawohlcv_dfs[str(feed.exchange)][str(feed.pair)]
+            prefix_to_threshold_type = {"vb": "volume", "db": "dollar", "tb": "tick"}
```
calina-c, line 101: I don't think this mapping is needed.
You can have a feed.thresold.get_bars(self, df_pandas) method
- retrieve the threshold type from self
- build volume/tick/dollar bars accordingly

Or if that logic shouldn't reside there, create a get_threshold_bars(df_pandas, prefix) in the lake module. Return whatever kind of bars correspond to the prefix/type given

AmandaZYY: added a _get_threshold_bars within ohlcv_data_factory

pdr_backend/lake/ohlcv_data_factory.py
```diff
@@ -85,8 +86,31 @@ def get_mergedohlcv_df(self) -> pl.DataFrame:
...
+                "Cumulative Dollar Value",
+                "Cumulative Ticks",
+            ]
+            df = rawohlcv_dfs[str(feed.exchange)][str(feed.pair)]
+            if feed.threshold is not None:
+                logger.info(f"Get {feed.threshold.prefix} bars for %s", feed)
+                bars = []
+                df_pandas = df.to_pandas()
+                bars = self._get_threshold_bars(
+                    df_pandas, feed.threshold.prefix, feed.threshold.threshold
```
calina-c, line 106: you can just pass the feed.threshold to the function.
inside the function, get these two attributes

calina-c (approved):

PR #1536, opened 2024-08-20
Fix #1533: The Predictoors Table

### LINKED ISSUE #1533: [PdrDashboard, v0.5] The Predictoors Table
Create a table that contains all the columns from the mockup provided bellow

image

kdetry:
Fixes #1533

### DIFF
--- a/pdr_backend/pdr_dashboard/pages/predictoors.py
+++ b/pdr_backend/pdr_dashboard/pages/predictoors.py
@@ -1,10 +1,9 @@
-from typing import List, Dict, Any, Tuple, Union
+from typing import List, Dict, Any, Tuple
 
 from dash import html, dash_table
 import dash_bootstrap_components as dbc
 from pdr_backend.pdr_dashboard.util.data import (
     get_feed_column_ids,
-    find_with_key_value,
 )
 from pdr_backend.pdr_dashboard.dash_components.plots import (
     FeedModalFigures,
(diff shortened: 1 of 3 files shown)


pdr_backend/pdr_dashboard/pages/predictoors.py
```diff
@@ -100,29 +101,24 @@ def get_search_bar_row(self):
 
     def get_main_container(self):
         # TODO: replace with predictoors data
```
calina-c, line 103: I think you can remove this TODO now

pdr_backend/pdr_dashboard/pages/predictoors.py
```diff
@@ -132,108 +128,46 @@ def get_feeds_table_area(
...
-                feed["contract"], feed_subcription_stats
-            )
+        ## split the pair column into two columns
+        for data_item in temp_data:
+            print("data_item--->", data_item)
 
-            feed_item.update(subscription_result)
+            temp_pred_item = {}
+            temp_pred_item["addr"] = str(data_item["user"])
+            temp_pred_item["gross_income_(OCEAN)"] = data_item["total_payout"]
```
calina-c, line 156: should we remove the (OCEAN) part?

calina-c (changes requested):
just a few minor comments, and please apply black formatting

pdr_backend/pdr_dashboard/pages/predictoors.py
kdetry: According to the task, there should be the `(OCEAN)` phrase on the column header

https://github.com/oceanprotocol/pdr-backend/issues/1533

calina-c: on the column header, yes... but is it really necessary as a data key? With filtering and formatting and all that, I think adding parantheses (which are special characters) is maybe not such a good idea

calina-c: I'm fine either way, just asking for your opinion too

kdetry: I can agree that it is not pretty 👍 we used like that on the feeds table too

PR #1549, opened 2024-08-22
Fix #1545: Fix formatted column values sorting

### LINKED ISSUE #1545: [PdrDashboard] Table columns sorting not working for formatted values
### Describe the bug
When sorting table columns that contains formatted values, the sorting is wrong

### To Reproduce
Steps to reproduce:
1. Go to Feeds or Predictoors page
2. Sort table columns that are formated for ex like: "1M",  "1.5K"

### Expected behavior
Even if the column values are displayed in a formatted way, they should be sortable by default, non formatted values

### Screenshots
Screenshot 2024-08-22 at 10 07 20

KatunaNorbert:
Fix #1545


pdr_backend/pdr_dashboard/callbacks/callbacks_predictoors.py
```diff
@@ -80,6 +82,16 @@ def filter_table(
         if new_table_data:
             columns = get_feed_column_ids(new_table_data[0])
 
+        if sort_by:
+            # Extract sort criteria
+            sort_col = sort_by[0]["column_id"]
+            ascending = sort_by[0]["direction"] == "asc"
+
+            # Sort by raw "Price" even if the "Formatted Price" is displayed
+            new_table_data = sorted(
```
calina-c, line 91: does it take into account natural language? I.e. "2K" > "2M"?

KatunaNorbert: it uses the not formatted values to sort, so it should work fine

calina-c:
can you also add a test for sorting?

calina-c (approved):

PR #1551, opened 2024-08-23
Fix #1550

### LINKED ISSUE #1550: [PdrDashboard, v0.5] The predictoor table content fixes
- gross income should represent the income without any costs so just payout- stake when predictions is correct
- income from stakes should only count for the income when predictions are right, so there shouldn't be negative values. In worst case should be 0
- add new column called 'stake loses’ that contains the stake amount lost on wrong predictions

kdetry:
Fixes #1550


KatunaNorbert:
"Income from stakes" value should not include the predictoor staked amount, only the profit from other predictoors stake.
Ideally we can decompose the "gross income" into "income from stakes" and "income from subscriptions". If we are not able to do it then we should remove the "income from stakes" column

KatunaNorbert:
I reordered the table columns to go from generic to more specific metrics.
I also have an extra request. Could you add a new column to the end of the table called "Net Income(OCEAN)" that represents: Gross Income - (Stake Loss + Tx Costs) ?

calina-c (approved):

PR #1561, opened 2024-08-29
Bump web3.py to 7.2.0

trizin:
Credits to @calina-c for migration code #1558


setup.py
```diff
@@ -18,7 +18,7 @@
     "dateparser==1.2.0",
     "duckdb==1.0.0",
     "enforce_typing",
-    "eth-account==0.11.0",
```
calina-c, line 21: please remove the eth-account dependency altogether, it is implicit from web3.py. I can do that if needed.

calina-c (approved):

PR #1562, opened 2024-08-29
Fix #1560: Reduce and improve db queries

### LINKED ISSUE #1560: [PdrDashboard] Reduce and improve db queries
### Background / motivation

Reduce number of queries to the db so we can more efficiently use date period filtering

### TODOs / DoD

- [ ] identify duplicate query values and try to reduce number of queries to the db

KatunaNorbert:
Fixes #1560


pdr_backend/pdr_dashboard/pages/predictoors.py
```diff
@@ -87,10 +87,8 @@ def get_search_bar_row(self):
         )
 
     def get_main_container(self):
-        predictoor_stats = self.app.db_getter.predictoor_payouts_stats()
-
         predictoor_cols, predictoor_data, raw_predictoor_data = (
-            self.get_data_for_predictoors_table(predictoor_stats)
+            self.get_data_for_predictoors_table(self.app.predictoors_data)
```
calina-c, line 91: you can use the same self.app.predictoors data in self.get_data_for_predictoors_table, since it's the same instance. No need to pass it around as a parameter

pdr_backend/pdr_dashboard/test/conftest.py
```diff
@@ -54,7 +54,7 @@ def sample_table_rows():
 
 def pytest_setup_options():
     options = Options()
-    options.add_argument("--headless")
+    # options.add_argument("--headless")
```
calina-c, line 57: please reinstate this, we do not need to run visual stuff on the CI machines

pdr_backend/pdr_dashboard/util/data.py
```diff
@@ -71,7 +75,8 @@ def get_predictoors_data_from_payouts(
         list: List of processed user payouts stats data.
     """
 
-    for data in user_payout_stats:
+    payout_stats = deepcopy(user_payout_stats)
+    for data in payout_stats:
         formatted_data = format_dict(
```
calina-c, line 80: I would suggest format_dict had an extra argument "only_include_keys" which does pick from dict implicitly, and since pick from dict is really simple, we don't need an explicit function, just the functionality in format_dict.

I also think this goes well, logically speaking, into the responsibility of formatting the dict

pdr_backend/pdr_dashboard/util/data.py
```diff
@@ -126,3 +131,95 @@ def find_with_key_value(
...
+
+    if result:
+        return {
+            "avg_accuracy": float(result["avg_accuracy"]),
+            "avg_stake_per_epoch_(OCEAN)": float(result["avg_stake"]),
+            "volume_(OCEAN)": float(result["volume"]),
+        }
+
+    return {
+        "avg_accuracy": 0,
```
calina-c, line 149: I'm thinking maybe we should have some defaults in the result itself? Just an idea, not part of the main code review

pdr_backend/pdr_dashboard/util/data.py
```diff
@@ -126,3 +131,95 @@ def find_with_key_value(
...
+        "volume_(OCEAN)": 0,
+    }
+
+
+def get_feeds_subscription_stat_with_contract(
+    contract: str, feed_subcription_stats: List[Dict[str, Any]]
+) -> Dict[str, Union[float, int, str]]:
+    result = find_with_key_value(feed_subcription_stats, "contract", contract)
+
+    if result:
```
calina-c, line 160: can you treat the other case first and unindent?

pdr_backend/pdr_dashboard/util/data.py
```diff
@@ -126,3 +131,95 @@ def find_with_key_value(
...
+        sales_str = f"{result['sales']}"
+
+        df_buy_count_str = (
+            f"{result['df_buy_count']}-DF" if result["df_buy_count"] > 0 else ""
+        )
+        ws_buy_count_str = (
+            f"{result['ws_buy_count']}-WS" if result["ws_buy_count"] > 0 else ""
+        )
+
+        if df_buy_count_str or ws_buy_count_str:
```
calina-c, line 170: or alternately build the result itself. It could also improve this part right here modify result["sales"] if needed

pdr_backend/pdr_dashboard/util/data.py
```diff
@@ -126,3 +131,95 @@ def find_with_key_value(
...
+
+    return {
+        "price_(OCEAN)": 0,
+        "sales": 0,
+        "sales_raw": 0,
+        "sales_revenue_(OCEAN)": 0,
+    }
+
+
+def get_feeds_data_for_feeds_table(
```
calina-c, line 191: this is an excellent candidate of why presentation should go into db getter: it's just renaming some columns and presenting them differently. (Not part of the main code review, but discussed for a further iteration)

setup.py
```diff
@@ -43,7 +43,6 @@
     "time_machine==2.15.0",
     "typeguard==4.3.0",
     "xgboost==2.1.1",
-    "dash_bootstrap_components==1.6.0",
```
calina-c, line 46: are we sure we don't need this? Aren't we using it in other dashboards?

calina-c (changes requested):
Left a few comments

setup.py
KatunaNorbert: it's a duplicate, that's why a removed it

PR #1576, opened 2024-09-02
[Pdr dashboard] Move data loading to dbgetter

### LINKED ISSUE #1575: [Pdr Dashboard] Move data loading from app to db getter

calina-c:
Closes #1575

### DIFF
--- a/pdr_backend/pdr_dashboard/util/db.py
+++ b/pdr_backend/pdr_dashboard/util/db.py
@@ -47,13 +47,13 @@ class AppDataManager:
 
         # initial data formatting for tables, columns and raw data
         self.feeds_cols, self.feeds_table_data, self.raw_feeds_data = (
-            self._formatted_for_feeds_table
+            self._formatted_data_for_feeds_table
         )
         (
             self.predictoors_cols,
             self.predictoors_table_data,
             self.raw_predictoors_data,
-        ) = self._formatted_for_predictoors_table
+        ) = self._formatted_data_for_predictoors_table
 
         valid_addresses = [p["user"].lower() for p in self.predictoors_data]
         self.favourite_addresses = [
@@ -375,7 +375,7 @@ class AppDataManager:
         }
 
     @property
-    def _formatted_for_feeds_table(
+    def _formatted_data_for_feeds_table(
         self,
     ) -> Tuple[List[Dict[str, str]], List[Dict[str, Any]], List[Dict[str, Any]]]:
 
@@ -413,7 +413,7 @@ class AppDataManager:
         return columns, formatted_data, new_feed_data
 
     @property
-    def _formatted_for_predictoors_table(
+    def _formatted_data_for_predictoors_table(
         self,
     ) -> Tuple[List[Dict[str, str]], List[Dict[str, Any]], List[Dict[str, Any]]]:
 
--- a/pdr_backend/pdr_dashboard/util/db.py
+++ b/pdr_backend/pdr_dashboard/util/db.py
@@ -41,16 +41,19 @@ class DBGetter:
         # initial data loaded from database
         self.feeds_payout_stats = self._init_feed_payouts_stats()
         self.feeds_subscriptions = self._init_feed_subscription_stats()
+
         self.predictoors_data = self._init_predictoor_payouts_stats()
         self.feeds_data = self._init_feeds_data()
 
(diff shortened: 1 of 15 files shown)


pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -317,3 +368,153 @@ def predictoors_stats(self):
...
+                self.feeds_data, "pair", search_value, selected_feeds
+            )
+            if search_value
+            else filtered_data
+        )
+
+        return selected_feeds + filtered_data
+
+    @enforce_types
+    def format_predictoors_home_page_table_data(self) -> List[Dict[str, Any]]:
```
KatunaNorbert, line 470: I suggest renaming this to "get_formatted_predictoors_home_page_teble_data" because as it is now it's not clear that it returns something, it suggest that it's only formatting and changing something inside dbGetter

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -1,20 +1,61 @@
...
+# pylint: disable=too-many-instance-attributes
 class DBGetter:
-    def __init__(self, lake_dir: str):
-        self.lake_dir = lake_dir
+    def __init__(self, ppss):
+        self.lake_dir = ppss.lake_ss.lake_dir
+        self.network_name = ppss.web3_pp.network
+
+        # fetch token prices
+        self.fee_cost = calculate_tx_gas_fee_cost_in_OCEAN(
```
KatunaNorbert, line 35: I don't think this should be inside dbGetter, it has nothing to do with db data. Let's movi it back to app

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -1,20 +1,61 @@
...
+        self.feeds_payout_stats = self._init_feed_payouts_stats()
+        self.feeds_subscriptions = self._init_feed_subscription_stats()
+        self.predictoors_data = self._init_predictoor_payouts_stats()
+        self.feeds_data = self._init_feeds_data()
+
+        # initial data formatting for tables, columns and raw data
+        self.feed_cols, self.feed_table_data, self.feed_data_raw = (
+            self._format_for_feeds_table()
+        )
+        self.predictoor_cols, self.predictoor_table_data, self.raw_predictoor_data = (
```
KatunaNorbert, line 51: Could you make the variable names more constant? For example: raw_predictoor_data and feed_data_raw make them: raw_predictoor_data and raw_feed_data or the other way around.

pdr_backend/pdr_dashboard/util/db.py
calina-c: Good point! How about a compromise solution: name it "formatted_pre..." and marking it as a property? I think that's the best representation of what it does.

pdr_backend/pdr_dashboard/util/db.py
calina-c: I could, yes. I kept some original names but I can make it even more uniform 👍

pdr_backend/pdr_dashboard/util/db.py
KatunaNorbert: that could also work

pdr_backend/pdr_dashboard/util/db.py
KatunaNorbert: What I mean is, in general if I want to access the fee_costs I wouldn't expect it do be in the db

calina-c: > What I mean is, in general if I want to access the fee_costs I wouldn't expect it do be in the db

I want to rename db_getter either way, to data manager. I think that is more reflective of what it's doing right now. I think it fits then :)

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -369,7 +374,8 @@ def predictoors_stats(self):
             "Gross Income": tot_gross_income,
         }
 
-    def _format_for_feeds_table(
+    @property
+    def _formatted_for_feeds_table(
```
KatunaNorbert, line 378: `_formatted_for_feeds_table ` -> `_formatted_data_for_feeds_table `

kdetry (approved):
lgtm

KatunaNorbert (approved):

Issue #1585, opened 2024-09-06
[Pdr Dashboard] Proposal: use dataframes instead of dictionaries as sources for Pdr Dashboard DataTables


calina-c:
There are various performance and readability issues with large dictionary manipulation in Pdr Dashboard.

- formatting tables is clunky and difficult to read, and incurrs large costs with iteration -> with dataframes, the application of formatting functions is optimised
- when retrieving data against multiple source queries, e.g. in feeds, we reach complexities of C1 * C2 * C3, where we select each row in the feeds query (as a dictionary result), then lookup that specific row in the subscriptions dictionary to provide extra data, then lookup that specific row in the payouts dictionary as well. -> with dataframes, these are simple joins

Originally, query results are provided in dataframe format, and Dash DataTable supports dataframes as sources. But we complicate matters by converting the results to dictionaries, then formatting and iterating through those dictionaries in a very difficult-to-follow manner, then using the dictionaries as sources for DataTable.

The DataTable is capable to automatically retrieve column headers from the dataframe as well, reducing the amount of parameters we keep passing around through the app.

The change I'm proposing is radical, but suprisingly easy to implement. I took the liberty of making a very limited PoC where:
- visually: I already successfully retrieve the feeds table
- Show how joins can effectively supplant dictionary iteration
- Adapt the format_table function for dataframes to show simplicity of manipulation

I did not go into callback management, nor did I refactor the entire application, since it was only a PoC, but I think it is enough to show this change is warranted. If we want to evolve to larger and larger datasets, in my opinion we should move away from the dictionaries.

KatunaNorbert:
Sounds good. I would be interested to see the improvement results of the POC over a large set of data, maybe after integrating the historical lake data.

calina-c:
Closed by #1593

PR #1593, opened 2024-09-10
[Pdr dashboard] Use dataframes as Pdr dashboard datasources

calina-c:
Work on #1585


KatunaNorbert:
I'm getting an error when trying to select a different time period( 1W, 1M )

Also when selecting perdictoors in the home page, it shows the available data period for predictoors, but then after I choose a feed that switches to "there is no data avaialable"

calina-c:
> I'm getting an error when trying to select a different time period( 1W, 1M )
> 
> Also when selecting perdictoors in the home page, it shows the available data period for predictoors, but then after I choose a feed that switches to "there is no data avaialable"

I'll test it manually. Are there any dash tests for that part btw?

KatunaNorbert:
> I'll test it manually. Are there any dash tests for that part btw?

No, there are not, could pe complex to test it but we might find a simple way for it

KatunaNorbert (approved):

PR #1596, opened 2024-09-11
Fix #1595: Resolve DuckDB concurrency issue

### LINKED ISSUE #1595: [Lake, PdrDashboard] Resolve DuckDB data concurrency
### Background / motivation

We want to have Lake update and Dashboard constantly running agains the same data at the same time without any issue.

The problem is that DuckDB doesn't support multiple read connections when there is a write connections.

The propose solution is to have Lake constantly export the new data base data to parquet files and the dashboard to run queries and access data from the parquet files instead of the db file, this way we can avoid the multiple connections to the same db file.

### TODOs / DoD

- [ ] Update lake to export db data to parquet files and keep them updated
- [ ] Update dashboard to read data and execute queries agains parquet files

KatunaNorbert:
Fixes #1595


kdetry:
I have some refactorings on this branch, review after my changes

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -349,3 +351,101 @@ def get_query_move_table_data(self, from_table, to_table):
         return (
             f"INSERT INTO {to_table.table_name} SELECT * FROM {from_table.table_name};"
         )
+
+    @enforce_types
+    def export_tables_to_parquet_files(
+        self,
+        seconds_between_exports: int,
+        number_of_files_after_which_combine_into_one: int,
```
idiom-bytes, line 359: i don't understand how this is used... parquet is designed to have hundreds of files existing in the same folder, and you should be able to just query the folder

please revise this answer

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -57,6 +58,11 @@ def __init__(self, ppss):
             addr for addr in ppss.predictoor_ss.my_addresses if addr in valid_addresses
         ]
 
+    def format_to_parquet_file_path(self, table_class: Type[LakeMapper]) -> str:
```
idiom-bytes, line 61: please improve the DX...

1. rename this function into something simpler
self.tbl_parquet_path(Table)

2. perhaps move it to the TBL object...
```f"""SELECT * FROM {BronzePrediction.parquet_path}"""

File: pdr_backend/pdr_dashboard/test/test_util_db.py [hunk 3]
```diff
@@ -1,25 +1,25 @@
...
     db_mgr._query_db(query)
 
-    lake_dir = db_mgr.lake_dir
-    mock_duckdb_data_store.assert_called_once_with(lake_dir, read_only=True)
-    mock_duckdb_data_store.return_value.query_data.assert_called_once_with(query)
+    # Assert that duckdb.execute was called once with the correct query
+    mock_duckdb_execute.assert_called_once_with(query)
+
+    # Mocking the return value from duckdb.execute(query)
+    mock_duckdb_execute.return_value.pl.assert_called_once()
```
idiom-bytes, line 22: 1. I'm not sure what this test is supposed to do, it doesn't seem to cover/assert/prove anything.
2. There is no test/coverage for parquet functionality.

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -349,3 +351,101 @@ def get_query_move_table_data(self, from_table, to_table):
...
+        parquet_file = f"{table}_{max_timestamp_from_db}.parquet"
+        parquet_file_path = os.path.join(table_folder_path, parquet_file)
+
+        query = f"""
+        COPY (SELECT * FROM {table} WHERE timestamp > {max_timestamp_from_parquet})
+        TO '{parquet_file_path}' (FORMAT 'parquet');
+        """
+        self.execute_sql(query)
+
+    def _should_combine_files(self, table_folder_path: str, file_limit: int) -> bool:
```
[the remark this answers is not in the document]
KatunaNorbert: After our discussion I'm going to remove the combining part and re-export the files

pdr_backend/lake/duckdb_data_store.py
KatunaNorbert: will remove the combine part but will keep the param under another name. We can use it to nuke everything and re-export the db as 1 file periodically. Once a lot of files are created it could impact the performance.

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -25,14 +25,15 @@
 )
 from pdr_backend.util.constants_opf_addrs import get_opf_addresses
 from pdr_backend.util.time_types import UnixTimeMs
+from pdr_backend.lake.lake_mapper import LakeMapper
 
 logger = logging.getLogger("predictoor_dashboard_utils")
```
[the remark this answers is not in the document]
KatunaNorbert: There is no code for handling parquet exports, just for reading from them

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -349,3 +351,101 @@ def get_query_move_table_data(self, from_table, to_table):
...
+        )
+        current_timestamp = UnixTimeS(int(time.time())).to_milliseconds()
+
+        return (
+            current_timestamp - max_timestamp_from_parquet
+            >= seconds_between_exports * 1000
+        )
+
+    def _get_max_timestamp_from_parquet_files(self, table_folder_path: str) -> int:
+        try:
```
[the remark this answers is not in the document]
KatunaNorbert: So, what this means is that for bronze tables we have to re-export the entire table each time because as a saved row changes we can't just update the table. Am I right?

pdr_backend/pdr_dashboard/util/db.py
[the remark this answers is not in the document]
KatunaNorbert: If we need to create another DuckDB object inside the dashboard then a new connection is created and we run into concurrency issue, because the object automatically creates a connection inside constructor

pdr_backend/pdr_dashboard/test/test_util_db.py
KatunaNorbert: There were no test written yet until we decide what the final solution is

kdetry (approved):
I think it is okay now

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -349,3 +352,105 @@ def get_query_move_table_data(self, from_table, to_table):
...
+        number_of_files_after_which_re_export_db: int,
+    ):
+        export_folder_path = get_export_folder_path(self.base_path)
+        tables = self.query_data("SHOW TABLES")["name"]
+
+        # Ensure export folder exists
+        os.makedirs(export_folder_path, exist_ok=True)
+
+        for table in tables:
+            table_folder_path = os.path.join(export_folder_path, table)
```
calina-c, line 369: I think that can be a function of the table? Like we define its csv representation folder only on-demand, we can also define its export path on demand

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -349,3 +352,105 @@ def get_query_move_table_data(self, from_table, to_table):
...
+        current_timestamp = UnixTimeS(int(time.time())).to_milliseconds()
+
+        return (
+            current_timestamp - max_timestamp_from_parquet
+            >= seconds_between_exports * 1000
+        )
+
+    def _get_max_timestamp_from_parquet_files(self, table_folder_path: str) -> int:
+        try:
+            result = duckdb.execute(
```
calina-c, line 397: you can use scalar=True here, it's already implemented

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -349,3 +352,105 @@ def get_query_move_table_data(self, from_table, to_table):
...
+            self.query_scalar(f"SELECT MAX(timestamp) FROM {table}") or 0
+        )
+        max_timestamp_from_parquet = self._get_max_timestamp_from_parquet_files(
+            table_folder_path
+        )
+
+        if max_timestamp_from_db <= max_timestamp_from_parquet:
+            return  # No new data to export
+
+        parquet_file = f"{table}_{max_timestamp_from_db}.parquet"
```
calina-c, line 416: I would suggest getting this from the table itself, don't format it here. Or even transfer using the table itself. The table should know its own name and its parquet root and export root (and csv root etc.). But I think that clashes with Roberto's vision.

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -349,3 +352,105 @@ def get_query_move_table_data(self, from_table, to_table):
...
+    def _nuke_table_folders_and_re_export_db(self, table_folder_path: str):
+        # Delete parquet files from table directory
+        delete_files(table_folder_path)
+
+
+def get_export_folder_path(base_path):
+    return f"{base_path}/exports"
+
+
+def tbl_parquet_path(base_path: str, table_class: Type[LakeMapper]) -> str:
```
calina-c, line 446: see here as well, it belongs in the table class itself :)

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -75,16 +77,13 @@ def _query_db(self, query: str, scalar=False) -> Union[List[dict], Exception]:
...
+            # If scalar, fetch a single result
             if scalar:
-                result = db.query_scalar(query)
-            else:
-                df = db.query_data(query)
-                result = df.to_pandas()
-
-            db.duckdb_conn.close()
-            return result
+                result = duckdb.execute(query).fetchone()
```
calina-c, line 82: wait, why did we remove the query_scalar and moved it here?

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -75,16 +77,13 @@ def _query_db(self, query: str, scalar=False) -> Union[List[dict], Exception]:
...
+                return result[0] if result and len(result) == 1 else result
+            # For non-scalar, fetch as a dataframe and convert to dictionary
+            df = duckdb.execute(query).pl()
```
calina-c, line 85: no, no dictionary, why dictionary?

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -75,16 +77,13 @@ def _query_db(self, query: str, scalar=False) -> Union[List[dict], Exception]:
...
-                result = db.query_scalar(query)
-            else:
-                df = db.query_data(query)
-                result = df.to_pandas()
-
-            db.duckdb_conn.close()
-            return result
+                result = duckdb.execute(query).fetchone()
+                return result[0] if result and len(result) == 1 else result
+            # For non-scalar, fetch as a dataframe and convert to dictionary
```
calina-c, line 84: I panicked here. Please remove this comment. This is a dataframe and not a dictionary

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -94,7 +93,7 @@ def _init_feeds_data(self):
         df = self._query_db(
             f"""
                 SELECT contract, pair, timeframe, source
-                FROM {BronzePrediction.get_lake_table_name()}
+                FROM {tbl_parquet_path(self.lake_dir, BronzePrediction)}
```
calina-c, line 96: yes, definitely move it to the classes and use BronzePrediction.get_tbl_parquet_name or something like that.

KatunaNorbert: moving it there doesn't really help because the Table doesn't know about PPSS and lake dir or exports dir, so I would need to send that as param, and only thing the table will do is since it has the table name just add the '.parquet' at the end.

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -349,3 +352,105 @@ def get_query_move_table_data(self, from_table, to_table):
...
+
+            self._export_table_to_parquet(table, table_folder_path)
+
+    def _should_export(
+        self, table_folder_path: str, seconds_between_exports: int
+    ) -> bool:
+        max_timestamp_from_parquet = self._get_max_timestamp_from_parquet_files(
+            table_folder_path
+        )
+        current_timestamp = UnixTimeS(int(time.time())).to_milliseconds()
```
calina-c, line 388: why time.time and not UnixTimeMs.from_dt(datetime.now())?

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -349,3 +352,105 @@ def get_query_move_table_data(self, from_table, to_table):
         return (
             f"INSERT INTO {to_table.table_name} SELECT * FROM {from_table.table_name};"
         )
+
+    @enforce_types
+    def export_tables_to_parquet_files(
+        self,
+        seconds_between_exports: int,
```
calina-c, line 359: I think this should be UnixTimeS to make it clearer. And later, instead of multiplying it by 1000, call to_ms

pdr_backend/lake/duckdb_data_store.py
```diff
@@ -349,3 +352,105 @@ def get_query_move_table_data(self, from_table, to_table):
...
+        TO '{parquet_file_path}' (FORMAT 'parquet');
+        """
+        self.execute_sql(query)
+
+    def _should_nuke_table_folders_and_re_export_db(
+        self, table_folder_path: str, file_limit: int, table_name: str
+    ) -> bool:
+        if "bronze" in table_name:
+            return True
+        files = [
```
calina-c, line 430: Is there a way to count the files from os, without listing the contents? I think there should be one. Otherwise we are creating a list which we are never using, just to count its number of elements.

pdr_backend/lake/etl.py
```diff
@@ -341,3 +344,14 @@ def do_bronze_queries(self):
                 "do_bronze_queries - completed query %s",
                 etl_query,
             )
+
+    @enforce_types
+    def _export_table_data_to_parquet_files(self):
+        db = DuckDBDataStore(self.ppss.lake_ss.lake_dir)
+
+        if self.ppss.lake_ss.export_db_data_to_parquet_files:
```
calina-c, line 352: if not... return
unindent :)

pdr_backend/pdr_dashboard/test/test_util_db.py
```diff
@@ -98,7 +99,7 @@ def test_get_feed_daily_subscriptions_by_feed_id(_sample_app):
 
     result = db_mgr.feed_daily_subscriptions_by_feed_id(feed_id)
     all_subscriptions = db_mgr._query_db(
-        f"SELECT * FROM {Subscription.get_lake_table_name()}"
+        f"SELECT * FROM {tbl_parquet_path(db_mgr.lake_dir, Subscription)}"
```
calina-c, line 102: I think it's warranted to have a function in the the mapped tables that takes the lake dir parameter and returns the parquet path. Even with the parameter, we avoid mixture of concerns by importing stuff from the duckdb data store here.

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -66,7 +68,7 @@ def set_start_date_from_period(self, period: int):
         self.start_date = start_dt
 
     @enforce_types
-    def _query_db(self, query: str, scalar=False) -> Union[List[dict], Exception]:
+    def _query_db(self, query: str, scalar=False) -> Union[List[dict], Any]:
```
calina-c, line 71: we should fix this typing, it's pandas.dataframe or a scalar (int, float, I think?)

calina-c (changes requested):
I requested a few changes :)

pdr_backend/lake/duckdb_data_store.py
KatunaNorbert: We could use os.walk() or os.scandir() to go trough the directory and if content is file then update a variable which tracks the number of files

pdr_backend/pdr_dashboard/test/test_util_db.py
KatunaNorbert: Well, if we send the lake dir as parameter, we still have to import from duckdb data store here to get the dir

pdr_backend/lake/duckdb_data_store.py
calina-c: https://stackoverflow.com/a/63307131
here is a possible solution that doesn't walk/list the dir

pdr_backend/pdr_dashboard/test/test_util_db.py
calina-c: not really. That will be on the (already imported) table :)

pdr_backend/pdr_dashboard/util/db.py
KatunaNorbert: because query_scalar works on an existing db connection within duckdb data store, we don't want to have a new db connection

pdr_backend/lake/duckdb_data_store.py
KatunaNorbert: It's implemented if we use duckdb data store object to query, if we query files directly using duckdb, it's not implemented

pdr_backend/pdr_dashboard/test/test_util_db.py
KatunaNorbert: how? Table doesn't know about duckdb or ppss, whenever it's used inside it's method those are sent as params

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -74,16 +78,12 @@ def _query_db(self, query: str, scalar=False) -> Union[List[dict], Exception]:
...
-                result = db.query_scalar(query)
-            else:
-                df = db.query_data(query)
-                result = df.to_pandas()
-
-            db.duckdb_conn.close()
-            return result
+                result = duckdb.execute(query).fetchone()
+                return result[0] if result and len(result) == 1 else result
+            df = duckdb.execute(query).pl()
```
calina-c, line 85: what is the purpose of pl?

calina-c (approved):

pdr_backend/pdr_dashboard/util/db.py
KatunaNorbert: to convert the result to Polars DataFrame

calina-c: why? Dash Table only accepts pandas. And the next line is literally a conversion to pandas which worked without the need for pl(). So instead of converting it to pl and then pandas, why not have it directly as pandas?

calina-c:
LGTM

PR #1598, opened 2024-09-13
Remove old payouts query and fix tests.

calina-c:


pdr_backend/pdr_dashboard/test/test_plots.py
```diff
@@ -18,7 +18,9 @@
 def test_process_payouts(_sample_app):
 
     ## convert List[Payout] to List[dict]
-    payouts = _sample_app.data.payouts(None, None, 0).to_dict(orient="records")
+    payouts = _sample_app.data.payouts_from_bronze_predictions(None, None).to_dict(
```
KatunaNorbert, line 21: We are no longer fetching payouts and working with that, but predictions, which includes payouts, but in most places is better to replace payouts with predictions( ex: this function name, result, test name etc). Only use payouts when the actual payout atribute is used from within the predictions

KatunaNorbert (approved):

pdr_backend/pdr_dashboard/test/test_plots.py
KatunaNorbert: We could also handle the renaming inside another PR

PR #1614, opened 2024-09-17
Fix #1607: Use caching for static dashboard queries

### LINKED ISSUE #1607: [PdrDashboard] Cache static query results
### Background / motivation

To improve app performance and db load, we can cache query results that are periodically updated when new parquet data is exported from the Lake. Results should be stored inside parquet files and re-created periodically when new data is exported

### TODOs / DoD

- [ ] Save appropriate query result as parquet files and read from them to show data
- [ ] When new data is exported from the lake, update query result parquet files

KatunaNorbert:
Fixes #1607

### DIFF
--- a/pdr_backend/pdr_dashboard/util/db.py
+++ b/pdr_backend/pdr_dashboard/util/db.py
@@ -38,6 +38,7 @@ class AppDataManager:
         self.network_name = ppss.web3_pp.network
         self.start_date: Optional[datetime] = None
         self.lake_dir = ppss.lake_ss.lake_dir
+        self.second_between_caches = ppss.lake_ss.seconds_between_parquet_exports
 
         self.min_timestamp, self.max_timestamp = (
             self.get_first_and_last_slot_timestamp()
@@ -100,7 +101,7 @@ class AppDataManager:
             # Check if cache file exists and is less than 1 hour old
             if os.path.exists(cache_file_path):
                 file_age = time.time() - os.path.getmtime(cache_file_path)
-                if file_age < 3600:  # 3600 seconds = 1 hour
+                if file_age < self.second_between_caches:
                     query = f"SELECT * FROM '{cache_file_path}'"
             else:
                 # If cache file doesn't exist, run the query and cache the result
--- a/pdr_backend/pdr_dashboard/util/db.py
+++ b/pdr_backend/pdr_dashboard/util/db.py
@@ -183,9 +183,10 @@ class AppDataManager:
                 contract
             ORDER BY volume DESC
         """
+        period_days = (datetime.now() - self.start_date).days if self.start_date else 0
         df = self._query_db(
             query,
-            cache_file_name=f"feed_payouts_stats_{self.start_date.day if self.start_date else 0}",
+            cache_file_name=f"feed_payouts_stats_{period_days}_days",
         )
 
         df["avg_accuracy"] = df["avg_accuracy"].astype(float)
@@ -233,10 +234,12 @@ class AppDataManager:
             ORDER BY
                 apr DESC
         """
-        start_date_day = self.start_date.day if self.start_date else 0
+        start_date_day = (
+            (datetime.now() - self.start_date).days if self.start_date else 0
+        )
         df = self._query_db(
             query,
-            cache_file_name=f"predictoor_payouts_stats_{start_date_day}",
+            cache_file_name=f"predictoor_payouts_stats_{start_date_day}_days",
         )
 
         df["avg_accuracy"] = df["avg_accuracy"].astype(float)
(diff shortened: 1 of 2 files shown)


pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -424,11 +496,11 @@ def feeds_metrics(self) -> dict[str, Any]:
             SELECT COUNT(DISTINCT(contract, pair, timeframe, source))
             FROM {tbl_parquet_path(self.lake_dir, Prediction)}
         """
+        start_date_day = self.start_date.day if self.start_date else 0
```
calina-c, line 499: why not have this in the qurey db function itself, and pass in the cache_file_prefix instead of the full name?

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -66,9 +68,63 @@ def set_start_date_from_period(self, period: int):
...
+        cache_file_path = os.path.join(cache_file_dir, f"{cache_file_name}.parquet")
+
+        try:
+            # Ensure the cache directory exists
+            os.makedirs(cache_file_dir, exist_ok=True)
+
+            # Check if cache file exists and is less than 1 hour old
+            if os.path.exists(cache_file_path):
+                file_age = time.time() - os.path.getmtime(cache_file_path)
+                if file_age < 3600:  # 3600 seconds = 1 hour
```
calina-c, line 99: I would have this configurable :)

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -66,9 +68,63 @@ def set_start_date_from_period(self, period: int):
...
+                    return None
+                return resp[0] if len(resp) == 1 else resp
+
+            # For non-scalar queries, fetch the result as a DataFrame and return as pandas.DataFrame
+            pl_resp = duckdb.execute(query).pl()
+            return pl_resp.to_pandas()
+
+        except FileNotFoundError:
+            print(f"Error: The directory '{cache_file_dir}' does not exist.")
+        except Exception as e:
```
calina-c, line 120: I think the except block should be broken down. It is too vague now where the exception is raised

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -66,9 +68,63 @@ def set_start_date_from_period(self, period: int):
...
+                if resp is None:
+                    return None
+                return resp[0] if len(resp) == 1 else resp
+
+            # For non-scalar queries, fetch the result as a DataFrame and return as pandas.DataFrame
+            pl_resp = duckdb.execute(query).pl()
+            return pl_resp.to_pandas()
+
+        except FileNotFoundError:
+            print(f"Error: The directory '{cache_file_dir}' does not exist.")
```
calina-c, line 119: no printing, please

calina-c (changes requested):

pdr_backend/pdr_dashboard/util/db.py
KatunaNorbert: Maybe we can create at some point a PPSS config for Dashboard, right now I don't know where to place this

pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -63,12 +65,70 @@ def start_date_ms(self) -> int:
...
-        start_dt = datetime.now() - timedelta(days=period) if int(period) > 0 else None
+        start_dt = (
+            datetime.now(tz=timezone.utc) - timedelta(days=period)
+            if int(period) > 0
+            else None
+        )
         self.start_date = start_dt
 
+    @enforce_types
+    def _check_cache_query_data(
```
calina-c, line 76: btw I think a good usage of the cache would also be with formatted dataframes. We do formatting a lot, and it implies copying and formatting a dataframe under certain rules. Maybe we could also define the values for the formatted version of the result and add them to the cache as well

pdr_backend/pdr_dashboard/util/db.py
calina-c: can you please make an issue? You can resolve this comment after pasting the link as a response :D

KatunaNorbert: I'm still thinking if creating a PPSS section for the dashboard it's needed. We only have this field for now.

KatunaNorbert: I have a better idea here. Since the new cache value is related to the new data exported to parquet files, instead of having a separate config param for caching, just use the already existent `seconds_between_parquet_exports`.
The cache update should be related to the new exported data anyway.

calina-c (approved):

PR #1617, opened 2024-09-20
Fix #1616: Fix caching doesn't renew saved values

### LINKED ISSUE #1616: [PdrDashboard] Caching renewal check is not properly working
### Describe the bug

When the data has to be renewed, the new data is fetched but not save to the cache files

KatunaNorbert:
Fixes #1616


pdr_backend/pdr_dashboard/util/db.py
```diff
@@ -72,6 +72,24 @@ def set_start_date_from_period(self, period: int):
         )
         self.start_date = start_dt
 
+    def is_cache_valid(self, cache_file_path: str, seconds_between_caches: int) -> bool:
+        """Checks if the cache file exists and is within the valid time limit."""
+        if os.path.exists(cache_file_path):
```
calina-c, line 77: try doing the negative first to avoid indentation

calina-c (approved):

pdr_backend/pdr_dashboard/util/duckdb_file_reader.py
```diff
@@ -0,0 +1,106 @@
...
+        if self._is_cache_valid(cache_file_path, self.seconds_between_caches):
+            query = f"SELECT * FROM '{cache_file_path}'"
+        else:
+            self._run_query_and_cache(cache_file_path, query)
+
+        # Fetch and return the query result
+        return self._fetch_query_result(query, scalar)
+
+    @enforce_types
+    def _query_db(
```
kdetry, line 74: I have a concern about naming. I think that using an underscore at the beginning of a method name indicates that it is a private method.
As I see we used underscores on every method name in this practice. It could be better fix it

Issue #1594, opened 2024-09-10
[Pdr Dashboard] Follow-up tasks after updating from dictionaries to dataframes


calina-c:
- [x] Investigate whether get_sales_str can be compounded with format_sales_info by looking into "neighboring" columns: done in https://github.com/oceanprotocol/pdr-backend/pull/1599
- [x] payouts is now converted to dictionary everywhere, due to legacy processing. The performance would be improved with native dataframe joining/filtering instead of using the feed x predictoor product. https://github.com/oceanprotocol/pdr-backend/pull/1615
- [ ] optimise some of the df copy parts including initializing of formatted dfs
- [ ] add typing and cleanup if any unused code parts still exist: https://github.com/oceanprotocol/pdr-backend/pull/1633
- [x] look into using polars instead of pandas dataframes and convert them JIT for the DashTable. i.e. perform formatting and manipulation on polars frames, but feed the converted pandas frames to the tables. https://github.com/oceanprotocol/pdr-backend/pull/1619 -> confirmed that DashTable can receive the converted dicts, no need for pandas conversion as "middleware".
- [x] #1628

PR #1619, opened 2024-09-20
[PdrDashboard] Use polars syntax for dataframes in pdr dashboard

calina-c:
works on #1594.


pdr_backend/pdr_dashboard/dash_components/plots.py
```diff
@@ -345,38 +345,63 @@ def process_payouts(
...
+        pl.when(pl.col("payout") > 0).then(1).otherwise(0).alias("correct_prediction"),
+        pl.when(pl.col("payout") > 0)
+        .then(1)
+        .otherwise(0)
+        .cum_sum()
+        .alias("correct_predictions_crt"),
+        pl.col("slot")
+        .map_elements(lambda x: UnixTimeS(int(x)).to_milliseconds(), pl.Int64)
+        .alias("slot_in_unixts"),
+    )
```
kdetry, line 368: If I understood it correct, we can do it like that 

```
p = p.with_columns(
    [
        pl.lit(1).alias("predictions_crt"),
        (pl.col("payout").clip_min(0) - pl.col("stake")).alias("profit_change"),
        pl.when(pl.col("payout") > 0)
        .then(1)
        .otherwise(0)
        .alias("correct_prediction"),
        pl.col("slot")
        .cast(pl.Int64)
        .map_elements(lambda x: UnixTimeS(x).to_milliseconds(), pl.Int64).alias("slot_in_unixts")
    ]
)

p = p.with_columns(
    [
        pl.col("predictions_crt").cum_sum().alias("predictions_crt"),
        pl.col("correct_prediction").cum_sum().alias("correct_predictions_crt")
    ]
)
```

pdr_backend/pdr_dashboard/dash_components/plots.py
```diff
@@ -345,38 +345,63 @@ def process_payouts(
...
+                return_dtype=pl.List(pl.Float64),
+            )
+            .alias("acc_bounds")
+        )
+
+        df = df.with_columns(pl.col("acc_bounds").list.get(0).alias("acc_l"))
+        df = df.with_columns(pl.col("acc_bounds").list.get(1).alias("acc_u"))
+
+        processed.acc_l = df["acc_l"]
+        processed.acc_u = df["acc_u"]
```
kdetry, line 389: We can avoid from using with_columns unnecessary

```
df = p.with_columns(
    [
        pl.struct(["correct_predictions_crt", "predictions_crt"])
        .map_elements(
            lambda x: proportion_confint(
                count=x["correct_predictions_crt"], nobs=x["predictions_crt"]
            ),
            return_dtype=pl.List(pl.Float64),
        )
        .alias("acc_bounds"),
        pl.col("acc_bounds").list.get(0).alias("acc_l"),
        pl.col("acc_bounds").list.get(1).alias("acc_u")
    ]
)

# Assigning the results to `processed` in one step
processed.acc_l = df["acc_l"]
processed.acc_u = df["acc_u"]
```

pdr_backend/pdr_dashboard/dash_components/plots.py
```diff
@@ -345,38 +345,63 @@ def process_payouts(
...
+
+    p = p.with_columns(
+        pl.struct(["correct_predictions_crt", "predictions_crt"])
+        .map_elements(
+            lambda x: x["correct_predictions_crt"] / x["predictions_crt"] * 100,
+            return_dtype=pl.Float64,
+        )
+        .alias("accuracies")
     )
```
kdetry, line 398: what is the difference with the foolowing one? I couldn't get sorry

```
p = p.with_columns(
    ((pl.col("correct_predictions_crt") / pl.col("predictions_crt")) * 100)
    .alias("accuracies")
)
```

pdr_backend/pdr_dashboard/dash_components/plots.py
```diff
@@ -524,43 +549,59 @@ def get_figures_and_metrics(
...
+        .map_elements(
+            lambda x: datetime.datetime.combine(x, datetime.time()), pl.Datetime
+        )
+        .alias("day_dt"),
+    )
+    result.subscription_dates = subscriptions["day_dt"]
+
+    payouts = payouts.with_columns(
+        pl.struct(["payout", "stake"])
+        .map_elements(lambda x: max(x["payout"], 0) - x["stake"], pl.Float64)
```
kdetry, line 578: why do we map here, what is the diff with the following code?


```
payouts = payouts.with_columns(
    [
        (pl.col("payout").clip_min(0) - pl.col("stake")).alias("profit"),
        pl.lit(1).alias("count"),
        pl.lit(1).cum_sum().alias("cnt_cumsum"),
        pl.when(pl.col("payout") > 0)
        .then(1)
        .otherwise(0)
        .cum_sum()
        .alias("cnt_corrpred")
    ]
)
```

pdr_backend/pdr_dashboard/dash_components/plots.py
```diff
@@ -570,37 +611,48 @@ def get_feed_figures(
...
 
-    if payouts.empty:
+    if payouts.is_empty():
         return result
 
-    payouts["profit"] = payouts.apply(
-        lambda x: max(x["payout"], 0) - x["stake"], axis=1
+    payouts = payouts.with_columns(
+        pl.struct(["payout", "stake"])
+        .map_elements(lambda x: max(x["payout"], 0) - x["stake"], pl.Float64)
```
kdetry, line 627: the same map question here

kdetry (changes requested):
I didn't repeatedly comment, but there may be unnecessary mappings in polar phrases. I would appreciate it if you could review them.

pdr_backend/pdr_dashboard/dash_components/plots.py
calina-c: doesn't work because "acc_bounds" is only defined as part of the first "with_columns". I can't split the list in the same one. I can only do it in a second "with_columns". Or, if you have any other ideas I'm happy to try something else

pdr_backend/pdr_dashboard/dash_components/plots.py
calina-c: I redid the unnecessary structs, you are absolutely right. I didn't know I could manipulate p.cols directly

calina-c:
@kdetry I addressed your comments. I'm also thinking about adding a follow-up task for the formatting part (another issue and bullet in #1594), where we could work with cols instead of values. What do you think?

pdr_backend/pdr_dashboard/dash_components/plots.py
kdetry: you are right

kdetry (approved):
lgtm

PR #1626, opened 2024-09-23
Fix #1618: Fix bronze table query error

### LINKED ISSUE #1618: [PdrDashboard] Querying parquet files sometimes throws file error
### Describe the bug

When running queries on the parquet table files, sometimes the query fails with the following error:
`ERROR Error querying the database: Invalid Input Error: No magic bytes found at end of file '/home/berkay/pdr-backend/lake_data/exports/bronze_pdr_predictions/bronze_pdr_predictions_1726742963000.parquet'`

Couldn't reproduce on local machine and only running into it inside the VM deployment

KatunaNorbert:
Fixes #1618


KatunaNorbert:
Found the issue. The problem was that we were constantly nuking the bronze table folder and re-exporting the parquet file and because the lake ETL runs each few seconds the file was constantly deleted and re-created which caused the reading error. Modified to only nuke it after the seconds_between_export have passed from last file update

kdetry (approved):
lgtm, if we can merge after the test issue

PR #1629, opened 2024-09-24
[Fix #1628] Add direct col processing to dataframe formatting

### LINKED ISSUE #1628: [PdrDashboard] Formatting with format_df should work with df columns directly, not values.

calina-c:
Closes #1628


pdr_backend/pdr_dashboard/util/format.py
```diff
@@ -111,74 +123,98 @@ def format_df(
...
+        (pl.col(col).round(2).cast(pl.String) + pl.lit("%")).alias(col)
     )
-    ws_buy_str = (
-        f"{format_currency(ws_buy, suffix='', show_decimal=False)} WS"
-        if ws_buy > 0
-        else ""
+
+
+@enforce_types
+def format_sales_info_data(df: pl.DataFrame) -> pl.DataFrame:
```
kdetry, line 155: It could be I missed what you do, I am not sure but it might be like the following:

```
def format_sales_info_data(df: pl.DataFrame) -> pl.DataFrame:
    if df.is_empty():
        return df.with_columns(pl.col("sales_str").alias("sales_str"))

    total = format_currency(df, "sales")["sales"]
    df_buy = format_currency(df, "df_buy_count", suffix=" DF")["df_buy_count"]
    ws_buy = format_currency(df, "ws_buy_count", suffix=" WS")["ws_buy_count"]

    # Combine all operations and directly assign to sales_str
    df = df.with_columns(
        [
            pl.when(pl.col("df_buy_count") > 0).then(df_buy).otherwise(pl.lit("")).alias("df_buy_str"),
            pl.when(pl.col("ws_buy_count") > 0).then(ws_buy).otherwise(pl.lit("")).alias("ws_buy_str"),
            
            # Create sales_str
            (
                total
                + pl.lit(" (")
                + pl.concat_str([pl.col("df_buy_str"), pl.col("ws_buy_str")], separator=", ")
                + pl.lit(")")
            )
            .str.replace("( )", "")  # If both df_buy_str and ws_buy_str are empty, remove the parentheses
            .alias("sales_str")
        ]
    )

    return df
```

kdetry (approved):
I have a suggestion, feel free to consider if it is convenient

pdr_backend/pdr_dashboard/util/format.py
calina-c: it's another two-step, unfortunately. When using pl.col("df_buy_str") it's the one defined in the same operation.

PR #1631, opened 2024-09-24
Fix #1627: Loading UI improvements

### LINKED ISSUE #1627: [PdrDashboard] Loading UI improvements
### Background / motivation

Improve UI for better data loading and display experience

### TODOs / DoD

- [ ] Add loading icons only for the data that's waited to be loaded, for the others display the component( ex: Navbar menu is displayed only when available lake data range is received from the backend )
- [ ] Add 1W for selected period to improve short term visibility and initial data loading
- [ ] Add margin to the app body to have some spacing between the content and the browser width&height
- [ ] If no predictoor addrs is configured, select one by default so when app loads, the chars on the home page are not empty

kdetry:
Fixes #1627


pdr_backend/pdr_dashboard/assets/styles.css
```diff
@@ -30,6 +30,8 @@ body {
   margin: 0px;
   padding: 0px;
   width: 100%;
+  padding-left: 20px;
+  padding-right: 20px;
```
KatunaNorbert, line 34: we want to have this on the entire page content, so we can move this to the html body

KatunaNorbert:
If I select the 'All' period, I run into this error:
Screenshot 2024-09-24 at 14 18 46

KatunaNorbert:
On the default predictoor selection task, chose a predictoor which has some predictions( in my case it didn't and resulted the same empty charts ), and I also get an issue when I deselect the predictoor:
Screenshot 2024-09-24 at 14 24 42

pdr_backend/pdr_dashboard/assets/styles.css
```diff
@@ -121,6 +121,77 @@ body {
...
+    animation: shimmer 2s infinite;
+}
+
+.metric_unit {
+    display: flex;
+    flex-direction: column;
+    font-size: 20px;
+}
+
+#feeds_page_metrics_row_wrapper:has(> *[data-dash-is-loading="true"]) .metric_value {
```
KatunaNorbert, line 162: I think for the metrics it's better if we wrap the component inside a` dcc.Loading` component, just like the metrics from the home page.
Also on the home page, for the graphs, instead of a displaying a loading component when data is loading, display some empty graphs and a loading message(maybe also text and loading component) just like in the table modals

kdetry: The dcc.Loading does not let a skeleton loading. It is suggested on the dash docs

KatunaNorbert: Ok, the target is to use the same loading pattern everywhere, so If you want to use skeleton loading the replace the dcc.Loading. Initially I was thinking on using skeleton just for the table, because we can't use dcc.Loading for the cells

kdetry:
can you check if the data exist on the selected predictoor and the selected feed for the 1W period?

KatunaNorbert:
Loading for the charts inside the modal looks good. Could you also add it to the home page charts to keep them consistent.
Another thing that would look good if the name of the chart is also displayed alongside the loading screen.

KatunaNorbert:
Sometimes when switching between pages and date range I run into the following error:
Screenshot 2024-10-02 at 16 03 25
Screenshot 2024-10-02 at 16 06 34

KatunaNorbert:
The initial load take too much with the latest updates. No matter on which page I am, the initial load takes around 8 seconds each time which is just too much

pdr_backend/pdr_dashboard/callbacks/callbacks_common.py
```diff
@@ -7,10 +7,26 @@
...
+    check_data_loaded,
+)
 
 
 def get_callbacks_common(app):
+    @app.callback(
+        Output("is-initial-data-loaded", "data"),
+        Input("is-initial-data-loaded", "data"),
+    )
+    def initial_data_load(is_data_loaded):
```
calina-c, line 22: I think this thread has some pertinent options: https://community.plotly.com/t/is-there-a-way-to-trigger-load-on-initial-page-load-only-and-not-every-time-a-change-is-made-to-the-page/57504

pdr_backend/pdr_dashboard/callbacks/callbacks_feeds.py
```diff
@@ -1,39 +1,63 @@
...
+            html.Div(),
+            filter_options[0],
+            filter_options[1],
+            filter_options[2],
+            filter_options[3],
+        )
 
     @app.callback(
-        Output("feeds_page_table", "data"),
+        Output("feeds_page_table", "data", allow_duplicate=True),
```
calina-c, line 60: why was this moved here? The previous version had the other one marked with allow_duplicate

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
```diff
@@ -85,7 +137,8 @@ def get_display_data_from_db(
 
     @app.callback(
         Output("predictoors_table", "data", allow_duplicate=True),
-        Output("predictoors_table", "selected_rows"),
+        Output("predictoors_table", "selected_rows", allow_duplicate=True),
+        Output("table-rows-count-predictoors_table", "children", allow_duplicate=True),
```
calina-c, line 141: I don't think we should add more outputs. Why not add a callback that has predictoors_table -> data as an input, that updates the count label?

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
```diff
@@ -151,11 +204,16 @@ def update_predictoors_table_on_search(
...
+            selected_predictoor_indices,
+            f"({len(filtered_data)})",
+        )
 
     @app.callback(
-        Output("feeds_table", "data"),
-        Output("feeds_table", "selected_rows"),
+        Output("feeds_table", "data", allow_duplicate=True),
+        Output("feeds_table", "selected_rows", allow_duplicate=True),
+        Output("table-rows-count-feeds_table", "children", allow_duplicate=True),
```
calina-c, line 216: same here

pdr_backend/pdr_dashboard/pages/feeds.py
```diff
@@ -95,10 +68,9 @@ def get_metrics_row(self):
             children=[
                 get_metric(
                     label=key,
-                    value=value,
                     value_id=f"feeds_page_{key}_metric",
                 )
-                for key, value in self.app.data.feeds_metrics_data.items()
+                for key in ["Feeds", "Accuracy", "Volume", "Sales", "Revenue"]
```
calina-c, line 73: I think we can define these in constants. What do you think? We have col headers there, so we might as well have metrics names

calina-c (commented):
I love this!

pdr_backend/pdr_dashboard/predictoor_dash.py
```diff
@@ -21,8 +21,12 @@ def predictoor_dash(ppss: PPSS, debug_mode: bool):
     if not debug_mode:
         webbrowser.open(f"http://127.0.0.1:{port}/")
 
-    app = Dash(__name__, external_stylesheets=[dbc.themes.BOOTSTRAP])
-    app.config["suppress_callback_exceptions"] = True
+    print("Starting Dash app")
```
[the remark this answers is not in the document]
kdetry: removed

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
```diff
@@ -9,11 +9,64 @@
...
+        Output("home_page_table_control_predictoors_table", "children"),
+        Output("home_page_table_control_feeds_table", "children"),
+        Output("table-rows-count-predictoors_table", "children"),
+        Output("table-rows-count-feeds_table", "children"),
+        [
+            Input("is-initial-data-loaded", "data"),
+        ],
+    )
+    @check_data_loaded(output_count=9)
+    def get_initial_data(_):
```
calina-c, line 34: should we also update the "stored predictoors" with what is initially selected? I remember @KatunaNorbert saying we can not load the ppss.yml favourite addresses directly into the state, but maybe with this newer implementation we can now. What do you guys think?

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
```diff
@@ -9,11 +9,64 @@
...
 def get_callbacks_home(app):
+    @app.callback(
+        Output("feeds_table", "data"),
+        Output("feeds_table", "selected_rows"),
+        Output("predictoors_table", "data"),
+        Output("predictoors_table", "selected_rows"),
+        Output("is-initial-table-loaded", "data"),
+        Output("home_page_table_control_predictoors_table", "children"),
+        Output("home_page_table_control_feeds_table", "children"),
+        Output("table-rows-count-predictoors_table", "children"),
```
calina-c, line 27: since we have a separate callback for rows counting, we can remove the row count outputs from here and let that callback do its job without interference.

pdr_backend/pdr_dashboard/util/helpers.py
```diff
@@ -90,4 +154,69 @@ def get_date_period_text_for_selected_predictoors(payouts: pl.DataFrame) -> str:
...
+
+    return (
+        df["base_token"].unique().to_list(),
+        df["quote_token"].unique().to_list(),
+        df["source"].unique().to_list(),
+        df["timeframe"].unique().to_list(),
+    )
+
+
+def check_data_loaded(
```
calina-c, line 188: I'm also wondering whether this decorator can help us with the sorting part which we eliminated as an option at some point

pdr_backend/pdr_dashboard/util/helpers.py
```diff
@@ -90,4 +154,69 @@ def get_date_period_text_for_selected_predictoors(payouts: pl.DataFrame) -> str:
...
+
+                return dash.no_update  # single output
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+def get_feeds_for_favourite_predictoors(app, feed_data, predictoor_addrs):
```
calina-c, line 213: I think this can be in the data manager class and replace all app.data with self

calina-c (changes requested):
I added some more suggestions :)

pdr_backend/pdr_dashboard/util/helpers.py
kdetry: maybe we can add another decorator for this solution

kdetry:
We use another PR for these changes

https://github.com/oceanprotocol/pdr-backend/pull/1661

PR #1635, opened 2024-09-26
Fix #1630: PdrDashboard cache predictoor addrs on browser

### LINKED ISSUE #1630: [PdrDashboard] Allow predictoor addresses configuration from browser
### Background / motivation

The dashboard app currently allows the configuration of the predictoor addresses to track inside the ppss.yaml file, which is a good approach if users are running the app locally, but because we want to deploy the app and remotely access is, this is no longer useful.
As a solution, allow the user to configure the addresses trough the browser and save them to local storage, so user can 

### TODOs / DoD

- [ ] Update UI to allow configuration of predictoor addresses to follow and save to browser storage
- [ ] When configured, load from local storage when app is loaded and display specific data just like loading from ppss

KatunaNorbert:
Fixes #1630


pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
```diff
@@ -120,21 +120,24 @@ def update_predictoors_table_on_search(
         ]
 
         if "show-favourite-addresses.value" in dash.callback_context.triggered_prop_ids:
-            custom_predictoors = formatted_predictoors_data.filter(
-                formatted_predictoors_data["full_addr"].is_in(
-                    app.data.favourite_addresses
+            if len(stored_predictoor_addrs) > 0:
```
calina-c, line 123: I would move this to a helper function that retrieves custom_predictoors either from the cache or ppss

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
```diff
@@ -225,3 +237,62 @@ def select_or_clear_all_predictoors(_, __, rows):
...
+
+    @app.callback(
+        Output("predictoor-addrs-local-store", "data"),
+        Input("save_predictoors", "n_clicks"),
+        State("predictoor_addrs", "value"),
+    )
+    def save_predictoors_to_browser_storage(n_clicks, value):
+        if not value and not n_clicks:
+            return dash.no_update
+        addresses = value.split("\n")
```
calina-c, line 273: I don't think we need to be passing strings, State can accept a list of strings AFAIK

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
```diff
@@ -225,3 +237,62 @@ def select_or_clear_all_predictoors(_, __, rows):
...
+    def update_show_favourite_check_and_selected_predictoor_rows(
+        saved_predictoor_addrs,
+    ):
+        return (
+            [
+                True if len(saved_predictoor_addrs) > 0 else dash.no_update,
+                dash.no_update,
+                dash.no_update,
+            ]
+            if len(saved_predictoor_addrs) > 0 or app.data.favourite_addresses
```
calina-c, line 296: let's untangle this a bit, it looks difficult to read and test

pdr_backend/pdr_dashboard/pages/home.py
```diff
@@ -286,3 +301,42 @@ def get_table(self, table_id, columns, data):
...
+        return dbc.Modal(
+            html.Div(
+                [
+                    html.H2("Configure Favourite Predictoor Addresses"),
+                    html.P(
+                        """Add predictoor addresses to be saved 
+                        and automatically selected when the app opens.""",
+                        style={"marginBottom": 0},
+                    ),
+                    html.P("Enter one address per line."),
```
calina-c, line 315: https://dash.plotly.com/dash-html-components/datalist could also be an option

pdr_backend/pdr_dashboard/test/test_callbacks_home.py
```diff
@@ -89,8 +89,7 @@ def test_favourite_addresses_search_input(
     fav_addr_toggle.click()
     time.sleep(2)
     p_all, p_sel = _predictoor_count(dash_duo)
-    assert len(p_all) == 58
```
calina-c, line 92: I think this will sort itself out once the "storage" is initialised from ppss

calina-c (changes requested):

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
```diff
@@ -166,6 +169,7 @@ def update_predictoors_table_on_search(
         [
             State("feeds_table", "selected_rows"),
             State("feeds_table", "data"),
+            State("predictoor-addrs-local-store", "data"),
```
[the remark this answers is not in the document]
KatunaNorbert: the problem is the we can only get the stored data trough the callbacks. Maybe I can do some cleaning around the way we set the ppss addresses

pdr_backend/pdr_dashboard/pages/home.py
KatunaNorbert: from what I see that is mainly used as a dropdown select

pdr_backend/pdr_dashboard/test/test_callbacks_home.py
KatunaNorbert: we can't do that, the local storage can be accessed only trough callbacks

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
```diff
@@ -187,6 +192,13 @@ def update_feeds_table_on_search(
             predictoor_feeds_only, predictoors_addrs, search_value, selected_feeds
         )
 
+        if (
+            stored_predictoors_addrs
```
calina-c, line 196: what is the logic of this? Can you explain it here or in a comment? Please also note that selected_feeds is a list of dictionaries (table data), while your assignation is a list of ints, so the change is improper to the nature of the variable. Are you instead looking to change selected_feed_indices instead?

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
```diff
@@ -225,3 +237,62 @@ def select_or_clear_all_predictoors(_, __, rows):
...
+    )
+    def open_predictoors_config_modal(n_clicks, predictoor_addrs):
+        """
+        Select or clear all rows in the feeds table.
+        """
+        if len(predictoor_addrs) > 0:
+            predictoor_addrs = "\n".join(predictoor_addrs)
+        else:
+            predictoor_addrs = ""
+        if n_clicks <= 0:
```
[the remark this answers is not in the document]
calina-c: how about
predictoor_addrs_str = "\n".join(predictoor_addrs) if predictoor_addrs else ""

predictoor addrs will evaluate to False if it is None, and to False if it is an empty list, so you're good on that
also please use the specific `_str` variable instead, for better clarity that this is not a list (anymore).

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
```diff
@@ -224,3 +236,62 @@ def select_or_clear_all_predictoors(_, __, rows):
...
+    def open_predictoors_config_modal(n_clicks, predictoor_addrs):
+        """
+        Select or clear all rows in the feeds table.
+        """
+        if predictoor_addrs and len(predictoor_addrs) > 0:
+            predictoor_addrs = "\n".join(predictoor_addrs)
+        else:
+            predictoor_addrs = ""
+        if n_clicks <= 0:
+            return [dash.no_update, predictoor_addrs]
```
calina-c, line 260: Then simply return
bool(n_clicks > 0), predictoor_addrs_str

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
calina-c: did you look into this? Can you have a list of strings instead of a customly created string?

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
```diff
@@ -224,3 +236,62 @@ def select_or_clear_all_predictoors(_, __, rows):
...
+        Input("save_predictoors", "n_clicks"),
+        State("predictoor_addrs", "value"),
+    )
+    def save_predictoors_to_browser_storage(n_clicks, value):
+        if not value and not n_clicks:
+            return dash.no_update
+        addresses = value.split("\n")
+        addresses = [
+            addr.strip() for addr in addresses if addr.strip()
+        ]  # Remove empty lines
```
calina-c, line 275: addresses = filter(len, str_list) would probably have the same effect of removing empty lines

calina-c (changes requested):
I added some more comments. Please look into the old unresolved ones as well.

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
KatunaNorbert: Basically what that block does is to select all feeds of the stored predictoors. We want to do this only on the initial load that's why the `sorted(predictoors_addrs) == sorted(stored_predictoors_addrs)` check, when user selects other predictoors we don't want to automatically select the feeds

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
KatunaNorbert: took and looks and filter() filters out empty string, while strip() removes whitespaces which returns different results

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
KatunaNorbert: I don't get it, isn't the 'addresses' a list of string?

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
KatunaNorbert: It would be hard to move something that is using the cache separately, because the cache can only be accessed inside the callback

pdr_backend/pdr_dashboard/callbacks/callbacks_home.py
calina-c: I was talking about the `value`. But you are right, that's the value of the textarea. Marking this as resolved

calina-c (approved):

PR #1644, opened 2024-10-01
Fix #1643: Omit Canceled slots while calculating accuracy

### LINKED ISSUE #1643: [Analytics] Omit Canceled slots when calculating feed accuracies
### Describe the bug
The feed accuracy could vary a lot when Canceled slots are taken in consideration while calculating accuracy, and because the Trueval of those slots are always set as False and the price doesn't change, the predictions are becoming irrelevant and shouldn't be considered in the overall accuracy.

### Expected behavior
The accuracy of the feeds should be calculated without taking Canceled slots in consideration.

KatunaNorbert:
Fixes #1643


kdetry (approved):
lgtm

PR #1649, opened 2024-10-02
Fix #1648: Display date range last date time and update frequency + label renamings

### LINKED ISSUE #1648: [PdrDashoard] Display data range end date time and label renamings
### Background / motivation

Because lake data and metrics are periodically updated, display the available range date end time and the frequency of which the update occurs. Also rename a few labels that are not clear enough on the app.

### TODOs / DoD

- [ ] Display data range end time
- [ ] Display data update frequency
- [ ] Labels renaming

KatunaNorbert:
Fixes #1648

### DIFF
--- a/pdr_backend/pdr_dashboard/util/helpers.py
+++ b/pdr_backend/pdr_dashboard/util/helpers.py
@@ -64,12 +64,17 @@ def select_or_clear_all_by_table(
 
 @enforce_types
 def _format_date_text(
-    start_date: Optional[datetime], end_date: Optional[datetime]
+    start_date: Optional[datetime],
+    end_date: Optional[datetime],
+    show_end_date_time: bool = False,
 ) -> str:
     if not start_date or not end_date:
         return "there is no data available"
 
-    return f"{start_date.strftime('%d-%m-%y')} -> {end_date.strftime('%d-%m-%y')}"
+    return f"""
+        {start_date.strftime('%d-%m-%y')} -> {end_date.strftime('%d-%m-%y')}
+        {end_date.strftime(' %H:%M') if show_end_date_time else ''}
+    """
 
 
 @enforce_types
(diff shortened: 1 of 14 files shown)


pdr_backend/pdr_dashboard/util/helpers.py
```diff
@@ -85,4 +90,4 @@ def get_date_period_text_for_selected_predictoors(payouts: pl.DataFrame) -> str:
 
 @enforce_types
 def get_date_period_text_header(start_date: UnixTimeS, end_date: UnixTimeS) -> str:
-    return _format_date_text(start_date.to_dt(), end_date.to_dt())
+    return _format_date_text(start_date.to_dt(), end_date.to_dt(), True)
```
calina-c, line 93: please add the keyword argument in full, it is vague to just have "True"

pdr_backend/pdr_dashboard/util/helpers.py
```diff
@@ -64,12 +64,17 @@ def select_or_clear_all_by_table(
...
+    start_date: Optional[datetime],
+    end_date: Optional[datetime],
+    show_end_date_time: bool = False,
 ) -> str:
     if not start_date or not end_date:
         return "there is no data available"
 
-    return f"{start_date.strftime('%d-%m-%y')} -> {end_date.strftime('%d-%m-%y')}"
+    return f"""
+        {start_date.strftime('%d-%m-%y')} -> {end_date.strftime('%d-%m-%y')}
```
calina-c, line 75: only end date gets the time part?

KatunaNorbert: I think the start date is not relevant since it is the same hour each time only the date changes, and only when ETL is run for the first time. The end date is constantly changing and it's much more relevant with the incremental update because is constantly changing

calina-c (approved):

PR #1661, opened 2024-10-11
Fix #1627: Loading UI improvements with Segfault

### LINKED ISSUE #1627: [PdrDashboard] Loading UI improvements

kdetry:
Fixes #1627


calina-c:
On this branch in particular, I suggest you only run pdr_dashboard tests. The rest of the suite is not problematic and it will decrease your CI debugging time.

KatunaNorbert (approved):

PR #1731, opened 2025-02-03
Fix #1730: Available lake data end time is not updated

### LINKED ISSUE #1730: [PdrDashboard] Available lake data end date not updating
### Describe the bug
Available lake data end date should constantly update when the cached data is updated and reflect the state of cached data. Currently it's only getting updated when the app first starts.
This is important because is the main value that reflects the state of available data for dashboard.

KatunaNorbert:
Fixes #1730


KatunaNorbert:
For easy testing I deployed the fix to dashboard.predictoor.ai, It should update every 5 min when the page is refreshed

